PR #8, opened 2020-03-23
[WiP][PoC][Travis] Revamped test matrix

alongosz:
Let's see first if this even works before proceeding with all the details

#### Checklist:
- [ ] PR description is updated.


.travis.yml
```diff
@@ -58,33 +90,15 @@ branches:
...
   # Disable memory_limit for composer
   - echo "memory_limit=-1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
-  # Disable XDebug for all jobs as we don't generate test coverge on travis
+  # Disable XDebug for all jobs as we don't generate test coverage on Travis
   - phpenv config-rm xdebug.ini
   # make sure we use UTF-8 encoding
   - echo "default_charset=UTF-8" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
-  # Install igbinary & lzf PHP extensions if necessary
-  - if [ "$ENABLE_IGBINARY" = true ] ; then pecl install igbinary ; fi
-  - if [ "$ENABLE_LZF" = true ] ; then printf "no\n" | pecl install lzf ; fi
```
andrerom: I guess we then leave this for Symfony responsibility to test against right?

makes sense, we added it in earlier branches given we had to do overloading on kernel side to take advantage of these features, but that is not the case anymore in 3.x where all is taken from Symfony.

alongosz: > I guess we then leave this for Symfony responsibility to test against right?
> 
> makes sense, we added it in earlier branches given we had to do overloading on kernel side to take advantage of these features, but that is not the case anymore in 3.x where all is taken from Symfony.

Yes. For now I'd like to keep unused things out of the setup as it already is too big.

PR #11, opened 2020-03-26
EZP-30921: Dropped deprecated eZc database abstraction layer

alongosz:
| Question                                  | Answer
| ---------------------------------------- | ------------------
| **JIRA issue**                          | [EZP-30921](https://jira.ez.no/browse/EZP-30921)
| **Type**                                   | improvement
| **Target eZ Platform version** | `v3.0`
| **BC breaks**                          | yes
| **Tests pass**                          | yes
| **Doc needed**                       | yes

This PR contains cumulative miscellaneous changes needed in order to completely drop eZc Database Handler (a.k.a Zeta Components Database abstraction layer) from eZ Platform Kernel.

It is blocked by #28 and thus contains it.
The current range for review is: https://github.com/ezsystems/ezplatform-kernel/pull/11/files/c787764aaf9d67af214782994f7a0beb09735e61.

#### Checklist:
- [x] All parts of [EZP-30921](https://jira.ez.no/browse/EZP-30921) are completed.
- [x] Tests pass.
- [x] Code follows Coding Standards (use `$ composer fix-cs`).
- [x] PR is ready for a review.

PR #14, opened 2020-03-27
EZP-31246: Refactored Test setup to stop relying on Database Handler

alongosz:
| **JIRA issue**                          | [EZP-31246](https://jira.ez.no/browse/EZP-31246) blocking [EZP-30921](https://jira.ez.no/browse/EZP-30921)
| **Target version**                    | `ezplatform-kernel:1.0` (internal package test setup)

### TL;DR; Key points:

1. Yaml test data: ab1c97375 and c2ecf1aa1
2. `FixtureImporter`: fcaa722a4
3. Usage of `FixtureImporter`: a9cfb3a27 and 1dae04abe

### Summary of changes

This PR drops most of[1] eZc Database Handler usages from our tests (both integration and unit).
You can review it commit by commit - I tried to keep them as separate as possible, though maintaining an order during reviewing is probably recommended, as some changes rely on previous changes.

The most interesting aspect though, is not the refactoring of query building, but concepts I introduced in order to instantiate test setup. It heavily relied on Database Handler, so had to be changed anyway.

The initial data fixtures were stored in `eZ/Publish/Core/Repository/Tests/Service/Integration/Legacy/_fixtures/test_data.php`.
The file was big, triggered unnecessary static analysis, and overall was difficult to maintain.

I decided to go with Yaml for Kernel 1.0: `eZ/Publish/API/Repository/Tests/_fixtures/Legacy/data/test_data.yaml`. 
It feels more light for IDE in terms of both viewing and updating. It's obviously still too big, but one thing at a time. 
Added via ab1c97375 and updated via c2ecf1aa1.

As an intermediate step, I've upgraded SetupFactory to use that new data file and import it using Doctrine QueryBuilder (7908cea64047e7cd3b0783cfe1bd72ddf1281798).
This led to another key point, which is moving this responsibility to a set of separate classes. Hence, I've introduced via fcaa722a4733e63e76f7a0404df0c1711564963d `FixtureImporter` with in-memory caching layer and `Fixture` instances for PHP files (for some cases) and Yaml files (the only one now - `test_data.yaml`). It greatly improved readability of both Setup Factory (a9cfb3a27ca38b0906dc6ce504a0001083b86973) and unit tests base TestCase (1dae04abe1a3a3d571a658b8717afd7b761aaf32).

Less important thing, but useful in the long run: `\eZ\Publish\Core\Persistence\Legacy\Tests\TestCase::getDatabaseConnection` now fails a test case on connection error instead of re-throwing. Connection error usually does not require analyzing stack trace anyway. This change will make our test cases fetching connection not to report unhandled exception (dbeb8f767112d264f7528a1b4fb094d7f633c0d7).

You might also wonder why I've dropped some tests via 25dccae (`eZ\Publish\Core\Persistence\Doctrine\Tests`). Those are for Database Handler being dropped, this is why they're gone.

Other commits are strictly related to refactoring.

---
_[1] There's still `\eZ\Publish\Core\Persistence\Legacy\Tests\TestCase::getDatabaseHandler` which I cannot drop until #11 (or PRs to be split from that) are completed and merged._

#### Checklist:
- [x] Tests are passing.


eZ/Publish/SPI/Tests/Persistence/FileFixtureFactory.php
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @see \eZ\Publish\SPI\Tests\Persistence\Fixture
+ */
+final class FileFixtureFactory
+{
+    public function buildFixture(string $filePath): Fixture
+    {
+        $fileInfo = new SplFileInfo($filePath);
+        $extension = $fileInfo->getExtension();
+        // note: there's no dependency injection available here, so using simple switch
```
adamwojs:
```suggestion
```

eZ/Publish/SPI/Tests/Persistence/FixtureImporter.php
```diff
@@ -0,0 +1,160 @@
...
+            $this->resetSequences($tablesList);
+        }
+    }
+
+    /**
+     * @param string[] $tables a list of table names
+     *
+     * @throws \Doctrine\DBAL\DBALException
+     */
+    private function truncateTables(array $tables)
```
adamwojs:
```suggestion
    private function truncateTables(array $tables): void
```

eZ/Publish/SPI/Tests/Persistence/PhpArrayFileFixture.php
```diff
@@ -0,0 +1,22 @@
...
+declare(strict_types=1);
+
+namespace eZ\Publish\SPI\Tests\Persistence;
+
+/**
+ * Data fixture stored in PHP file which returns it as an array.
+ *
+ * @internal for internal use by Repository test setup
+ */
+class PhpArrayFileFixture extends BaseInMemoryCachedFileFixture
```
adamwojs:
```suggestion
final class PhpArrayFileFixture extends BaseInMemoryCachedFileFixture
```

PR #18, opened 2020-03-28
EZP-31528: Offset option added to search reindex command

SerheyDolgushev:
| **JIRA issue**                          | [EZP-31528](https://jira.ez.no/browse/EZP-31528)

It might take a lot of time to rebuild the search index on big projects, and there should be an option to resume interrupted command

#### Checklist:
- [x] Tests are implemented.

### DIFF
--- a/eZ/Bundle/EzPublishCoreBundle/Command/ReindexCommand.php
+++ b/eZ/Bundle/EzPublishCoreBundle/Command/ReindexCommand.php
@@ -158,6 +158,12 @@ class ReindexCommand extends Command
                 InputOption::VALUE_OPTIONAL,
                 'Number of child processes to run in parallel for iterations, if set to "auto" it will set to number of CPU cores -1, set to "1" or "0" to disable',
                 'auto'
+            )->addOption(
+                'offset',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'Offset, useful when the script was interrupted and need to be resumed',
+                0
             )->setHelp(
                 <<<EOT
 The command <info>%command.name%</info> indexes the current configured database in the configured search engine index.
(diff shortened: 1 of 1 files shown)


eZ/Bundle/EzPublishCoreBundle/Command/ReindexCommand.php
```diff
@@ -374,12 +383,13 @@ private function getStatementContentSince(DateTime $since, $count = false)
...
      * @param mixed $locationId
      * @param bool $count
+     * @param int $offset
      *
      * @return \Doctrine\DBAL\Driver\Statement
      *
      * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
      */
-    private function getStatementSubtree($locationId, $count = false)
+    private function getStatementSubtree($locationId, $count = false, int $offset = 0)
```
adamwojs, line 392: `$offset` parameter is always passed so there is no need for default value. 

_It's private method so you can safely change the method signature without BC break._

eZ/Bundle/EzPublishCoreBundle/Command/ReindexCommand.php
```diff
@@ -396,15 +407,17 @@ private function getStatementSubtree($locationId, $count = false)
 
     /**
      * @param bool $count
+     * @param int $offset
      *
      * @return \Doctrine\DBAL\Driver\Statement
      */
-    private function getStatementContentAll($count = false)
+    private function getStatementContentAll($count = false, int $offset = 0)
```
adamwojs, line 414: `$offset` parameter is always passed so there is no need for default value.

eZ/Bundle/EzPublishCoreBundle/Command/ReindexCommand.php
```diff
@@ -158,6 +158,12 @@ protected function configure()
                 InputOption::VALUE_OPTIONAL,
                 'Number of child processes to run in parallel for iterations, if set to "auto" it will set to number of CPU cores -1, set to "1" or "0" to disable',
                 'auto'
+            )->addOption(
+                'offset',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'Offset, useful when the script was interrupted and need to be resumed',
+                0
```
adamwojs, line 166: Validation is missing here. Throw `InvalidArgumentException` when `offset < 0`.  For example:  https://github.com/ezsystems/ezplatform-kernel/blob/master/eZ/Bundle/EzPublishCoreBundle/Command/ReindexCommand.php#L193-L195 .

adamwojs (commented):
Q: How will I know which offset should be used to resume indexing?

SerheyDolgushev:
> Q: How will I know which offset should be used to resume indexing?

The use case for the offset is:
1. Full reindex started as a background job and all its output if forwarded to a log file. It will take 12 hours to rebuild the index.
2. After the reindex script is running 9 hours it stopped by some reason (server was rebooted, etc).
3. Using the log from the first step, it should be possible to find out how much content was indexed already. And use it as the offset parameter.

SerheyDolgushev:
@adamwojs Sorry, any update on this one? Do you think something needs to be changed here?

gggeek:
Needs a fix to the doc (and maybe the code): the IDs list options should be incompatible with the Offset

PR #19, opened 2020-03-28
EZP-31529: Added --no-interaction option to cmd regenerating URL aliases

SerheyDolgushev:
| **JIRA issue**                          | [EZP-31529](https://jira.ez.no/browse/EZP-31529)

Regenerate URL aliases command is an important part of upgrade scripts. And because it ignores `no-interaction` option it is much harder to automate the upgrade scripts.

#### Checklist:
- [x] Tests are implemented.

### DIFF
--- a/eZ/Bundle/EzPublishCoreBundle/Command/RegenerateUrlAliasesCommand.php
+++ b/eZ/Bundle/EzPublishCoreBundle/Command/RegenerateUrlAliasesCommand.php
@@ -82,7 +82,7 @@ EOT;
                 'force',
                 'f',
                 InputOption::VALUE_NONE,
-                'Prevents confirmation dialog. Please use it carefully.'
+                'Prevents confirmation dialog when used with --no-interaction. Please use it carefully.'
             )->setHelp(
                 <<<EOT
 {$beforeRunningHints}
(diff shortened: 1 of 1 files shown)


SerheyDolgushev:
Sorry, https://symfony.com/doc/current/components/console/usage.html says following about `--no-interaction` option:
> You can suppress any interactive questions from the command you are running with

Do you propose to add additional `--force` flag just to make sure the user is aware of what he is doing? So no confirmation message will be shown only if both `--force` and `--no-interaction` are passed? And the script will continue to work as now if just one of those flags is passed?

SerheyDolgushev:
Cool, will implement this change tonight.

eZ/Bundle/EzPublishCoreBundle/Command/RegenerateUrlAliasesCommand.php
```diff
@@ -78,6 +78,11 @@ protected function configure()
                 InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
                 'Only Locations with provided ID\'s will have URL aliases regenerated',
                 []
+            )->addOption(
+                'force',
+                'f',
+                InputOption::VALUE_NONE,
+                'Prevents confirmation dialog. Please use it carefully.'
```
alongosz: Let's be more precise:
```suggestion
                'Prevents confirmation dialog when used with --no-interaction. Please use it carefully.'
```
requesting because currently, when `--no-interaction` is not set, `--force` will have no effect - confirmation will be shown.

SerheyDolgushev: Done in https://github.com/ezsystems/ezplatform-kernel/pull/19/commits/8d4b804fae8879673f5a7dd78300d0b2254e6b16

PR #21, opened 2020-03-28
Fixed integration tests to rely on `ezpublish.persistence.connection`

alongosz:
| **Target eZ Platform version** | `v3.0` internal test framework
| **Tests pass**                          | TBD

The `ezpublish.persistence.connection` service is a more stable abstraction than `ezpublish.api.storage_engine.legacy.connection`. 
While the latter one is available for tests ran on `ezplatform-kernel` due to [some Container tricks](https://github.com/ezsystems/ezplatform-kernel/blob/0a7dac34e860d6aa179944bf42e9af2f4b43b0b1/eZ/Publish/Core/settings/containerBuilder.php#L87), it's not easily propagated to the dependencies, see failing [ezplatform-richtext job](https://travis-ci.org/github/ezsystems/ezplatform-richtext/builds/667649690). 
On the other hand `ezpublish.persistence.connection` seems to be available there also (requires confirming with CI).

Changed also Base integration test cases `getDatabaseConnection`, `getRawDatabaseConnection`, and `getSetupFactory` methods not to throw exception but to fail any test case as there's no point in running it nor examining stack trace (so added as a string only).

Side: fixed CS issues in the modified methods and methods related to their invocation chain.

#### Checklist:
- [x] Tests are passing.
- [x] Failing RichText test is passing.

### DIFF
--- a/eZ/Publish/API/Repository/Tests/BaseTest.php
+++ b/eZ/Publish/API/Repository/Tests/BaseTest.php
@@ -9,6 +9,7 @@
 namespace eZ\Publish\API\Repository\Tests;
 
 use Doctrine\DBAL\Connection;
+use ErrorException;
 use eZ\Publish\API\Repository\Exceptions\ContentFieldValidationException;
 use eZ\Publish\API\Repository\Exceptions\ForbiddenException;
 use eZ\Publish\API\Repository\Exceptions\NotFoundException;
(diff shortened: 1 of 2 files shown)


alongosz:
Tests that [were failing](https://travis-ci.org/github/ezsystems/ezplatform-richtext/builds/667649690) are [passing now](https://travis-ci.org/github/ezsystems/ezplatform-richtext/builds/668135188). Not possible to make Behat green due to cyclic dependency issues.
Merging to unblock CI hopefully, :crossed_fingers:

PR #26, opened 2020-03-31
EZEE-3069: Ensured type of $destination property value is int for location based url aliases

adamwojs:
| **JIRA issue**                          | [EZEE-3069](https://jira.ez.no/browse/EZEE-3069)
| **Target eZ Platform version** | `3.0`

According to doc: if url alias type is `URLAlias::LOCATION` then `destination` property contains location id which should be typeof int.  

#### Changes

Actual bugfix: https://github.com/ezsystems/ezplatform-kernel/pull/26/files#diff-50956c9b2b9e4174ee6c9013cf358264R111

Actual test: https://github.com/ezsystems/ezplatform-kernel/pull/26/files#diff-ce61cc0866ae34a229c2e0d39d33cd6bR125-R148

all other intorduced changes allows to use strict comparison in `assertPropertiesCorrect`.

#### Checklist:
- [X] Tests are implemented.

### DIFF
--- a/eZ/Publish/API/Repository/Tests/BaseTest.php
+++ b/eZ/Publish/API/Repository/Tests/BaseTest.php
@@ -202,21 +202,21 @@ abstract class BaseTest extends TestCase
      * @param mixed[] $expectedValues
      * @param \eZ\Publish\API\Repository\Values\ValueObject $actualObject
      */
-    protected function assertPropertiesCorrect(array $expectedValues, ValueObject $actualObject, bool $strictComparison = false)
+    protected function assertPropertiesCorrect(array $expectedValues, ValueObject $actualObject)
     {
         foreach ($expectedValues as $propertyName => $propertyValue) {
             if ($propertyValue instanceof ValueObject) {
-                $this->assertStructPropertiesCorrect($propertyValue, $actualObject->$propertyName, [], $strictComparison);
+                $this->assertStructPropertiesCorrect($propertyValue, $actualObject->$propertyName);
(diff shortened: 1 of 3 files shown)


eZ/Publish/API/Repository/Tests/BaseTest.php
```diff
@@ -202,21 +202,21 @@ protected function getSetupFactory(): SetupFactory
      * @param mixed[] $expectedValues
      * @param \eZ\Publish\API\Repository\Values\ValueObject $actualObject
      */
-    protected function assertPropertiesCorrect(array $expectedValues, ValueObject $actualObject)
+    protected function assertPropertiesCorrect(array $expectedValues, ValueObject $actualObject, bool $strictComparison = false)
```
alongosz: -1, this should reflect PHPUnit API, at least on a method naming level (because there's this debt that we didn't design those assertions PHPUnit way, by extending proper classes).
So minimum: should be `assertPropertiesEqual`, `assertPropertiesSame`.

janit:
I can confirm a patched version works on Platform.sh where I initially reported EZEE-3069.

eZ/Publish/API/Repository/Tests/BaseTest.php
adamwojs: Sorry but proper refactoring of  this method should be processes as separate dedicated story to don't obfuscate real sope of the JIRA issue and unblock people which waiting for the fix. Moved to https://jira.ez.no/browse/EZP-31534

PR #28, opened 2020-03-31
EZP-31302: Refactored LSE to rely on Doctrine Connection

alongosz:
| **JIRA issue**                          | [EZP-31302](https://jira.ez.no/browse/EZP-31302) blocking [EZP-30921](https://jira.ez.no/browse/EZP-30921)

### TL; DR;

This PR replaces eZc Database Handler, Query and Expression Builder with Doctrine Connection and Query Builder in Legacy Search Engine gateways, Criterion Handlers, and Sort Clauses.

### Key differences with respect to eZc Database Query Builder.

Doctrine Query Builder advocates aliasing tables when building queries. In case of joins it's even required. 
While aliasing table as itself is possible, it's error prone, redundant, and not efficient (queries sent over network are much longer).
Therefore, when doing refactoring, most of the tables were given an alias. Some tables have intentionally different aliases to avoid collisions between outer and inner nested queries.

The most common aliases are:
- `c` for `ezcontentobject` table,
- `v` for `ezcontentobject_version` table
- `f_def` for `ezcontentobject_attribute` table (stands for Field Definition, following Domain language).

I've also used named parameters for easier debugging of queries. Names cannot be literal/fixed because the same Criterion for a query can be applied multiple times with different arguments. I could generate placeholder names, but Doctrine actually does that for us when no name is given.

What could be doubtful at first sight is parameter binding for outer queries instead of sub queries. This is specific to Doctrine Query Builder. Sub-queries are used just for convenience, but in the end they're just injected as strings into outer queries. Therefore any parameter defined in sub-query needs to be actually set on the outer one. It proved to be quite error prone, so maybe we need to rethink the design of Criteria and Sort Clauses handlers in the future.

### Doc

Criterion Handlers and Sort Clauses are considered extension points, so please see the doc in the PR for breaking changes.

#### Checklist:
- [x] Tests are passing.

### DIFF
--- a/eZ/Publish/Core/Search/Legacy/Content/Common/Gateway/CriterionHandler/LogicalNot.php
+++ b/eZ/Publish/Core/Search/Legacy/Content/Common/Gateway/CriterionHandler/LogicalNot.php
@@ -1,17 +1,15 @@
 <?php
 
 /**
- * File containing the DoctrineDatabase logical not criterion handler class.
- *
  * @copyright Copyright (C) eZ Systems AS. All rights reserved.
  * @license For full copyright and license information view LICENSE file distributed with this source code.
  */
 namespace eZ\Publish\Core\Search\Legacy\Content\Common\Gateway\CriterionHandler;
 
+use Doctrine\DBAL\Query\QueryBuilder;
 use eZ\Publish\Core\Search\Legacy\Content\Common\Gateway\CriterionHandler;
 use eZ\Publish\Core\Search\Legacy\Content\Common\Gateway\CriteriaConverter;
 use eZ\Publish\API\Repository\Values\Content\Query\Criterion;
-use eZ\Publish\Core\Persistence\Database\SelectQuery;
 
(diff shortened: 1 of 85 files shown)


eZ/Publish/Core/Search/Legacy/Content/Common/Gateway/CriterionHandler/LogicalNot.php
```diff
@@ -30,26 +28,15 @@ public function accept(Criterion $criterion)
...
     public function handle(
         CriteriaConverter $converter,
-        SelectQuery $query,
+        QueryBuilder $queryBuilder,
         Criterion $criterion,
         array $languageSettings
     ) {
-        return $query->expr->not(
-            $converter->convertCriteria($query, $criterion->criteria[0], $languageSettings)
+        return sprintf(
```
ViniTou, line 37: Why is this no longer handled by $queryBuilder?

PR #37, opened 2020-04-05
EZP-31542: Added PHP 7.4 support  (eZ Platform 3.0)

adamwojs:
| **JIRA issue**                          | [EZP-31542](https://jira.ez.no/browse/EZP-31542)

The most significant changes from our perspective: 

```
Trying to use values of type null, bool, int, float or resource as an array (such as $null["key"]) will now generate a notice. This does not affect array accesses performed by list(). RFC: https://wiki.php.net/rfc/notice-for-non-valid-array-container
```
```
Passing invalid characters to ''base_convert()'', ''bindec()'', ''octdec()'' and ''hexdec()'' will now generate a deprecation notice. The result will still be computed as if the invalid characters did not exist. Leading and trailing whitespace, as well as prefixes of type 0x (depending on base) continue to be allowed.
```

```
Throwing exceptions from __toString() is now permitted. Previously this resulted in a fatal error. Existing recoverable fatals in string conversions have been converted to Error exceptions. RFC: https://wiki.php.net/rfc/tostring_exceptions
```

Full upgrade notes: https://github.com/php/php-src/blob/PHP-7.4.5/UPGRADING

#### Checklist:
- [x] Tests are implemented.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,12 @@ matrix:
     - name: 'Code Style Check'
       php: 7.3
       env: CHECK_CS=1
+# 7.4
+    - php: 7.4
+      env: TEST_CONFIG="phpunit.xml"
+    - php: 7.4
+      env: TEST_CONFIG="phpunit-integration-legacy.xml" DB="postgresql" DATABASE="pgsql://postgres@localhost/testdb"
+
 
 # test only master, stable branches and pull requests
 branches:
--- a/eZ/Publish/Core/FieldType/Relation/Type.php
+++ b/eZ/Publish/Core/FieldType/Relation/Type.php
@@ -239,12 +239,14 @@ class Type extends FieldType
      */
     public function fromHash($hash)
     {
-        $destinationContentId = $hash['destinationContentId'];
-        if ($destinationContentId !== null) {
-            $destinationContentId = (int)$destinationContentId;
+        if ($hash !== null) {
+            $destinationContentId = $hash['destinationContentId'];
+            if ($destinationContentId !== null) {
+                return new Value((int)$destinationContentId);
+            }
         }
 
-        return new Value($destinationContentId);
+        return $this->getEmptyValue();
     }
 
     /**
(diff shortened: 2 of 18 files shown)


eZ/Publish/Core/FieldType/Relation/Type.php
```diff
@@ -241,12 +241,14 @@ protected function getSortInfo(BaseValue $value)
      */
     public function fromHash($hash)
     {
-        $destinationContentId = $hash['destinationContentId'];
-        if ($destinationContentId !== null) {
-            $destinationContentId = (int)$destinationContentId;
+        if ($hash !== null) {
```
ViniTou, line 244: In my opinion returning early would improve readability here.

.travis.yml
```diff
@@ -24,26 +24,44 @@ env:
...
+      php: 7.4
+      env: TEST_CONFIG="phpunit.xml"
     - name: "Kernel Behat Core tests"
       php: 7.3
       env:
         - COMPOSE_FILE="doc/docker/base-dev.yml:doc/docker/selenium.yml"
         - BEHAT_OPTS="--profile=core --tags=~@broken"
         - APP_ENV=behat
         - APP_DEBUG=1
+    - name: "Kernel Behat Core tests (PHP 7.4)"
```
mnocon, line 40: I'd remove this job as well

adamwojs: Fixed in https://github.com/ezsystems/ezplatform-kernel/pull/37/commits/af8b9ca0a48aaf5805e90034d607f7e0d16120dc

eZ/Publish/Core/FieldType/Relation/Type.php
alongosz: ^

adamwojs: Suggestion applied in https://github.com/ezsystems/ezplatform-kernel/pull/37/commits/e0e13406a98f7b72222c0987fa5e4ff006b65307

PR #38, opened 2020-04-07
EZP-31546: Fixed ignored path prefixes for binaries and images

webhdx:
| **JIRA issue**                          | [EZP-31546](https://jira.ez.no/browse/EZP-31546)
| **Target eZ Platform version** | `v3.0.1`

#### Checklist:

### DIFF
--- a/eZ/Publish/Core/IO/ConfigScopeChangeAwareIOService.php
+++ b/eZ/Publish/Core/IO/ConfigScopeChangeAwareIOService.php
@@ -33,6 +33,9 @@ class ConfigScopeChangeAwareIOService implements IOServiceInterface, ConfigScope
         $this->configResolver = $configResolver;
         $this->innerIOService = $innerIOService;
         $this->prefixParameterName = $prefixParameterName;
+
+        // set initial prefix on inner IOService
+        $this->setPrefix($this->configResolver->getParameter($this->prefixParameterName));
     }
 
     public function setPrefix($prefix): void
--- a/eZ/Publish/Core/IO/Tests/ConfigScopeChangeAwareIOServiceTest.php
+++ b/eZ/Publish/Core/IO/Tests/ConfigScopeChangeAwareIOServiceTest.php
@@ -16,7 +16,7 @@ use eZ\Publish\Core\MVC\ConfigResolverInterface;
 use eZ\Publish\Core\MVC\Symfony\SiteAccess;
 use PHPUnit\Framework\TestCase;
 
-class ConfigScopeChangeAwareIOServiceTest extends TestCase
+final class ConfigScopeChangeAwareIOServiceTest extends TestCase
 {
     protected const PREFIX = 'test-prefix';
     protected const PREFIX_PARAMETER_NAME = 'param';
--- a/eZ/Publish/Core/IO/Tests/ConfigScopeChangeAwareIOServiceTest.php
+++ b/eZ/Publish/Core/IO/Tests/ConfigScopeChangeAwareIOServiceTest.php
@@ -1,0 +1,320 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace eZ\Publish\Core\IO\Tests;
+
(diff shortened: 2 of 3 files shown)


ITernovtsiy:
This fix will not work properly, as problem is in fact of `innerIOService` shared across instances of `ConfigScopeChangeAwareIOService` services.
So, the last instantiated service prefix used globally.

For example, you have these services definition:
```
ezpublish.fieldType.ezimage.io_service.draft:
        parent: ezpublish.core.io.service
        calls:
            - [ setPrefix, [ "%image_draft_storage_prefix%" ] ]
ezpublish.fieldType.ezbinaryfile.io_service:
        parent: ezpublish.core.io.service
        calls:
            - [ setPrefix, [ "%binaryfile_storage_prefix%" ] ]
```

The container will have 2 instances of `ConfigScopeChangeAwareIOService`, but with shared `innerIOService` instance. So, when `setPrefix` called for `ezpublish.fieldType.ezbinaryfile.io_service`, it overrides the one defined for `ezpublish.fieldType.ezimage.io_service.draft`.

cc: @SerheyDolgushev

webhdx:
Good catch @ITernovtsiy. I will use different approach then.

webhdx:
I've changed the solution, please take a look @ITernovtsiy @SerheyDolgushev. Also it's ready for review. It's now complete with unit tests.

eZ/Publish/Core/IO/Tests/ConfigScopeChangeAwareIOServiceTest.php
```diff
@@ -0,0 +1,320 @@
...
+
+use eZ\Publish\Core\IO\ConfigScopeChangeAwareIOService;
+use eZ\Publish\Core\IO\IOServiceInterface;
+use eZ\Publish\Core\IO\Values\BinaryFile;
+use eZ\Publish\Core\IO\Values\BinaryFileCreateStruct;
+use eZ\Publish\Core\MVC\ConfigResolverInterface;
+use eZ\Publish\Core\MVC\Symfony\SiteAccess;
+use PHPUnit\Framework\TestCase;
+
+class ConfigScopeChangeAwareIOServiceTest extends TestCase
```
adamwojs, line 19:
```suggestion
final class ConfigScopeChangeAwareIOServiceTest extends TestCase
```

eZ/Publish/Core/IO/Tests/ConfigScopeChangeAwareIOServiceTest.php
```diff
@@ -0,0 +1,320 @@
...
+{
+    protected const PREFIX = 'test-prefix';
+    protected const PREFIX_PARAMETER_NAME = 'param';
+
+    /** @var \eZ\Publish\Core\IO\ConfigScopeChangeAwareIOService */
+    protected $IOService;
```
alongosz, line 25: CS
```suggestion
    protected $ioService;
```

eZ/Publish/Core/IO/ConfigScopeChangeAwareIOService.php
```diff
@@ -33,6 +33,9 @@ public function __construct(
         $this->configResolver = $configResolver;
         $this->innerIOService = $innerIOService;
         $this->prefixParameterName = $prefixParameterName;
+
+        // set initial prefix on inner IOService
+        $this->setPrefix($this->configResolver->getParameter($this->prefixParameterName));
```
alongosz, line 38: If you have to do it here, the service should be `lazy` to avoid getting parameter from uninitialized Config Resolver.

webhdx:
Done @adamwojs @alongosz

ITernovtsiy:
@webhdx I think it makes sense to keep consistency and define `ezpublish.fieldType.ezbinaryfile.io_service.config_scope_change_aware:` with `decorates:` too.
It will reduce the risk of prefix mismatched again, as this service will definitely use own copy of the origin io service.

Currently, I'm not sure about what's the prefix is in `@ezpublish.core.io.service` itself. Should it always be the same as in `ezbinaryfile` field type or it should use some own value.

webhdx:
It would make sense to make `ezpublish.core.io.service` abstract but we can't do it due to BC. I will leave it as is.

I've fixed `ezpublish.fieldType.ezbinaryfile.io_service` in d723df3

SerheyDolgushev:
Looks good to me

PR #41, opened 2020-04-08
EZEE-3091: Used twig dump to display object arguments

ViniTou:
| **JIRA issue**                          | [EZEE-3091](https://jira.ez.no/browse/EZEE-3091)

Before SiteAccess factory, there were only scalar values used in `$arguments` array in `\eZ\Publish\Core\Persistence\Cache\PersistenceLogger::logCall` method.
When object was used it unsuccessfully tried to cast it to string.

#### Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,7 @@
         "symfony/console": "^5.0",
         "symfony/expression-language": "^5.0",
         "symfony/validator": "^5.0",
+        "symfony/var-dumper": "^5.0",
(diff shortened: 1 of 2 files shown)


emodric:
Is this really a good idea? It will allow having `dump` function on production, which, if forgotten and left in code could reveal sensitive info from PHP variables (instead of just crashing the rendering).

emodric:
There is a dedicated `profiler_dump` Twig function which can be used instead.

PR #47, opened 2020-04-09
EZEE-3085: Added specification ContentTypeSpecification

devmysz:
| **JIRA issue**                          | [EZEE-3085](https://jira.ez.no/browse/EZEE-3085)
| **Target eZ Platform version** | `v3.0.1` - please update accordingly

Added specification ContentTypeSpecification

#### Checklist:
- [x] Tests are implemented. 

### DIFF
--- a/eZ/Publish/SPI/Specification/Tests/Content/ContentContainerSpecificationTest.php
+++ b/eZ/Publish/SPI/Specification/Tests/Content/ContentContainerSpecificationTest.php
@@ -10,9 +10,9 @@ namespace eZ\Publish\SPI\Specification\Tests\Content;
 
 use eZ\Publish\API\Repository\Values\Content\Content;
 use eZ\Publish\API\Repository\Values\ContentType\ContentType;
-use eZ\Publish\Core\Search\Tests\TestCase;
 use eZ\Publish\SPI\Specification\Content\ContentContainerSpecification;
 use eZ\Publish\SPI\Specification\Content\ContentSpecification;
+use PHPUnit\Framework\TestCase;
 
 /**
  * @covers \eZ\Publish\SPI\Specification\Content\ContentContainerSpecification
@@ -28,13 +28,12 @@ final class ContentContainerSpecificationTest extends TestCase
 
     /**
      * @covers \eZ\Publish\SPI\Specification\Content\ContentContainerSpecification::isSatisfiedBy
-     * @dataProvider isSatisfiedByProvider
+     * @dataProvider providerForIsSatisfiedBy
      */
     public function testIsSatisfiedBy(
         bool $isContainer,
         bool $shouldBeSatisfied
-    ): void
-    {
+    ): void {
         $contentContainerSpecification = new ContentContainerSpecification();
 
         $contentTypeMock = $this->getMockBuilder(ContentType::class)
@@ -54,7 +53,7 @@ final class ContentContainerSpecificationTest extends TestCase
         );
     }
 
-    public static function isSatisfiedByProvider(): array
+    public function providerForIsSatisfiedBy(): array
     {
         return [
             [true, true],
--- a/eZ/Publish/SPI/Specification/Tests/Content/ContentContainerSpecificationTest.php
+++ b/eZ/Publish/SPI/Specification/Tests/Content/ContentContainerSpecificationTest.php
@@ -1,0 +1,64 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace eZ\Publish\SPI\Specification\Tests\Content;
+
+use eZ\Publish\API\Repository\Values\Content\Content;
+use eZ\Publish\API\Repository\Values\ContentType\ContentType;
+use eZ\Publish\Core\Search\Tests\TestCase;
+use eZ\Publish\SPI\Specification\Content\ContentContainerSpecification;
+use eZ\Publish\SPI\Specification\Content\ContentSpecification;
+
+/**
+ * @covers \eZ\Publish\SPI\Specification\Content\ContentContainerSpecification
+ */
+final class ContentContainerSpecificationTest extends TestCase
+{
+    public function testConstructor(): void
+    {
+        $contentTypeSpecification = new ContentContainerSpecification();
+
+        $this->assertInstanceOf(ContentSpecification::class, $contentTypeSpecification);
+    }
+
+    /**
+     * @covers \eZ\Publish\SPI\Specification\Content\ContentContainerSpecification::isSatisfiedBy
+     * @dataProvider isSatisfiedByProvider
+     */
+    public function testIsSatisfiedBy(
+        bool $isContainer,
+        bool $shouldBeSatisfied
+    ): void
+    {
+        $contentContainerSpecification = new ContentContainerSpecification();
+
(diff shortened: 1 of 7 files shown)


Nattfarinn (changes requested):
Provide unit tests for `ContentTypeSpecification`.

alongosz (changes requested):
Could you elaborate on the purpose of this and naming?
I don't see this doing anything right now (not configured?).

I see that naming was taken from `SpecificationInterface` defined by @mikadamczyk in AdminUI in 2018.
@mikadamczyk if you remember, could you explain the choice of naming? Is it some kind of a design pattern or does it reflect existing solution? I don't remember being part of this discussion.

Without the broader context I don't feel like this belongs to the Repository domain

michal-myszka:
> Could you elaborate on the purpose of this and naming?
> I don't see this doing anything right now (not configured?).
> 
> I see that naming was taken from `SpecificationInterface` defined by @mikadamczyk in AdminUI in 2018.
> @mikadamczyk if you remember, could you explain the choice of naming? Is it some kind of a design pattern or does it reflect existing solution? I don't remember being part of this discussion.
> 
> Without the broader context I don't feel like this belongs to the Repository domain

This becomes from Behavioral Design Pattern - Specification:
https://designpatternsphp.readthedocs.io/en/latest/Behavioral/Specification/README.html

The `ContentTypeSpecification` is currently used in 
https://github.com/ezsystems/ezplatform-page-fieldtype/pull/141/files#diff-5adf309520410df755badb9f39017b56R51
 and should be available for the other packages.

alongosz:
> This becomes from Behavioral Design Pattern - Specification:
> https://designpatternsphp.readthedocs.io/en/latest/Behavioral/Specification/README.html
> 
> The `ContentTypeSpecification` is currently used in
> https://github.com/ezsystems/ezplatform-page-fieldtype/pull/141/files#diff-5adf309520410df755badb9f39017b56R51
> and should be available for the other packages.

Because it's used in the following way there
```php
$contentTypeSpecification = new ContentTypeSpecification($expectedType);
```
the `ContentTypeSpecification` implementation should also be a part of SPI. We don't give BC promise on anything from Core, unless documented. Documented Core stuff is a technical debt at this point.
If you were to define `ContentTypeSpecification` in the Page Field Type package, then what we have, minus that implementation would be fine. 
We could consider `ContentTypeSpecification` an API, but then it can't be "newable" externally.

TBH it looks like an overkill, but as long as it follows A(S)PI rules, I'm fine ;)

eZ/Publish/SPI/Specification/Tests/Content/ContentContainerSpecificationTest.php
```diff
@@ -0,0 +1,64 @@
...
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace eZ\Publish\SPI\Specification\Tests\Content;
+
+use eZ\Publish\API\Repository\Values\Content\Content;
+use eZ\Publish\API\Repository\Values\ContentType\ContentType;
+use eZ\Publish\Core\Search\Tests\TestCase;
```
alongosz, line 13: Seems like an accidental import. You've meant to add the he one from PHPUnit, right? ;)

eZ/Publish/SPI/Specification/Tests/Content/ContentContainerSpecificationTest.php
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * @covers \eZ\Publish\SPI\Specification\Content\ContentContainerSpecification::isSatisfiedBy
+     * @dataProvider isSatisfiedByProvider
+     */
+    public function testIsSatisfiedBy(
+        bool $isContainer,
+        bool $shouldBeSatisfied
+    ): void
+    {
```
alongosz, line 37: CS
```suggestion
    ): void {
```

eZ/Publish/SPI/Specification/Tests/Content/ContentTypeSpecificationTest.php
```diff
@@ -0,0 +1,78 @@
...
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace eZ\Publish\SPI\Specification\Tests\Content;
+
+use eZ\Publish\API\Repository\Values\Content\Content;
+use eZ\Publish\API\Repository\Values\ContentType\ContentType;
+use eZ\Publish\Core\Search\Tests\TestCase;
```
alongosz, line 13: Same comment as before

eZ/Publish/SPI/Specification/Tests/Content/ContentContainerSpecificationTest.php
```diff
@@ -0,0 +1,64 @@
...
+            ->method('getContentType')
+            ->willReturn($contentTypeMock);
+
+        $this->assertEquals(
+            $contentContainerSpecification->isSatisfiedBy($contentMock),
+            $shouldBeSatisfied
+        );
+    }
+
+    public static function isSatisfiedByProvider(): array
```
alongosz, line 57: Optional: when I see method starting with `is...` I immediately want to see `:bool` as a return type.
I'd avoid it, by e.g. naming it: `providerForIsSatisfiedBy` or, following verb based method naming conventions `getDataForIsSatisfiedBy`.
Both are fine for tests.
But not insisting.

eZ/Publish/SPI/Specification/Tests/Content/ContentTypeSpecificationTest.php
```diff
@@ -0,0 +1,78 @@
...
+            ->method('getContentType')
+            ->willReturn($contentTypeMock);
+
+        $this->assertEquals(
+            $contentTypeSpecification->isSatisfiedBy($contentMock),
+            $shouldBeSatisfied
+        );
+    }
+
+    public static function isSatisfiedByProvider(): array
```
alongosz, line 71: Optional: the same comment

eZ/Publish/SPI/Specification/Tests/Content/ContentContainerSpecificationTest.php
alongosz, line 57: Question: does PHPUnit handle well `static` providers? Is it documented somewhere? Usually it's non-static.

alongosz (changes requested):
Ok for SPI,
Remarks for tests:

eZ/Publish/SPI/Specification/Tests/Content/ContentContainerSpecificationTest.php
michal-myszka: Right .. :)

eZ/Publish/SPI/Specification/Tests/Content/ContentContainerSpecificationTest.php
michal-myszka: I found a lot of usages in our codebase, but that's fact that in PHPUnit docs there is no any information about static call - so I've changed to non-static method.

alongosz (approved):

Nattfarinn (approved):

ViniTou (approved):

katarzynazawada (approved):

michal-myszka:
Merged up to master

PR #51, opened 2020-04-20
EZP-31588: As a Developer I want to configure search engine features with tags which follows same convention

adamwojs:
| **JIRA issue**                          | [EZP-31588](https://jira.ez.no/browse/EZP-31588)

Tags used to register Search Engine features in the dependency injection container should be consistent and follow Symfony convention (snake case).

Commits between https://github.com/ezsystems/ezplatform-kernel/commit/73ae9f70097be965563c5005bfb7022cf63ac61a - https://github.com/ezsystems/ezplatform-kernel/commit/d0ad33f8caddb031700d0ca9d880d16245fc6448 contains refactoring of existing backward compatibility layer for `ezplatform.field_type.*`  `ezplatform.query_type.*`. 

Basically duplicated logic of iteration over service ids tagged wtith current and deprecated tag has been extracted to `\eZ\Publish\Core\Base\Container\Compiler\TaggedServiceIdsIterator\BackwardCompatibleIterator`. 

Commit https://github.com/ezsystems/ezplatform-kernel/commit/76bce15df51f7642cc61b352fa0bc2b29a809b85 deprecates  `ezpublish.searchEngine` and `ezpublish.searchEngineIndexer` in favour of `ezplatform.search_engine` and `ezplatform.search_engine.indexer`

#### Checklist:
- [X] Tests are implemented.
- [X] PR is ready for a review.

### DIFF
--- a/eZ/Bundle/EzPublishCoreBundle/DependencyInjection/Compiler/RegisterSearchEngineIndexerPass.php
+++ b/eZ/Bundle/EzPublishCoreBundle/DependencyInjection/Compiler/RegisterSearchEngineIndexerPass.php
@@ -6,16 +6,20 @@
  */
 namespace eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Compiler;
 
+use eZ\Publish\Core\Base\Container\Compiler\TaggedServiceIdsIterator\BackwardCompatibleIterator;
+use LogicException;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Reference;
-use LogicException;
 
 /**
  * This compiler pass registers eZ Publish search engines indexers.
  */
 class RegisterSearchEngineIndexerPass implements CompilerPassInterface
 {
+    public const SEARCH_ENGINE_INDEXER_SERVICE_TAG = 'ezplatform.search_engine.indexer';
+    public const DEPRECATED_SEARCH_ENGINE_INDEXER_SERVICE_TAG = 'ezpublish.searchEngineIndexer';
+
     /**
      * Container service id of the SearchEngineIndexerFactory.
      *
@@ -39,12 +43,23 @@ class RegisterSearchEngineIndexerPass implements CompilerPassInterface
         }
 
         $searchEngineIndexerFactoryDefinition = $container->getDefinition($this->factoryId);
-        foreach ($container->findTaggedServiceIds('ezpublish.searchEngineIndexer') as $id => $attributes) {
+
+        $iterator = new BackwardCompatibleIterator(
+            $container,
+            self::SEARCH_ENGINE_INDEXER_SERVICE_TAG,
+            self::DEPRECATED_SEARCH_ENGINE_INDEXER_SERVICE_TAG
+        );
+
+        foreach ($iterator as $serviceId => $attributes) {
             foreach ($attributes as $attribute) {
                 if (!isset($attribute['alias'])) {
                     throw new LogicException(
-                        'ezpublish.searchEngineIndexer service tag needs an "alias" attribute to ' .
-                        'identify the search engine.'
+                        sprintf(
+                            'Service "%s" tagged with "%s" or "%s" needs an "alias" attribute to identify the search engine',
+                            $serviceId,
+                            self::SEARCH_ENGINE_INDEXER_SERVICE_TAG,
+                            self::DEPRECATED_SEARCH_ENGINE_INDEXER_SERVICE_TAG
+                        )
                     );
                 }
 
@@ -52,7 +67,7 @@ class RegisterSearchEngineIndexerPass implements CompilerPassInterface
                 $searchEngineIndexerFactoryDefinition->addMethodCall(
                     'registerSearchEngineIndexer',
                     [
-                        new Reference($id),
+                        new Reference($serviceId),
                         $attribute['alias'],
                     ]
                 );
--- a/eZ/Bundle/EzPublishCoreBundle/DependencyInjection/Compiler/RegisterSearchEnginePass.php
+++ b/eZ/Bundle/EzPublishCoreBundle/DependencyInjection/Compiler/RegisterSearchEnginePass.php
@@ -6,16 +6,20 @@
  */
 namespace eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Compiler;
 
+use eZ\Publish\Core\Base\Container\Compiler\TaggedServiceIdsIterator\BackwardCompatibleIterator;
+use LogicException;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Reference;
-use LogicException;
 
 /**
  * This compiler pass will register eZ Publish search engines.
  */
 class RegisterSearchEnginePass implements CompilerPassInterface
 {
+    public const SEARCH_ENGINE_SERVICE_TAG = 'ezplatform.search_engine';
+    public const DEPRECATED_SEATCH_ENGINE_SERVICE_TAG = 'ezpublish.searchEngine';
+
     /**
      * Container service id of the SearchEngineFactory.
      *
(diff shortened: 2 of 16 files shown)


eZ/Publish/Core/Base/Container/Compiler/TaggedServiceIdsIterator/BackwardCompatibleIterator.php
```diff
@@ -0,0 +1,55 @@
...
+ */
+declare(strict_types=1);
+
+namespace eZ\Publish\Core\Base\Container\Compiler\TaggedServiceIdsIterator;
+
+use Iterator;
+use IteratorAggregate;
+use Symfony\Component\DependencyInjection\TaggedContainerInterface;
+
+final class BackwardCompatibleIterator implements IteratorAggregate
```
alongosz, line 15:
```suggestion
/**
 * @internal
 */
final class BackwardCompatibleIterator implements IteratorAggregate
```
right?

eZ/Publish/Core/Base/Tests/Container/Compiler/TaggedServiceIdsIterator/DeprecationErrorCaptor.php
```diff
@@ -0,0 +1,50 @@
...
+declare(strict_types=1);
+
+namespace eZ\Publish\Core\Base\Tests\Container\Compiler\TaggedServiceIdsIterator;
+
+/**
+ * Captures user deprecation warnings emitted using trigger_error function.
+ *
+ * @internal
+ */
+final class DeprecationErrorCaptor
```
alongosz, line 16: Does this captor demand a ransom? ;)

I'm not really sold on this naming, the most commonly used name for this kind of classes is Collector, I think. Unless this follows some kind of existing pattern?

General question: why do you need to add extra complexity to error handling? We rather rely on Monolog and Symfony Console handlers on that. Is it just to be able to process those errors in unit tests? For that we usually rely on warnings being converted to exceptions and validate them by `->expectException`. Doesn't this work here?

eZ/Bundle/EzPublishCoreBundle/DependencyInjection/Compiler/RegisterSearchEngineIndexerPass.php
```diff
@@ -39,12 +43,22 @@ public function process(ContainerBuilder $container)
...
+        );
+
+        foreach ($iterator as $id => $attributes) {
             foreach ($attributes as $attribute) {
                 if (!isset($attribute['alias'])) {
                     throw new LogicException(
-                        'ezpublish.searchEngineIndexer service tag needs an "alias" attribute to ' .
-                        'identify the search engine.'
+                        sprintf(
+                            '%s or %s service tag needs an "alias" attribute to identify the search engine.',
```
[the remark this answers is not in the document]
adamwojs: Fixed in https://github.com/ezsystems/ezplatform-kernel/pull/51/commits/77373994cd6e94f32c282e29d7498199bbca2fb7

eZ/Bundle/EzPublishCoreBundle/DependencyInjection/Compiler/RegisterSearchEnginePass.php
```diff
@@ -40,12 +44,21 @@ public function process(ContainerBuilder $container)
...
+        );
+
+        foreach ($iterator as $id => $attributes) {
             foreach ($attributes as $attribute) {
                 if (!isset($attribute['alias'])) {
                     throw new LogicException(
-                        'ezpublish.searchEngine service tag needs an "alias" attribute to ' .
-                        'identify the search engine.'
+                        sprintf(
+                            '%s or %s service tag needs an "alias" attribute to identify the search engine.',
```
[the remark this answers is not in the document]
adamwojs: Fixed in https://github.com/ezsystems/ezplatform-kernel/pull/51/commits/77373994cd6e94f32c282e29d7498199bbca2fb7

eZ/Publish/Core/Base/Container/Compiler/TaggedServiceIdsIterator/BackwardCompatibleIterator.php
adamwojs: Sure. Fixed in https://github.com/ezsystems/ezplatform-kernel/pull/51/commits/77373994cd6e94f32c282e29d7498199bbca2fb7.

eZ/Publish/Core/Base/Tests/Container/Compiler/TaggedServiceIdsIterator/DeprecationErrorCaptor.php
adamwojs: > I'm not really sold on this naming, the most commonly used name for this kind of classes is Collector, I think. Unless this follows some kind of existing pattern?

It was inspirated by https://www.javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/ArgumentCaptor.html but collector seems to be better naming choice,

> General question: why do you need to add extra complexity to error handling? We rather rely on Monolog and Symfony Console handlers on that. Is it just to be able to process those errors in unit tests? For that we usually rely on warnings being converted to exceptions and validate them by ->expectException. Doesn't this work here?

There are dedicated `->expectDeprecation(...)`, `->expectDeprecationMessage(...)` but for some reason they doesn't work. I need to figure out why.

adamwojs:
>What we can't have is a squash commit assigned to this specific Story in Release notes, but containing all this stuff (too unrelated). Could you, after review, but before QA, squash commits in such way we could merge it in a fast-forward way? One for BC layer, N for compiler passes related to specific features, and finally one for EZP-31588, prefixed with that ticket number.

Sure, but I would do only one commit for `BackwardCompatibleterator` + refactoring and one for EZP-31588. Plan for the future is to unify all available tags as separate PR's.

alongosz:
> > What we can't have is a squash commit assigned to this specific Story in Release notes, but containing all this stuff (too unrelated). Could you, after review, but before QA, squash commits in such way we could merge it in a fast-forward way? One for BC layer, N for compiler passes related to specific features, and finally one for EZP-31588, prefixed with that ticket number.
> 
> Sure, but I would do only one commit for `BackwardCompatibleterator` + refactoring and one for EZP-31588. Plan for the future is to unify all available tags as separate PR's.

Fine by me. However please do not copy JIRA story format in commit message, it's not very useful for browsing commit history (again, it's a history, so it's the past anyway... ;)). RN contains full JIRA title, so it's fine.
OT: after "As a Developer" there should be no comma. I don't know who started to introduce this grammar mistake :-)

eZ/Publish/Core/Base/Container/Compiler/TaggedServiceIdsIterator/BackwardCompatibleIterator.php
```diff
@@ -0,0 +1,58 @@
...
+                E_USER_DEPRECATED
+            );
+
+            yield $serviceId => $tags;
+        }
+
+        $taggedServiceIds = $this->container->findTaggedServiceIds($this->serviceTag);
+        foreach ($taggedServiceIds as $serviceId => $tags) {
+            yield $serviceId => $tags;
+        }
```
mikadamczyk, line 56: At the end we could add `yield from [];`. In case `$serviceIdsWithDeprecatedTags` and `$taggedServiceIds` will be empty.

eZ/Publish/Core/Base/Tests/Container/Compiler/TaggedServiceIdsIterator/DeprecationErrorCaptor.php
adamwojs: `->expectDeprecation(...)` and `->expectDeprecationMessage(...)` only works when the deprecation message is triggered without error suppression operator (\`@\`) e.g.  
 

```php
trigger_error('Lorem ipsum dolor...', E_USER_DEPRECATED)   // OK
@trigger_error('Lorem ipsum dolor...', E_USER_DEPRECATED)  // KO
```

alongosz:
> 
> One thing that worries me - I can't get the deprecations warning to show up in logs.
> Not sure if it's misconfiguration on my side, I've found this comment from @alongosz:
> 
> ```
> Unfortunately my tests show that @trigger_error doesn't get logged when building Symfony container.
> ```
> 
> From: [ezsystems/ezpublish-kernel#2291](https://github.com/ezsystems/ezpublish-kernel/pull/2291) (~2 years ago)
> 
> I'm happy to discuss it tomorrow, I've used the prod monolog config: https://github.com/ezsystems/ezplatform/blob/master/config/packages/prod/monolog.yaml which contains a deprecations handler, but I might be missing something in config.

@mnocon Monolog is not available during Container building because Logger is a Container service. Chicken and the egg problem.
Those should be available in Container deprecations log 
`./var/cache/<env>/<name><Env><Debug>ProjectContainerDeprecations.log`
e.g: `./var/cache/dev/appDevDebugProjectContainerDeprecations.log`

mnocon:
@alongosz you're right! Found them in the mentioned file.

Thanks, this can be merged then (together with related PRs).

PR #53, opened 2020-04-23
[Internal RFC][Composer] Proposed new Ibexa namespaces

alongosz:
| **Type**                                   | Internal RFC
| **Doc needed**                       | not yet

### Background

We have a namespace chaos, looking from the entire Product perspective. "Publish" is mixed with "Platform", each package provides API in its own way.

[We're Ibexa now](https://www.ibexa.co/blog/ez-systems-announces-company-name-change-to-ibexa), which is a perfect opportunity for some changes.

### Introducing `Ibexa\Platform` namespace

Instead of waiting for the good opportunity (`v4.0`) to perform namespace refactoring from `eZ\Publish` and `EzSystems\EzPlatform*` to `Ibexa\Platform`, let's start an evolution already in `3.x` instead of a revolution for `4.x`. You have to admit that a bulk upgrade of all the packages for the next major never worked for us really well. This way we could avoid increasing this technical debt.

### New convention

I'm proposing the new namespace convention, to be already included for eZ Platform `v3.1` by Ibexa (it doesn't matter that "eZ" is still with us):

* `Ibexa\Platform\API\<PackageName>` to map consumer API Services (and other "injectables") and Values,
* `Ibexa\Platform\SPI\<PackageName>` to map integrator Service Provider Interfaces (extension points and "newables"[1]),
* `Ibexa\Platform\Core\<PackageName>` to map core/internal implementation of APIs and SPIs, not to be used by 3rd party code,
* `Ibexa\Platform\Tests\` for tests (needs to be in `autoload-dev` section to avoid overlapping).

The `<PackageName>` is arbitrary - whatever makes sense for the package. For Kernel I've chosen `Repository` because this is closer to our domain and the package responsibility[2]:

So for Kernel this PR proposes accordingly:
* `Ibexa\Platform\API\Repository`,
* `Ibexa\Platform\SPI\Repository`,
* `Ibexa\Platform\Core\Repository`,

The `autoload-dev` test namespace could be:
* `Ibexa\Platform\Tests\Integration`

It would be the same or similar in every package. Since it's a dev-autoloaded namespace it should be fine.

### Bundles

Naturally, Bundles would still have their own namespaces and are still expected to be placed in `./src/bundle` directory.

### Examples

The idea for 1st party packages introducing APIs is to follow the same or similar pattern. Let's use EE SiteFactory as an example[3]:

* `Ibexa\Platform\API\SiteFactory`,
* `Ibexa\Platform\SPI\SiteFactory`,
* `Ibexa\Platform\Core\SiteFactory` (current contents of `./src/lib` there minus APIs and SPIs - if present).

**Content/Location Filtering feature practical example: #54**

### Advantages

Unified namespaces across the entire Product and unified directory structure which comes with that makes the Product more Developer friendly. 
Experienced Developer would know where to find certain things (let's start my IDE search by "API" because I'm looking for an API). 
Skilled Developer would know how to use proper parts of the Product. While A(S)PI organization of a Product is not common to PHP ecosystem, it's quite easy to explain in terms of what you can and cannot do.
For us it also provides those advantages - we should know what can be safely changed and what needs an extra attention just by looking at a file path or a class namespace.

Last but not least, we've been saying for ages, that we're API-oriented and everything is API. Let's expose that more clearly in a unified way.

### Alternatives

1. Following the currently used pattern, the namespaces could be e.g. `Ibexa\Platform\Kernel\API`, `Ibexa\Platform\SiteFactory\API`. While it would make Composer `autoload` definition simpler, it would give an impression of a separation. Since we're API-oriented, API should be one hub with different services.
2. Contracts could have its own repository which each package would require. While it would look pretty cool and light, it would be at the same time a maintenance hell. To provide new feature we would have to target Contracts repository first and link the builds to see that everything is working properly.

### Controversy

1. I admit that placing here "src" next to "eZ" will be confusing for newcomers. But this is an intermediate step to avoid bulk changes.
2. Given many packages already follow the `./src/{lib,bundle}` convention, it's gonna be a challenge. However smaller refactoring keeping portions of the namespace, but introducing API and SPI would be beneficial anyway.

---
_[1] "newable" - a class which will be used by 3rd party as `new ClassName` - it requires strict BC policy, including a guarantee a constructor would work across all minor versions._

_[2] Yeah I know Kernel has more responsibilities than simply delivering a Content Repository. Focusing here on the package essence though and future restructuring & decoupling plans._

_[3] Using existing package to provide more relatable example, but since I'm proposing evolution, this would actually be something for new packages or something slowly being introduced in the old ones, the in same way as here._

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -81,7 +81,8 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "EzSystems\\PlatformInstallerBundleTests\\": "eZ/Bundle/PlatformInstallerBundle/tests"
+            "EzSystems\\PlatformInstallerBundleTests\\": "eZ/Bundle/PlatformInstallerBundle/tests",
+            "Ibexa\\Platform\\Tests\\Integration\\": "tests/lib/integration"
         }
     },
     "config": {
--- a/composer.json
+++ b/composer.json
@@ -73,7 +73,10 @@
     "autoload": {
         "psr-4": {
             "EzSystems\\PlatformInstallerBundle\\": "eZ/Bundle/PlatformInstallerBundle/src",
-            "eZ\\": "eZ"
+            "eZ\\": "eZ",
+            "Ibexa\\Platform\\API\\Repository\\": "src/lib/API/Repository",
+            "Ibexa\\Platform\\SPI\\Repository\\": "src/lib/SPI/Repository",
+            "Ibexa\\Platform\\Core\\Repository\\": "src/lib/Core/Repository"
         }
     },
     "autoload-dev": {


alongosz:
@ezsystems/php-dev-team @lserwatka 
I've updated namespaces to be `Ibexa\Platform\*`. Updated the description and the exemplary PR consuming them (#54).

Nattfarinn:
Does it actually make sense to set the RFC right now? Is it business confirmed the product name will remain `Platform`?

lserwatka:
It is confirmed. We can use proposed namespace as Platform is generic enough to fit new product name.

alongosz:
@Nattfarinn please read the description. Both alternatives were covered, for both I've explained design or maintenance issues.
Reducing scope just to `Ibexa\Platform` would essentially imply alternative 1., which would contribute to the current mess. Not sure if it's worth it.

Nattfarinn:
@alongosz Please read carefully. `Contract` namespace does not imply separate package and/or repository and this is just but one of additional alternatives. This is not *Either/Or* case, there has to be internal discussion about what direction should we pick and what naming convention should we use.

alongosz:
> @alongosz Please read carefully. `Contract` namespace does not imply separate package and/or repository and this is just but one of additional alternatives.

Ah, now I understand, sorry. Let's sync internally.

vidarl:
Changing name spaces would cause a mayor headache for people doing upgrade.. Remember, if we do such a major change, engineering only have to do it once - and be done with it... Partners would need to cope with this for years, for every upgrade they do.

Changing, `eZ\Publish\API\Repository\LocationService`  to `Ibexa\Platform\API\Repository\LocationService`  is technically not a problem. It is easy manageable with search/replace.

However, if we'll do restructuring like `EzSystems\EzPlatformSiteFactory\SomeService` is becomming `Ibexa\Platform\API\SiteFactory\SomeService`, then we'll have a problem.
Then I think you also need to provide a tool which contains mapping of old->new namespace and which can do the required replacements in custom code

emodric:
If changing namespaces is an imperative (which I would argue it is not, since it shatters any preconcieved notion of stability in the long run - do you plan to change the namespaces if eZ Systems rebrands itself once again?), you should consider doing something similar to what Laminas did when taking over Zend Framwork: https://getlaminas.org/blog/2020-03-09-transferring-zf-to-laminas.html

As @vidarl pointed out, "placing here "src" next to "eZ" will be confusing for newcomers" will be the least controversial thing for developers/partners who will have to upgrade a hefty list of projects for years to come.

alongosz:
Thank you for all your feedback.
First of all, sorry for a bit late update, but most of this RFC has been internally rejected. We're still going for `Ibexa\Platform` namespace, but not with the structure described here. I'm closing this one and I'll link new one as soon as we're ready with all the details.

Regarding eZ Platform 3.x, the `Ibexa\Platform` namespace is mostly aimed for new things (if and only once approved). If existing namespaces are to be upgraded, it's going to be backward-compatible.

For the entire process we indeed should prepare [Rector](https://github.com/rectorphp/rector). It's actually going to be quite straightforward in case of namespaces.

PHPUnit upgraded their namespaces. Twig upgraded their namespaces. Everyone survived.

As said, I'm closing the PR because the approach is quite outdated.

emodric:
Twig provided BC with old namespaces for the entire 2.x version. Only in 3.x did they remove the BC layer.

PHPUnit is also different. With it, you have the luxury of using the old version as long as you wish and chance that eZ would break someone else by using new phpunit namespaces was minimal.

With eZ kernel, BC should be top prioritycso old namespaces should be kept even in 4.x with class aliasing. All other solutions will just antagonize developers.

PR #54, opened 2020-04-23
EZP-31569: Implemented Repository filtering

alongosz:
| **JIRA issue**                          | EZP-31569

### Summary

This PR introduces Content Repository filtering API allowing a Developer to query Content Repository to retrieve either Content or Location list without involvement of a search engine.

#### Content Service

```php
use eZ\Publish\API\Repository\Values\Content\ContentList;
use eZ\Publish\API\Repository\Values\Filter\Filter;

interface ContentService {

    public function find(Filter $filter, ?array $languages = null): ContentList;
}
```

#### LocationService

```php
use eZ\Publish\API\Repository\Values\Filter\Filter;
use eZ\Publish\API\Repository\Values\Content\LocationList;

interface LocationService {

    public function find(Filter $filter, ?array $languages = null): LocationList;
}
```

Notice that I've kept `$languages` argument which injects context languages and is used by SiteAccessAware layer. Initially I planned to rely entirely on the `Filter` criteria, but the code required for this was too inconsistent and too complicated when compared with current SiteAccessAware implementation and test coverage.
`$language` behaves the same as for the other existing API endpoints - if `null`, then it should be chosen from a context of a layer. This allows to set an empty array `[]` to force no extra languages.

### API consumption

`Filter` is a fluent Value Object defining its own setters using grammar which encourages chained use. It does not introduce its own builder to avoid complicating already complex solution.

```php
use eZ\Publish\API\Repository\Values\Content\Query\Criterion\ParentLocationId;
use eZ\Publish\API\Repository\Values\Content\Query\Criterion\LanguageCode;
use eZ\Publish\API\Repository\Values\Filter\Filter;

$filter = (new Filter())
    ->withCriterion(new ParentLocationId($parentFolder->contentInfo->mainLocationId))
    ->andWithCriterion(new LanguageCode(['eng-GB']));

$contentList = $contentService->find($filter);
foreach ($contentList as $contentItem) {
    // ...
}
```

Full list of `Filter` fluent setters:

- `withCriterion` - accepts a single Criterion, Criterion cannot be already set
- `andWithCriterion` - appends a Criterion using LogicalAnd operation, if no Criterion was set prior this operation, it will simply set it instead (like `withCriterion`)
- `orWithCriterion` - append a Criterion using LogicalOr operation, the same rules apply as for `andWithCriterion`
- `withSortClause` - append a Sort Clause
- `sliceBy` - set limit and offset for pagination
- `reset` - remove all Criteria, Sort Clauses, and pagination settings for fluent reuse.

It's worth to point out that a `Filter` stores only a single  Criterion. If multiple Criteria are needed they should be nested inside one of the logical operators - `LogicalAnd`, `LogicalOr`, or `LogicalNot`.

On the other hand a list of Sort Clauses is just an array.

Setting pagination offset and limit is not a requirement, though it's a recommended practice. Unlike its search counterpart , a Query,  `Filter` does not impose default page limit, making results unlimited when not set. This is in line with database queries behavior, so should be less surprising than what search Query offers. Still, it's recommended to set limit and use pagination.

To avoid making the solution even bigger than it is right now, all the search Criteria are reused for filtering. They all reside in the `\eZ\Publish\API\Repository\Values\Content\Query\Criterion` namespace, so there seems nothing wrong with that approach.

The same goes for Sort Clauses - the ones from the `\eZ\Publish\API\Repository\Values\Content\Query\SortClause` namespace are reused.

**Important**: Not all Criteria, nor the Sort Clauses are supported with Repository Filtering. Some of them will appear later on, others, like FullText Criterion were never meant to work on a bare Repository without a content indexed for search.

To avoid confusion, a Criterion which can be used for Repository Filtering needs to implement `\eZ\Publish\SPI\Repository\Values\Filter\FilteringCriterion` interface. Similarly, 
a Sort Clause needs to implement `\eZ\Publish\SPI\Repository\Values\Filter\FilteringSortClause` interface. To improve DX, with proper IDE used, it will be immediately apparent that a given fluent setter does not accept a given Criterion or Sort Clause not implementing those interfaces.

#### Supported Criteria

- `Ancestor`
- `ContentId`
- `ContentTypeGroupId`
- `ContentTypeId`
- `ContentTypeIdentifier`
- `DateMetadata`
- `LanguageCode`
- `ObjectStateId`
- `ObjectStateIdentifier`
- `ParentLocationId`
- `RemoteId`
- `SectionId`
- `SectionIdentifier`
- `Sibling`
- `Subtree`
- `Visibility`
- `UserEmail`
- `UserId`
- `UserLogin`
- `UserMetadata`
- `IsUserBased`
- `IsUserEnabled`
- `LocationId`
- `LocationRemoteId`
- `Location\Depth`
- `Location\IsMainLocation`
- `Location\Priority`
- `LogicalAnd`
- `LogicalNot`
- `LogicalOr`
- `MatchAll`
- `MatchNone`

The difference with search here is that all Location Criteria are actually supported regardless of filtering Content or Locations. I don't see a lot of technical reasons (maybe performance a bit, you'll see nested DISTINCT magic in case of Content gateway implementation), not to allow this. For instance, a `Location\Depth` query for Content items will return only those Content items which have any Location satisfying the given depth criteria. The same goes for `Location\Priority`. Even `Location\IsMainLocation` can be found useful because for non-main Location requirement it will match only Content items which have more than one (so at least one non-main) Location.

##### Content

- `ContentId`
- `ContentName`
- `DateModified`
- `DatePublished`
- `SectionIdentifier`
- `SectionName`

##### Location

- `Depth`
- `Id`
- `Path`
- `Priority`
- `Visibility`

### Service Provider Interfaces

The architecture relies on query builders both for Criteria and Sort Clauses which follow Visitor pattern.

For a Criterion to support Repository Filtering, it must:
1. Implement the `\eZ\Publish\SPI\Repository\Values\Filter\FilteringCriterion` interface
2. Provide an implementation of `\eZ\Publish\SPI\Repository\Values\Filter\CriterionQueryBuilder` interface

For a Sort Clause to support Repository filtering, it must:
1. Implement the `\eZ\Publish\SPI\Repository\Values\Filter\FilteringSortClause` interface
2. Provide an implementation of `\eZ\Publish\SPI\Repository\Values\Filter\SortClauseQueryBuilder` interface.

### Implementation details

Implementation aims to be a separate module, so expect its parts mostly in dedicated namespaces prefixed either as `Filter` or `Filtering`.

One of most notable things is probably an existence of `\eZ\Publish\SPI\Persistence\Filter\Doctrine\FilteringQueryBuilder` which extends Doctrine DBAL Query Builder. It just felt right with the current architecture to add dedicated responsibility to query builder rather than  introducing separate service. Query Builder has a BC promise, because it is injected into Query Builder implementations of Criteria and Sort Clauses. While extending 3rd party implementation is often risky, I haven't found any indication that this component is either gonna change in the future or is intended to be internal.

The implementation relies on `JOIN`s combined with `WHERE` constraints for filtering data as this usually proves to be more efficient than `WHERE IN (subquery)` approach. Both handlers and a mapper were written specifically for this feature as the data returned by gateways are a bit different. It was also an opportunity to use current stack domain language in data set. Still, when it was clean and possible, existing mappers were used to map data.

#### Changes to known patterns

- `\eZ\Publish\API\Repository\Values\Content\ContentList` is a Value, but instead of controlling properties by the usual `ValueObject` magic to avoid overriding them, proper getter has been introduced.
- its constructor is `@internal` so it's clear that
```php
new ContentList()
```
is not supported outside of Kernel.

### TODO

- [x] Implement API for `LocationService`
- [x] Implement Criteria handlers
- [x] Implement Sort Clause handlers
- [x] Provide pagination support
- ~Provide caching support~ // not feasible ATM, needs to be a separate Spike
- ~Implement needed Criteria and Sort Clauses (TBD)~ // re-using Content Query Criteria with proper markers

### Follow-ups (tracked by the EZP-31711 Epic)

1. Support for the following Criteria:
   * MapLocationDistance
   * IsFieldEmpty
   * Field (maybe, TBD)
   * FieldRelation
2. Better pagination support - a Pagerfanta adapter
3. Caching, if possible.

### Conclusion

There are still things to improve, but this seems like a reasonable MVP, covering a lot of Criteria and Sort Clauses.

Any Reviewer who knows Repository a bit could see a lot of similarities between Legacy Search Engine and this implementation. So the natural question would be - why we didn't reuse more of internal structure of LSE? Or maybe even why we didn't simply expose LSE in a search settings-independent way? The reason for that was an attempt to make more robust implementation by relying more on `JOIN`s than `WHERE content.id IN (subquery)` pattern. As a side-effect we've got rid of LSE limitation when it comes to use some Location Criteria when searching for Content.

### QA
This is a PHP API that can be used to retrieve Content or Location item lists from Repository.
Example EE project which relies on volume testing data set can be found in [that diff](https://github.com/ezsystems/ezplatform-ee/compare/master...alongosz:ezp-31662-content-location-filtering-demo) and more specifically with [that sample Filtering query](https://github.com/ezsystems/ezplatform-ee/commit/06e345a5b4603ae3fae135a5519ec4ab5efd0c18).

#### Checklist:
- [x] Tests are implemented.

### DIFF
--- a/eZ/Bundle/EzPublishCoreBundle/ApiLoader/RepositoryFactory.php
+++ b/eZ/Bundle/EzPublishCoreBundle/ApiLoader/RepositoryFactory.php
@@ -17,6 +17,7 @@ use eZ\Publish\Core\Repository\User\PasswordHashServiceInterface;
 use eZ\Publish\Core\Repository\Helper\RelationProcessor;
 use eZ\Publish\Core\Repository\Mapper;
 use eZ\Publish\Core\Search\Common\BackgroundIndexer;
+use eZ\Publish\SPI\Persistence\Filter\Content\Handler as ContentFilteringHandler;
 use eZ\Publish\SPI\Persistence\Handler as PersistenceHandler;
 use eZ\Publish\SPI\Repository\Strategy\ContentThumbnail\ThumbnailStrategy;
 use eZ\Publish\SPI\Repository\Validator\ContentValidator;
@@ -84,7 +85,8 @@ class RepositoryFactory implements ContainerAwareInterface
         Mapper\ContentMapper $contentMapper,
         ContentValidator $contentValidator,
         LimitationService $limitationService,
-        PermissionService $permissionService
+        PermissionService $permissionService,
+        ContentFilteringHandler $contentFilteringHandler
     ): Repository {
         $config = $this->container->get('ezpublish.api.repository_configuration_provider')->getRepositoryConfig();
 
@@ -105,6 +107,7 @@ class RepositoryFactory implements ContainerAwareInterface
             $limitationService,
             $this->languageResolver,
             $permissionService,
+            $contentFilteringHandler,
             [
                 'role' => [
                     'policyMap' => $this->policyMap,
--- a/eZ/Bundle/EzPublishCoreBundle/ApiLoader/RepositoryFactory.php
+++ b/eZ/Bundle/EzPublishCoreBundle/ApiLoader/RepositoryFactory.php
@@ -18,6 +18,7 @@ use eZ\Publish\Core\Repository\Helper\RelationProcessor;
 use eZ\Publish\Core\Repository\Mapper;
 use eZ\Publish\Core\Search\Common\BackgroundIndexer;
 use eZ\Publish\SPI\Persistence\Filter\Content\Handler as ContentFilteringHandler;
+use eZ\Publish\SPI\Persistence\Filter\Location\Handler as LocationFilteringHandler;
 use eZ\Publish\SPI\Persistence\Handler as PersistenceHandler;
 use eZ\Publish\SPI\Repository\Strategy\ContentThumbnail\ThumbnailStrategy;
 use eZ\Publish\SPI\Repository\Validator\ContentValidator;
@@ -86,7 +87,8 @@ class RepositoryFactory implements ContainerAwareInterface
         ContentValidator $contentValidator,
         LimitationService $limitationService,
         PermissionService $permissionService,
-        ContentFilteringHandler $contentFilteringHandler
+        ContentFilteringHandler $contentFilteringHandler,
+        LocationFilteringHandler $locationFilteringHandler
     ): Repository {
         $config = $this->container->get('ezpublish.api.repository_configuration_provider')->getRepositoryConfig();
 
@@ -108,6 +110,7 @@ class RepositoryFactory implements ContainerAwareInterface
             $this->languageResolver,
             $permissionService,
             $contentFilteringHandler,
+            $locationFilteringHandler,
             [
                 'role' => [
                     'policyMap' => $this->policyMap,
--- a/eZ/Bundle/EzPublishCoreBundle/DependencyInjection/EzPublishCoreExtension.php
+++ b/eZ/Bundle/EzPublishCoreBundle/DependencyInjection/EzPublishCoreExtension.php
@@ -587,9 +587,9 @@ class EzPublishCoreExtension extends Extension implements PrependExtensionInterf
             );
 
         $container->registerForAutoconfiguration(FilteringCriterionQueryBuilder::class)
-            ->addTag(FilteringCriterionQueryBuilder::SYMFONY_TAG_NAME);
+            ->addTag(ServiceTags::FILTERING_CRITERION_QUERY_BUILDER);
 
         $container->registerForAutoconfiguration(FilteringSortClauseQueryBuilder::class)
-            ->addTag(FilteringSortClauseQueryBuilder::SYMFONY_TAG_NAME);
+            ->addTag(ServiceTags::FILTERING_SORT_CLAUSE_QUERY_BUILDER);
     }
 }
--- a/eZ/Bundle/EzPublishCoreBundle/DependencyInjection/EzPublishCoreExtension.php
+++ b/eZ/Bundle/EzPublishCoreBundle/DependencyInjection/EzPublishCoreExtension.php
@@ -141,18 +141,7 @@ class EzPublishCoreExtension extends Extension implements PrependExtensionInterf
 
         $this->buildPolicyMap($container);
 
-        $container->registerForAutoconfiguration(QueryType::class)
-            ->addTag(QueryTypePass::QUERY_TYPE_SERVICE_TAG);
-
-        $container->registerForAutoconfiguration(ConfigScopeChangeSubscriber::class)
-                  ->addTag(
-                      'kernel.event_listener',
-                      ['method' => 'onConfigScopeChange', 'event' => MVCEvents::CONFIG_SCOPE_CHANGE]
-                  )
-                  ->addTag(
-                      'kernel.event_listener',
-                      ['method' => 'onConfigScopeChange', 'event' => MVCEvents::CONFIG_SCOPE_RESTORE]
-                  );
+        $this->registerForAutoConfiguration($container);
     }
 
     /**
@@ -578,5 +567,21 @@ class EzPublishCoreExtension extends Extension implements PrependExtensionInterf
         if ($container->getParameter('ezpublish.url_wildcards.enabled')) {
             $loader->load('url_wildcard.yml');
         }
+    }
+
+    private function registerForAutoConfiguration(ContainerBuilder $container): void
+    {
+        $container->registerForAutoconfiguration(QueryType::class)
+            ->addTag(QueryTypePass::QUERY_TYPE_SERVICE_TAG);
+
+        $container->registerForAutoconfiguration(ConfigScopeChangeSubscriber::class)
+            ->addTag(
+                'kernel.event_listener',
+                ['method' => 'onConfigScopeChange', 'event' => MVCEvents::CONFIG_SCOPE_CHANGE]
+            )
+            ->addTag(
+                'kernel.event_listener',
+                ['method' => 'onConfigScopeChange', 'event' => MVCEvents::CONFIG_SCOPE_RESTORE]
+            );
     }
 }
--- a/eZ/Bundle/EzPublishCoreBundle/DependencyInjection/EzPublishCoreExtension.php
+++ b/eZ/Bundle/EzPublishCoreBundle/DependencyInjection/EzPublishCoreExtension.php
@@ -18,6 +18,8 @@ use eZ\Bundle\EzPublishCoreBundle\SiteAccess\SiteAccessConfigurationFilter;
 use eZ\Publish\Core\MVC\Symfony\MVCEvents;
 use eZ\Publish\Core\QueryType\QueryType;
 use eZ\Publish\SPI\MVC\EventSubscriber\ConfigScopeChangeSubscriber;
+use eZ\Publish\SPI\Repository\Values\Filter\CriterionQueryBuilder as FilteringCriterionQueryBuilder;
+use eZ\Publish\SPI\Repository\Values\Filter\SortClauseQueryBuilder as FilteringSortClauseQueryBuilder;
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
 use Symfony\Component\Filesystem\Filesystem;
@@ -583,5 +585,11 @@ class EzPublishCoreExtension extends Extension implements PrependExtensionInterf
                 'kernel.event_listener',
                 ['method' => 'onConfigScopeChange', 'event' => MVCEvents::CONFIG_SCOPE_RESTORE]
             );
+
+        $container->registerForAutoconfiguration(FilteringCriterionQueryBuilder::class)
+            ->addTag(FilteringCriterionQueryBuilder::SYMFONY_TAG_NAME);
+
+        $container->registerForAutoconfiguration(FilteringSortClauseQueryBuilder::class)
+            ->addTag(FilteringSortClauseQueryBuilder::SYMFONY_TAG_NAME);
     }
 }
--- a/eZ/Bundle/EzPublishCoreBundle/DependencyInjection/ServiceTags.php
+++ b/eZ/Bundle/EzPublishCoreBundle/DependencyInjection/ServiceTags.php
@@ -1,0 +1,27 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace eZ\Bundle\EzPublishCoreBundle\DependencyInjection;
+
+/**
+ * "Enum" for the Symfony Service tag names provided by the Extension.
+ */
+class ServiceTags
+{
+    /**
+     * Auto-configured tag name for
+     * {@see \eZ\Publish\SPI\Repository\Values\Filter\CriterionQueryBuilder}.
+     */
+    public const FILTERING_CRITERION_QUERY_BUILDER = 'ezplatform.filter.criterion.query_builder';
+
+    /**
+     * Auto-configured tag name for
+     * {@see \eZ\Publish\SPI\Repository\Values\Filter\SortClauseQueryBuilder}.
+     */
+    public const FILTERING_SORT_CLAUSE_QUERY_BUILDER = 'ezplatform.filter.sort_clause.query_builder';
+}
--- a/eZ/Bundle/EzPublishCoreBundle/Tests/DependencyInjection/EzPublishCoreExtensionTest.php
+++ b/eZ/Bundle/EzPublishCoreBundle/Tests/DependencyInjection/EzPublishCoreExtensionTest.php
@@ -10,6 +10,7 @@ use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Compiler\QueryTypePass;
 use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\Parser\Common;
 use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\Parser\Content;
 use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\EzPublishCoreExtension;
+use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\ServiceTags;
 use eZ\Bundle\EzPublishCoreBundle\Tests\DependencyInjection\Stub\Filter\CustomCriterionQueryBuilder;
 use eZ\Bundle\EzPublishCoreBundle\Tests\DependencyInjection\Stub\Filter\CustomSortClauseQueryBuilder;
 use eZ\Bundle\EzPublishCoreBundle\Tests\DependencyInjection\Stub\QueryTypeBundle\QueryType\TestQueryType;
@@ -824,12 +825,12 @@ class EzPublishCoreExtensionTest extends AbstractExtensionTestCase
     {
         yield Filter\CriterionQueryBuilder::class => [
             CustomCriterionQueryBuilder::class,
-            Filter\CriterionQueryBuilder::SYMFONY_TAG_NAME,
+            ServiceTags::FILTERING_CRITERION_QUERY_BUILDER,
         ];
 
         yield Filter\SortClauseQueryBuilder::class => [
             CustomSortClauseQueryBuilder::class,
-            Filter\SortClauseQueryBuilder::SYMFONY_TAG_NAME,
+            ServiceTags::FILTERING_SORT_CLAUSE_QUERY_BUILDER,
         ];
     }
 
--- a/eZ/Bundle/EzPublishCoreBundle/Tests/DependencyInjection/EzPublishCoreExtensionTest.php
+++ b/eZ/Bundle/EzPublishCoreBundle/Tests/DependencyInjection/EzPublishCoreExtensionTest.php
@@ -10,8 +10,11 @@ use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Compiler\QueryTypePass;
 use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\Parser\Common;
 use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\Parser\Content;
 use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\EzPublishCoreExtension;
+use eZ\Bundle\EzPublishCoreBundle\Tests\DependencyInjection\Stub\Filter\CustomCriterionQueryBuilder;
+use eZ\Bundle\EzPublishCoreBundle\Tests\DependencyInjection\Stub\Filter\CustomSortClauseQueryBuilder;
 use eZ\Bundle\EzPublishCoreBundle\Tests\DependencyInjection\Stub\QueryTypeBundle\QueryType\TestQueryType;
 use eZ\Bundle\EzPublishCoreBundle\Tests\DependencyInjection\Stub\StubPolicyProvider;
+use eZ\Publish\SPI\Repository\Values\Filter;
 use Matthias\SymfonyDependencyInjectionTest\PhpUnit\AbstractExtensionTestCase;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Compiler\CheckExceptionOnInvalidReferenceBehaviorPass;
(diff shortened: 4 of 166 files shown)


andrerom:
> Productboard spec suggested this method to be called filter, ...

`filter` was never a good name, point was to avoid that people think this is search _(find is sometimes used for search, we for sure do in SearchService)_. But it's not a big deal, the object given is different by name at least.

> `ContentList` .. mark it internal

Then maybe document find to also return `Content[]` ? Or is there some other way we can doc hint that ContentList is collection of Content items?

> mentioned spec also suggested array $languageSettings as a 2nd argument. I'm not 

That was probably just to stay close to how it is everywhere else, however putting it on Filter is fine. Maybe we should do the same on Query then.

> To self: would a generic Filter (not Content and Location specific) be "implementable"?

It's better if it is specific, so we avoid the Content vs Location mess we have with Query. With no clear type hint on what goes where, only known on run time.

> Which Criteria and Sort Clauses are needed for MVP?

Ideally everything, except:
- Search features _(Fulltext, CustomField, ...)_
- Criteria/SortClauses with performance issues on SQL making it impossible to support them with this interface as we can't just say "Use Solr" _(Field)_

> Repository Query Builder" alternative: (pending)

Interesting

bdunogier:
About the language parameters:

> $filter->languageCodes = ['eng-GB']; // empty/not set for all (?)

Actually it should respect the siteaccess aware repository approach (in the SA aware implementation), and return items depending on the current SA's language settings. How did the repo behave before we added that again ?

andrerom:
> What about visibility ? I'd expect that by default, invisible items aren't returned on the frontend (and can be set to be returned), and are returned on the backend (and can be set to not be returned). How could we manage that ?

+20 but this should ideally be done Repo wide for all API's. Either as SA aware layer injects config for you on hidden or not for instance. Or we move this to permissions _(with something als SALimitation("admin"))_ to get permission system to do this for us.

bdunogier:
> Either as SA aware layer injects config for you on hidden or not for instance.

By default, I assumed (I should not) that it follows the same architecture than the rest of the repository, with that extra siteaccess aware layer (transparent). Can you confirm @alongosz ?
If yes, would that layer modify the query to add the language to the filter ? It sounds a bit odd, but it would work afaict.

> Basically it works by setting value from config if languageCodes argument is null (not set by API consumer), so it can work with this approach as well.

Good. Let's do this for that API endpoint as well. If no language is specified, it uses the current SA's languages parameters. OK @alongosz ?

alongosz:
Thanks @andrerom and @bdunogier for your comments.

> `filter` was never a good name, point was to avoid that people think this is search _(find is sometimes used for search, we for sure do in SearchService)_. But it's not a big deal, the object given is different by name at least.

This is exactly the approach now - `$contentService->find` should not be easily mistaken for SearchService, but I'm still open to other suggestions. We can always use `fetch` to make Legacy users happy ;)

> 
> > `ContentList` .. mark it internal
> 
> Then maybe document find to also return `Content[]` ? Or is there some other way we can doc hint that ContentList is collection of Content items?

It's already done via:
```php
   /**
     * @return \eZ\Publish\API\Repository\Values\Content\Content[]|\Traversable
     */
    public function getIterator(): Traversable;
```

> 
> > mentioned spec also suggested array $languageSettings as a 2nd argument. I'm not
> 
> That was probably just to stay close to how it is everywhere else, however putting it on Filter is fine. Maybe we should do the same on Query then.

+1

> 
> > To self: would a generic Filter (not Content and Location specific) be "implementable"?
> 
> It's better if it is specific, so we avoid the Content vs Location mess we have with Query. With no clear type hint on what goes where, only known on run time.

TBD, I might indeed do this to provide strict API

> 
> > Which Criteria and Sort Clauses are needed for MVP?
> 
> Ideally everything, except:
> 
> * Search features _(Fulltext, CustomField, ...)_
> * Criteria/SortClauses with performance issues on SQL making it impossible to support them with this interface as we can't just say "Use Solr" _(Field)_

Ok, here comes the list, with questions and remarks:

### Criteria:
* `Logical\LogicalAnd`, `Logical\LogicalOr`, `Logical\LogicalNot` // keywords in use, hence the redundancy in namespace\name.
* `Content\Id`
* `Content\RemoteId`
* `Content\Visibilty` // for "hidden" feature
* `ContentType\GroupId`
* `ContentType\Id`
* `ContentType\Identfier`
* `DateMetadata`
* `Field\Relation` // maybe, TBD
* `Field\IsFieldEmpty` // maybe, don't remember if it requires external storage
* `LanguageCode`
* `Location\Id`
* `Location\Depth`
* `Location\IsMainLocation`
* `Location\ParentId`
* `Location\ParentRemoteId` // I really miss that in Search
* `Location\Priority`
* `Location\RemoteId`
* `Location\Sibling`
* `Location\Subtree`
* `Location\Visibility` // in Search known just as Visibility, but I have the other idea for that one
* `MatchNone` // (`MatchAll` doesn't make much sense ATM to me, because this is default behavior of any SQL engine)
* `ObjectState\Id`
* `ObjectState\Identifier`
* `Section\Id`
* `Section\Identifier`
* `User\Email`
* `User\Id`
* `User\Login`
* `User\Metadata`
* `User\IsUserBased`
* `User\IsUserEnabled`
* `Version\Status` // new one to filter out published/drafts/archived
* `Version\No` // new one to filter out Version number
* `Version\Id` // new one, should be used for fetched data, not hardcoded, not sure if necessary though
* `Visibility` // a Composite on (`Content\Visibility` or `Location\Visibility`) - do we need it?

### SortClauses:
* `Content\Id` // does it really bring some value, maybe it should be default for Content filtering if none given[1]?
* `Content\Name`
* `Date\Modified`
* `Date\Trashed` // as a follow-up, might be useful for searching in Trash Story // cc @SylvainGuittard
* `Date\Published`
* `Location`
* `Location\Depth`
* `Location\Id` // does it really bring some value, maybe it should be default for Location filtering if none given[1]?
* `Location\IsMainLocation`
* `Location\Path`
* `Location\Priority`
* `Location\Visibility` // does it bring some value to have hidden/not hidden elements first? AFAICS it's not even used by LSE (SQL Search)
* `Section\Identifier`
* `Section\Name`

Note that consumption would be a bit different, more strict. This would be actually DX advantage, because usage with IDE should be more intuitive.

The list is huge, so we should focus on the most important ones and deliver the remaining ones as a follow-up.

_[1] IMHO we should have some fallback `ORDER BY` because otherwise order is SQL engine-dependent and sometimes non-deterministic._

> > Either as SA aware layer injects config for you on hidden or not for instance.
> 
> By default, I assumed (I should not) that it follows the same architecture than the rest of the repository, with that extra siteaccess aware layer (transparent). Can you confirm @alongosz ?

SA layer is enabled by default now, so some language always will be defined.

> If yes, would that layer modify the query to add the language to the filter ? It sounds a bit odd, but it would work afaict.

 It should set languages only if not set (`empty($filter->languageCodes) === true`). 

> 
> > Basically it works by setting value from config if languageCodes argument is null (not set by API consumer), so it can work with this approach as well.
> 
> Good. Let's do this for that API endpoint as well. If no language is specified, it uses the current SA's languages parameters.

+0.75 - I would like to avoid distinction between `[]` and `null` as it's confusing even for our Engineers. Let's stick with `[]`, hence `empty` above.

> What about **query types** ? Can we somehow re-use those ?

They're strictly typed for Search API.

> Also, what's the matter if the one in API is an interface ? It's not meant to be instantiated by our API consumers, is it ?

That's exactly the point of having interface on API for strict referencing in 3rd party code w/o possibility to instantiate.

> What about **visibility** ? I'd expect that by default, invisible items aren't returned on the frontend (and can be set to be returned), and are returned on the backend (and can be set to not be returned). How could we manage that ?

There's no way to distinguish front-end from backend. I've proposed 3 separate Criteria
* `Content\Visibility`
* `Location\Visibility`
* `Visibility` // Content OR Location

The same goes for Version status, though we can probably assume that if someone asks for `Content\*`, but does not specify Version\No and/or Status, published one is asumed.
Due to the nature of SQL each Criterion will have atomic responsibility. We can manipulate query when pre-processing it to attach some default Criteria if not given.

bdunogier:
> (visibility) There's no way to distinguish front-end from backend. I've proposed 3 separate Criteria

Yes there is: the siteaccess. In legacy, we had a setting for showing invisible objects, enabled for backoffice, and disabled for front. It would be a perfect task for the SA aware repository layer (add the criterion if it wasn't added by the user, as a default.

>> What about query types ? Can we somehow re-use those ?
> They're strictly typed for Search API.

Yes, I know that :)
The question isn't really technical, to be honest. The fact is that we added query types to simplify the Search API usage. We are now adding another API endpoint, very similar to search (criteria, sort clauses...), but without the simplification layer.

It is even more important now that we have added built-in types. I really insist that we try to make that consistent, even though I don't know how right now. In the end, 90% of what a query type returns can be used for other find operations (filter + sort + pagination).

alongosz:
Internal sync:
* Due to huge amount of Criteria for now we've decided to try and re-use existing Criteria namespace.
* Those Criteria that supported by Search only would just throw and exception.
* However we'll gonna add marker interfaces for Content/Location filtering-capable Criteria.
* Criterion handling will be delegated to a Content/Location filtering specific engine (implementations).
* We'll need to update extensibility documentation for custom Criteria
* The same goes for SortClauses.
* For now the first implementation will return published only items, with an option expand further to implement `Version\*` Criteria.
* Pagination: solution should be flexible enough to support different ways of pagination (offset & limit, cursor-based).

src/lib/SPI/Repository/Value/Filter/FilteringCriterion.php
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace Ibexa\Platform\SPI\Repository\Value\Filter;
+
+/**
+ * Marker for Content & Location filtering Criterion.
```
bdunogier, line 12: Does it mean a Criterion that supports both should only implement that interface ? What is its actual meaning ?

src/lib/API/Repository/Value/Filter/Filter.php
```diff
@@ -0,0 +1,99 @@
...
+        return $this;
+    }
+
+    /**
+     * @see \Ibexa\Platform\API\Repository\Value\Filter\Filter::withCriterion
+     */
+    public function andWithCriterion(Criterion $criterion): Filter
+    {
+        $this->criterion = null !== $this->criterion
+            ? new Criterion\LogicalAnd([$this->criterion, $criterion])
```
bdunogier, line 71: If `$this->criterion` is already an instance of `Criterion\LogicalAnd`, it should add the given criterion to it.

src/lib/SPI/Repository/Value/Filter/FilteringCriterion.php
alongosz: > Does it mean a Criterion that supports both should only implement that interface ? What is its actual meaning ?

That's the initial idea - if it supports both, it should implement that interface. 
But I need to apply this to all Criteria to see if keeping this common interface makes sense.

src/lib/API/Repository/Value/Filter/Filter.php
alongosz: > If `$this->criterion` is already an instance of `Criterion\LogicalAnd`, it should add the given criterion to it.

yeah, I felt there will be an issue with this simple expression. Actually working on unit tests now to visualize better the way complex Filter is built.

src/lib/SPI/Repository/Value/Filter/FilteringCriterion.php
andrerom: @alongosz A bit late to the party, but I did start at creating own separate Persistence Criteria some time back, for the purpose of usage in Trash filtering as well as for the methods proposed here.

Besides separation, another reason for doing so was to experiment with rather using joins instead of subqueries to make sure we don't hit performance issues. Even if newer mysql/mariadb versions are supposed to optimize this they seem to have plenty of cases where they won't be able to.

https://github.com/ezsystems/ezpublish-kernel/commit/90c954f52bc6564508db642522a137c4bfe96abf  _(see Query/Readme.md at the end of diff for overview)_

alongosz: > A bit late to the party, but FYI started at creating own separate Persistence Criteria some time back, for the purpose of usage in Trash filtering
> (...)
> [ezsystems/ezpublish-kernel@90c954f](https://github.com/ezsystems/ezpublish-kernel/commit/90c954f52bc6564508db642522a137c4bfe96abf)

AFAICS the commit is about Criterion handlers, not the Criteria themselves, right? If you planned to add separate Criteria also, then this wouldn't be the best direction IMHO. Criterion itself should not be coupled with any storage. It can be coupled with persistence abstraction for sure, but I have a feeling our persistence abstraction is too coupled with Legacy Storage TBH ;)

> Besides separation, another reason for doing so was to experiment with rather using joins instead of subqueries to make sure we don't hit performance issues. Even if newer mysql/mariadb versions are supposed to optimize subqueries, they seem to have plenty of cases where they won't be able to.

Criterion Handling is going to be separate here for Content/Location filtering, so it can utilize joins when needed, but only when underlying schema makes it correct (e.g.: I'd like to avoid programmatic cleanup of redundant data).

src/lib/API/Repository/Value/Filter/Filter.php
```diff
@@ -0,0 +1,155 @@
...
+     *
+     * To re-build Criterion from scratch `reset` it first.
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException if Criterion is already set
+     *
+     * @see reset
+     * @see andWithCriterion
+     * @see orWithCriterion
+     * @see \eZ\Publish\API\Repository\Values\URL\Query\Criterion\LogicalOr
+     * @see \eZ\Publish\API\Repository\Values\URL\Query\Criterion\LogicalAnd
```
adamwojs, line 84: Did you mean `\eZ\Publish\API\Repository\Values\Content\Query\Criterion\LogicalOr` and `\eZ\Publish\API\Repository\Values\Content\Query\Criterion\LogicalAnd` ?

src/lib/API/Repository/Value/Filter/Filter.php
```diff
@@ -0,0 +1,155 @@
...
+            // for better DX allow operation on uninitialized Criterion by setting it as-is
+            $this->criterion = $criterion;
+        } elseif ($this->criterion instanceof Criterion\LogicalAnd) {
+            $this->criterion->criteria[] = $criterion;
+        } else {
+            $this->criterion = new Criterion\LogicalAnd([$this->criterion, $criterion]);
+        }
+
+        return $this;
+    }
```
adamwojs, line 117: Q: Could we split this class into `Filter` and and `FilterBuilder`? 

* `Filter` will be VO (inc. immutable)
* `FilterBuilder` could be an interface and concrete implementation could be hidden in Core namespace
* Better responsibilities separation

tests/lib/unit/API/Repository/Value/Filter/FilterTest.php
```diff
@@ -0,0 +1,323 @@
...
+                        $parent2Criterion,
+                    ]
+                ),
+                $gerDECriterion,
+            ]
+        );
+
+        $filter = clone $filter;
+        $filter
+            ->reset()
```
adamwojs, line 310: This lines could be simpliied to "new Filter()"

tests/lib/unit/API/Repository/Value/Filter/FilterTest.php
```diff
@@ -0,0 +1,323 @@
...
+            self::assertNotEmpty($filter->getCriterion());
+        }
+        self::assertNotEmpty($filterWithSortClause->getSortClauses());
+
+        foreach (func_get_args() as $filter) {
+            $filter->reset();
+            self::assertEmpty($filter->getCriterion());
+            self::assertEmpty($filter->getSortClauses());
+        }
+    }
```
adamwojs, line 227: Sorry, but this implementation is over engineered. KISS:

A. Create Filter with criterion and sort clauses (simply use constructor).
B. Call `reset()` method
C. Assert if fields has been reset

tests/lib/unit/API/Repository/Value/Filter/FilterTest.php
```diff
@@ -0,0 +1,323 @@
...
+            new SortClause\ContentName(Query::SORT_DESC),
+        ];
+        $filter = new Filter();
+        $filter
+            ->withCriterion($criterion->criteria[0])
+            ->orWithCriterion($criterion->criteria[1])
+            ->withSortClause($sortClauses[0])
+            ->withSortClause($sortClauses[1]);
+
+        yield 'parent1 AND language1 OR parent2 AND language2' => [
```
adamwojs, line 283:
```suggestion
        yield '(parent=1 AND language=eng-GB) OR (parent=2 AND language=ger-DE)' => [
```

tests/lib/unit/API/Repository/Value/Filter/FilterTest.php
```diff
@@ -0,0 +1,323 @@
...
+        $criterion1 = new Criterion\ContentId(1);
+        $criterion2 = new Criterion\RemoteId(md5('/1/2/3/'));
+        $criterion3 = new Criterion\Ancestor('/1/2/');
+
+        $filter = new Filter();
+        $filter->withCriterion($criterion1);
+
+        $filter->andWithCriterion($criterion2);
+        $expectedCriterion = new Criterion\LogicalAnd([$criterion1, $criterion2]);
+        self::assertEquals($expectedCriterion, $filter->getCriterion());
```
adamwojs, line 137: Hmmm AFAICS we rather use `$this->assertEquals(...)` (727 occurrences) syntax instead of `self::assertEquals(...)` (131 occurrences) in `\eZ\Publish\API\Repository\Tests\`...

tests/lib/unit/API/Repository/Value/Filter/FilterTest.php
```diff
@@ -0,0 +1,323 @@
...
+use eZ\Publish\API\Repository\Exceptions\BadStateException;
+use eZ\Publish\API\Repository\Values\Content\Query;
+use eZ\Publish\API\Repository\Values\Content\Query\Criterion;
+use eZ\Publish\API\Repository\Values\Content\Query\SortClause;
+use eZ\Publish\API\Repository\Values\URL\Query\SortClause as URLQuerySortClause;
+use Ibexa\Platform\API\Repository\Value\Filter\Filter;
+use PHPUnit\Framework\TestCase;
+use function md5;
+
+class FilterTest extends TestCase
```
adamwojs, line 20:
```suggestion
final class FilterTest extends TestCase
```

tests/lib/integration/API/Repository/Tests/ContentService/ContentFilteringTest.php
```diff
@@ -0,0 +1,186 @@
...
+            // expected names
+            [
+                'eng-GB' => ['Nested English Folder 1', 'Nested English Folder 2'],
+                'eng-US' => ['Nested American Folder 1', 'Nested American Folder 2'],
+            ],
+            // expected total count
+            2,
+        ];
+
+        // ---------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
alongosz: > Why do you need this ???

Readability. While it's useful for building data step by step, it's difficult to distinguish separate test cases. We probably can go back to returning the whole data set at once if it's too confusing :)

src/lib/API/Repository/Value/Filter/Filter.php
alongosz: > Q: Could we split this class into `Filter` and and `FilterBuilder`?
> 
> * `Filter` will be VO (inc. immutable)
> * `FilterBuilder` could be an interface and concrete implementation could be hidden in Core namespace
> * Better responsibilities separation

I don't mind, WDYT @bdunogier? :)

tests/lib/unit/API/Repository/Value/Filter/FilterTest.php
alongosz: > This lines could be simpliied to "new Filter()"

AFAIR the point here was to test resetting after cloning.

tests/lib/unit/API/Repository/Value/Filter/FilterTest.php
alongosz: > Sorry, but this implementation is over engineered. KISS:
> 
> A. Create Filter with criterion and sort clauses (simply use constructor).
> B. Call `reset()` method
> C. Assert if fields has been reset

What if the behavior of the constructor started to differ from the behavior of the methods? We wouldn't detect breaking change. But I could simplify this test to create a separate Filter with all the calls. It just felt kinda nice to re-use previous tests results instead of wasting more memory and op. time ;)

tests/lib/unit/API/Repository/Value/Filter/FilterTest.php
alongosz: > Hmmm AFAICS we rather use `$this->assertEquals(...)` (727 occurrences) syntax instead of `self::assertEquals(...)` (131 occurrences) in `\eZ\Publish\API\Repository\Tests\`...

I believe calling a static method using `$this` is wrong, though indeed PHPUnit doc says "use whatever".

src/lib/SPI/Repository/Value/Filter/FilteringCriterion.php
andrerom: > AFAICS the commit is about Criterion handlers, not the Criteria themselves, right? If you planned to add separate Criteria also

yes, indeed own set of Criteria would be one of several followup there. But not due to coupling as you jump to conclusion on, rather:
- proper type hints on the new api's on what is allowed to use and not, like Adam already stated to do on URLService for the same reasons.

Beyond that rethink how CriteriaHandlers are done to see if we could make them use join s instead.

Anyway, good luck.

src/lib/API/Repository/Value/Filter/Filter.php
bdunogier: Hmm, it's not a builder, it's an actual value object. The way I suggested it, `with*` methods are fluent setters, because they can manipulate an immutable of their type (clone self, modify the clone, and return the clone).

There isn't enough flesh for a builder imho. A "real" builder would be "dynamic", like the one I had shown a while ago.

bdunogier:
Could you try to find some time to update the description so that it matches the last decisions ? It would really make reviews and feedback easier.

alongosz:
@bdunogier should be up to date now with the current state after rebasing, but I already see handling of languages needs to be changed to follow known pattern used in SiteAccess-aware services layer. Moreover Filter vs. FilterBuilder is still unresolved.
Reverted Ibexa namespace changes as the related PR was declined.

alongosz:
Internal sync: removed `createFindFilter` from `LocationService` and `ContentService`. `new Filter` is the only way now.

alongosz:
Internal sync: Moved Service Tags constants to dedicated ServiceTags class (30f125b9c).

alongosz:
Rebased to resolve conflicts after another merge and squashed fixups.

alongosz:
Merging, QA will be done on the RC tag.

PR #57, opened 2020-04-28
EZP-31595: Added embedded permission check in UrlAliasRouter

barw4:
| **JIRA issue**                          | [EZP-31595](https://jira.ez.no/browse/EZP-31595)
| **Target eZ Platform version** | v3.0

#### Checklist:
- [x] Tests are implemented.

### DIFF
--- a/eZ/Bundle/EzPublishCoreBundle/Resources/views/default/content/embed.html.twig
+++ b/eZ/Bundle/EzPublishCoreBundle/Resources/views/default/content/embed.html.twig
@@ -1,7 +1,6 @@
 {% set content_name=ez_content_name(content) %}
 
-{% if location is defined %}
-    <h3><a href="{{ ez_path(location) }}">{{ content_name }}</a></h3>
-{% else %}
-    <h3><a href="{{ ez_path(content) }}">{{ content_name }}</a></h3>
-{% endif %}
+{% set parameters = {
+    "isEmbed": true
+} %}
+<h3><a href="{{ ez_path(content, parameters) }}">{{ content_name }}</a></h3>
--- a/eZ/Bundle/EzPublishCoreBundle/Resources/views/default/content/embed.html.twig
+++ b/eZ/Bundle/EzPublishCoreBundle/Resources/views/default/content/embed.html.twig
@@ -3,4 +3,8 @@
 {% set parameters = {
     "isEmbed": true
 } %}
-<h3><a href="{{ ez_path(content, parameters) }}">{{ content_name }}</a></h3>
+{% if location is defined %}
+    <h3><a href="{{ ez_path(location, parameters) }}">{{ content_name }}</a></h3>
+{% else %}
+    <h3><a href="{{ ez_path(content, parameters) }}">{{ content_name }}</a></h3>
+{% endif %}
--- a/eZ/Publish/Core/MVC/Symfony/Routing/UrlAliasRouter.php
+++ b/eZ/Publish/Core/MVC/Symfony/Routing/UrlAliasRouter.php
@@ -7,11 +7,15 @@
 namespace eZ\Publish\Core\MVC\Symfony\Routing;
 
 use eZ\Publish\API\Repository\LocationService;
+use eZ\Publish\API\Repository\PermissionResolver;
+use eZ\Publish\API\Repository\Repository;
 use eZ\Publish\API\Repository\URLAliasService;
 use eZ\Publish\API\Repository\ContentService;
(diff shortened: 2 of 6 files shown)


eZ/Bundle/EzPublishCoreBundle/Resources/views/default/content/embed.html.twig
```diff
@@ -1,7 +1,6 @@
...
 
-{% if location is defined %}
-    <h3><a href="{{ ez_path(location) }}">{{ content_name }}</a></h3>
-{% else %}
-    <h3><a href="{{ ez_path(content) }}">{{ content_name }}</a></h3>
-{% endif %}
+{% set parameters = {
+    "isEmbed": true
+} %}
+<h3><a href="{{ ez_path(content, parameters) }}">{{ content_name }}</a></h3>
```
andrerom, line 6: This risks linking to main location when usage expected it to link to one of the other locations afaik, at least that was probably the intention of making sure to check for specified location variable.

eZ/Bundle/EzPublishCoreBundle/Resources/views/default/content/embed.html.twig
```diff
@@ -1,7 +1,10 @@
...
 
+{% set parameters = {
+    "isEmbed": true
+} %}
 {% if location is defined %}
-    <h3><a href="{{ ez_path(location) }}">{{ content_name }}</a></h3>
+    <h3><a href="{{ ez_path(location, parameters) }}">{{ content_name }}</a></h3>
 {% else %}
-    <h3><a href="{{ ez_path(content) }}">{{ content_name }}</a></h3>
+    <h3><a href="{{ ez_path(content, parameters) }}">{{ content_name }}</a></h3>
```
andrerom, line 9: Shouldn't this rather skip generating link if user does not have access?


As in:
- The link will always go to full view, which implies user needs to have full access
- So rather then doing the checks here in the router, the approach is maybe rather to check rights when rendering embed and pass variable to template to let it skip link if user does not have access to full view  _(if template is rendered it's already implied user has access to see it, at least embed view of it)_

barw4: > Shouldn't this rather skip generating link if user does not have access?

@andrerom I'm not really sure right now. The links are present also in other places in the app even when the user does not have access to linked content, so that's why I decided to kinda go with this solution. Should I change it for the solution you mentioned?

andrerom: I think so, otherwise we are just pushing the issue elsewhere as the user will click on a link and get exception. But maybe @alongosz has some perspective on this as well

eZ/Publish/Core/MVC/Symfony/Routing/UrlAliasRouter.php
```diff
@@ -64,13 +73,16 @@ public function __construct(
...
+        Repository $repository,
         RequestContext $requestContext,
         LoggerInterface $logger = null
     ) {
         $this->locationService = $locationService;
         $this->urlAliasService = $urlAliasService;
         $this->contentService = $contentService;
         $this->generator = $generator;
+        $this->repository = $repository;
+        $this->permissionResolver = $this->repository->getPermissionResolver();
```
webhdx, line 85: Never do this. Pass `PermissionResolver` as another dependency and let container builder to inject it.

eZ/Publish/Core/MVC/Symfony/Routing/UrlAliasRouter.php
```diff
@@ -284,12 +296,14 @@ public function getRouteCollection()
      * @param array $parameters An array of parameters
      * @param int $referenceType The type of reference to be generated (one of the constants)
      *
-     * @throws \LogicException
-     * @throws \Symfony\Component\Routing\Exception\RouteNotFoundException
-     * @throws \InvalidArgumentException
-     *
      * @return string The generated URL
      *
+     * @throws NotFoundException
```
webhdx, line 301: FQCN

eZ/Publish/Core/MVC/Symfony/Routing/UrlAliasRouter.php
Nattfarinn: +1, considering PermissionResolver is already decoupled and available as a service.

barw4:
Closing this PR as Router should be left intact in this case.

PR #60, opened 2020-05-05
[Tests] Fixed obsolete PHPUnit class names

alongosz:
| **Target version** | test setup for eZ Platform Kernel 1.0

We still had used some PHPUnit class names from `\PHPUnit_Framework_*` which were finally dropped in PHPUnit 8.

It applied to `MockObject` usage and exception expectations. For the latter one a more correct fix was due, which simplified our code a bit. While at it, some missing exception-related imports were fixed.

#### Checklist:
- [x] Tests are fixed.
- [x] Travis is passing.

### DIFF
--- a/eZ/Publish/API/Repository/Tests/FieldType/BaseIntegrationTest.php
+++ b/eZ/Publish/API/Repository/Tests/FieldType/BaseIntegrationTest.php
@@ -6,13 +6,13 @@
  */
 namespace eZ\Publish\API\Repository\Tests\FieldType;
 
+use eZ\Publish\API\Repository\Exceptions\ContentTypeFieldDefinitionValidationException;
+use eZ\Publish\API\Repository\Exceptions\NotFoundException;
 use eZ\Publish\API\Repository\Tests;
 use eZ\Publish\API\Repository;
 use eZ\Publish\API\Repository\Values\Content\Content;
 use eZ\Publish\API\Repository\Values\Content\Field;
 use eZ\Publish\API\Repository\Values\ContentType\FieldDefinition;
-use Exception;
-use PHPUnit_Framework_AssertionFailedError;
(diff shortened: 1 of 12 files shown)


eZ/Publish/API/Repository/Tests/FieldType/BaseIntegrationTest.php
```diff
@@ -923,7 +923,7 @@ public function testCopiedExternalData(Field $field)
      */
```
mikadamczyk, line 923: One line above it is another `@dep_ends`

eZ/Publish/SPI/FieldType/Tests/FieldTypeTest.php
```diff
@@ -603,34 +602,19 @@ public function testAcceptGetEmptyValue()
 
     /**
      * @param mixed $inputValue
-     * @param \Exception $expectedException
      *
      * @dataProvider provideInvalidInputForAcceptValue
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
```
mikadamczyk, line 608: If we check the occurrence of this exception should we add it in the docblock?

alongosz: We check for occurrence of dynamic `$exception` therefore the annotation is still required. Technically it's possible that the given `$exception` is gonna be different.

eZ/Publish/API/Repository/Tests/FieldType/BaseIntegrationTest.php
alongosz: > One line above it is another `@dep_ends`

Ok, removed all of them then via 53ead296a.

PR #63, opened 2020-05-13
EZP-31574: Added posibility to save and preview incomplete draft

Nattfarinn:
| **JIRA issue**                          | [EZP-31574](https://jira.ez.no/browse/EZP-31574)
| **Tests pass**                          | no

#### Checklist:
- [x] Tests are implemented.

### DIFF
--- a/eZ/Publish/API/Repository/ContentService.php
+++ b/eZ/Publish/API/Repository/ContentService.php
@@ -21,6 +21,7 @@ use eZ\Publish\API\Repository\Values\Content\ContentInfo;
 use eZ\Publish\API\Repository\Values\Content\Relation;
 use eZ\Publish\API\Repository\Values\User\User;
 use eZ\Publish\API\Repository\Values\Content\LocationCreateStruct;
+use eZ\Publish\API\Repository\Values\ValueObject;
 
 /**
  * This class provides service methods for managing content.
@@ -199,10 +200,12 @@ interface ContentService
      * @param \eZ\Publish\API\Repository\Values\Content\ContentCreateStruct $contentCreateStruct
      * @param \eZ\Publish\API\Repository\Values\Content\LocationCreateStruct[] $locationCreateStructs an array of {@link \eZ\Publish\API\Repository\Values\Content\LocationCreateStruct} for each location parent under which a location should be created for the content
      *                                                                                                While optional, it's highly recommended to use Locations for content as a lot of features in the system is usually tied to the tree structure (including default Role policies).
+     * @param string[]|null $fieldIdentifiersToValidate List of field identifiers for partial validation or null
+     *                      for case of full validation. Empty identifiers array is equal to no validation.
      *
      * @return \eZ\Publish\API\Repository\Values\Content\Content - the newly created content draft
      */
-    public function createContent(ContentCreateStruct $contentCreateStruct, array $locationCreateStructs = []): Content;
+    public function createContent(ContentCreateStruct $contentCreateStruct, array $locationCreateStructs = [], ?array $fieldIdentifiersToValidate = null): Content;
 
     /**
      * Updates the metadata.
@@ -307,10 +310,12 @@ interface ContentService
      *
      * @param \eZ\Publish\API\Repository\Values\Content\VersionInfo $versionInfo
      * @param \eZ\Publish\API\Repository\Values\Content\ContentUpdateStruct $contentUpdateStruct
+     * @param string[]|null $fieldIdentifiersToValidate List of field identifiers for partial validation or null
+     *                      for case of full validation. Empty identifiers array is equal to no validation.
      *
      * @return \eZ\Publish\API\Repository\Values\Content\Content the content draft with the updated fields
      */
-    public function updateContent(VersionInfo $versionInfo, ContentUpdateStruct $contentUpdateStruct): Content;
+    public function updateContent(VersionInfo $versionInfo, ContentUpdateStruct $contentUpdateStruct, ?array $fieldIdentifiersToValidate = null): Content;
 
     /**
      * Publishes a content version.
@@ -538,4 +543,18 @@ interface ContentService
      * @return \eZ\Publish\API\Repository\Values\Content\ContentUpdateStruct
      */
     public function newContentUpdateStruct(): ContentUpdateStruct;
+
+    /**
+     * Validates given content related ValueObject returning field errors structure as a result.
+     *
+     * @param array $context Additional context parameters to be used by validators.
+     * @param string[]|null $fieldIdentifiersToValidate List of field identifiers for partial validation or null
+     *                      for case of full validation. Empty identifiers array is equal to no validation.
+     *
+     * @return array Validation errors grouped by field definition and language code, in format:
+     *           $returnValue[string|int $fieldDefinitionId][string $languageCode] = $fieldErrors;
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
+     */
+    public function validate(ValueObject $object, array $context, ?array $fieldIdentifiersToValidate = null): array;
 }
--- a/eZ/Publish/Core/Repository/ContentService.php
+++ b/eZ/Publish/Core/Repository/ContentService.php
@@ -11,6 +11,7 @@ namespace eZ\Publish\Core\Repository;
 use eZ\Publish\API\Repository\ContentService as ContentServiceInterface;
 use eZ\Publish\API\Repository\PermissionResolver;
 use eZ\Publish\API\Repository\Repository as RepositoryInterface;
+use eZ\Publish\API\Repository\Values\ValueObject;
 use eZ\Publish\Core\FieldType\FieldTypeRegistry;
 use eZ\Publish\API\Repository\Values\Content\ContentDraftList;
 use eZ\Publish\API\Repository\Values\Content\DraftList\Item\ContentDraftListItem;
@@ -20,6 +21,7 @@ use eZ\Publish\API\Repository\Values\Content\RelationList\Item\RelationListItem;
 use eZ\Publish\API\Repository\Values\Content\RelationList\Item\UnauthorizedRelationListItem;
 use eZ\Publish\API\Repository\Values\User\UserReference;
 use eZ\Publish\Core\Repository\Mapper\ContentDomainMapper;
+use eZ\Publish\Core\Repository\Mapper\ContentMapper;
 use eZ\Publish\Core\Repository\Values\Content\Content;
 use eZ\Publish\Core\Repository\Values\Content\Location;
 use eZ\Publish\API\Repository\Values\Content\Language;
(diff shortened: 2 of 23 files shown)


eZ/Publish/API/Repository/ContentService.php
```diff
@@ -538,4 +539,6 @@ public function newContentMetadataUpdateStruct(): ContentMetadataUpdateStruct;
      * @return \eZ\Publish\API\Repository\Values\Content\ContentUpdateStruct
      */
     public function newContentUpdateStruct(): ContentUpdateStruct;
+
+    public function validate(ValueObject $object, array $context, ?array $fieldIdentifiers = null): array;
```
bdunogier, line 543: Should we use `$fieldGroupsIdentifiers` instead ? Do we need to go down to individual fields level ?

Nattfarinn: From API point of view I think it is worth going down to individual field level. You will be able to partially validate object programmatically and low level for whatever reason without the need to group fields in advance.

pspanja:
Does this mean it will be possible to publish an invalid draft?

bdunogier:
> AFAIR no, validation was to be moved to publish.

Yes, this is a must have. It must be possible to update or create a draft that is incomplete, but it can NOT be published unless it passes validation.

eZ/Publish/Core/settings/repository/inner.yml
```diff
@@ -202,3 +208,24 @@ services:
...
+
+    eZ\Publish\Core\Repository\Validator\ContentUpdateStructValidator:
+        arguments:
+            $contentMapper: '@eZ\Publish\Core\Repository\Mapper\ContentMapper'
+            $fieldTypeRegistry: '@eZ\Publish\Core\FieldType\FieldTypeRegistry'
+            $contentLanguageHandler: '@ezpublish.spi.persistence.language_handler'
+        tags:
+            - ezplatform.spi.content.validator
+
+    eZ\Publish\SPI\Repository\Validator\ContentValidator: '@eZ\Publish\Core\Repository\Strategy\ContentValidator\ContentValidatorStrategy'
```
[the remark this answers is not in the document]
emodric: Btw, Symfony standards recomend that vendor packages never use autoconfiguration and autowiring.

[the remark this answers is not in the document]
pspanja: @emodric I was wondering what's the deal with that, it's occasionally a real pain when used in vendor code.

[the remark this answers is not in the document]
emodric: Yeah, autowiring should just burn and die in vendor code :D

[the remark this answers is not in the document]
pspanja: I would not complain!

eZ/Publish/Core/Repository/Strategy/ContentValidator/ContentValidatorStrategy.php
```diff
@@ -0,0 +1,52 @@
...
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace eZ\Publish\Core\Repository\Strategy\ContentValidator;
+
+use eZ\Publish\API\Repository\Values\ValueObject;
+use eZ\Publish\Core\Base\Exceptions\InvalidArgumentException;
+use eZ\Publish\SPI\Repository\Validator\ContentValidator;
+
+class ContentValidatorStrategy implements ContentValidator
```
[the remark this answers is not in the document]
Nattfarinn: > Naming is a bit inconsistent with what you've introduced for thumbnails in the past, while the behavior follows similar pattern AFAICS. I mean, for thumbnails you have `ThumbnailChainStrategy`.

Yes, but it isn't the same case. `ThumbnailChainStrategy` has exact meaning: its purpose is to chain various strategies (like `FirstMatchingFieldStrategy`, `StaticStrategy`) and execute them one after another. Here we have one strategy that picks right validator for given object. I don't think Chain or Aggregate is valid here as it would imply this strategy aggregates sub-strategies.

eZ/Publish/Core/settings/repository/inner.yml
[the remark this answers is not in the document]
Nattfarinn: I know autowiring can be a bit of pain if used carelessly, but as long as we use it only for direct injection and not for interface-implementation autobinding everything should be transparent.

Do you have a case where autowiring fails in our code, or at least prevents you from doing what would be possible without it?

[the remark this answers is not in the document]
Nattfarinn: > The downside is that `eZ\Publish\Core\Repository\Strategy\ContentValidator\ContentValidatorStrategy` would need to implement dedicated interface for injection, so it's not gonna get tagged and injected to itself.

That is why I would like to keep it as it is. I don't think custom low level value object validators will be common. Same interface also make validators and validator strategies interchangeable.

Nattfarinn:
@pspanja @bdunogier : You will not be able to publish incomplete draft.

eZ/Publish/Core/settings/repository/inner.yml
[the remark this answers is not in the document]
emodric: It's not that it fails, it's just annoying to debug most of the time. Symfony best practices recommend that it should not be used in vendor code for a reason. I have yet to see the case where it was actually helpful when diving into `vendor/`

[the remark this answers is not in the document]
alongosz: > It's not that it fails, it's just annoying to debug most of the time. Symfony best practices recommend that it should not be used in vendor code for a reason. I have yet to see the case where it was actually helpful when diving into `vendor/`

It's really not a black & white subject. Autowiring & autoconfiguration is a way to avoid maintenance of a long list of definitions which are redundant. As long as applied for non-ambiguous cases, everything should be fine.
In this case I was referring to an autoconfiguration which would be beneficial not just internally but for 1st and 3rd party. If autoconfiguration is set, there's no point in vendor code to define its own explicitly, as that autoconfiguration applies to vendor definitions as well.

eZ/Publish/API/Repository/ContentService.php
```diff
@@ -202,7 +203,7 @@ public function loadContentListByContentInfo(array $contentInfoList, array $lang
      *
      * @return \eZ\Publish\API\Repository\Values\Content\Content - the newly created content draft
      */
-    public function createContent(ContentCreateStruct $contentCreateStruct, array $locationCreateStructs = []): Content;
+    public function createContent(ContentCreateStruct $contentCreateStruct, array $locationCreateStructs = [], ?array $fieldIdentifiersToValidate = null): Content;
```
alongosz, line 206: While in the previous implementation new boolean argument was rather straightforward, now I'm missing a bit of PHPDoc regarding `?array $fieldIdentifiersToValidate = null`. The questions such doc needs to answer are:
* what happens when value is default `null`?
* how is that different when value is `[]`? 
* how does it work in general?

eZ/Publish/API/Repository/ContentService.php
```diff
@@ -310,7 +311,7 @@ public function loadContentDraftList(?User $user = null, int $offset = 0, int $l
      *
      * @return \eZ\Publish\API\Repository\Values\Content\Content the content draft with the updated fields
      */
-    public function updateContent(VersionInfo $versionInfo, ContentUpdateStruct $contentUpdateStruct): Content;
+    public function updateContent(VersionInfo $versionInfo, ContentUpdateStruct $contentUpdateStruct, ?array $fieldIdentifiersToValidate = null): Content;
```
alongosz, line 314: applies to this part as well

eZ/Publish/Core/settings/repository/inner.yml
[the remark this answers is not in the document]
emodric: It is black and white. Either you follow Symfony best practices, or you don't. I have nothing against enabling autoconfiguration for 3rd party code (and by that, I mean project code residing in `src/`), but everything else just puts strain on every-day

eZ/Publish/Core/Repository/ContentService.php
```diff
@@ -596,14 +606,25 @@ public function createContent(APIContentCreateStruct $contentCreateStruct, array
             $contentCreateStruct->remoteId = $this->contentDomainMapper->getUniqueHash($contentCreateStruct);
         }
 
+        if (is_array($fieldIdentifiersToValidate)) {
```
adamwojs, line 609: !empty($fieldIdentifiersToValidate)

eZ/Publish/Core/Repository/ContentService.php
```diff
@@ -1336,50 +1233,59 @@ protected function internalUpdateContent(APIVersionInfo $versionInfo, APIContent
...
         );
+
         if (!$content->versionInfo->isDraft()) {
             throw new BadStateException(
                 '$versionInfo',
                 'The version is not a draft and cannot be updated'
             );
         }
 
+        if (is_array($fieldIdentifiersToValidate)) {
```
adamwojs, line 1244: !empty($fieldIdentifiersToValidate)

eZ/Publish/Core/Repository/Mapper/ContentMapper.php
```diff
@@ -0,0 +1,230 @@
...
+        return $fields;
+    }
+
+    /**
+     * Returns all language codes used in given $fields.
+     *
+     * @param \eZ\Publish\API\Repository\Values\Content\ContentUpdateStruct $contentUpdateStruct
+     * @param \eZ\Publish\API\Repository\Values\Content\Content $content
+     *
+     * @return array
```
adamwojs, line 168:
```suggestion
     * @return string[]
```

eZ/Publish/Core/Repository/Mapper/ContentMapper.php
```diff
@@ -0,0 +1,230 @@
...
+
+        return array_keys($languageCodes);
+    }
+
+    /**
+     * Returns only updated language codes.
+     *
+     * @param \eZ\Publish\API\Repository\Values\Content\ContentUpdateStruct $contentUpdateStruct
+     *
+     * @return array
```
adamwojs, line 188:
```suggestion
     * @return string[]
```

eZ/Publish/Core/Repository/Mapper/ContentMapper.php
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /**
+     * Clones $field with overriding specific properties from given $overrides array.
+     *
+     * @param Field $field
+     * @param array $overrides
+     *
+     * @return Field
+     */
```
adamwojs, line 214: Nitpick: Please use FQCN

eZ/Publish/Core/Repository/Validator/ContentCreateStructValidator.php
```diff
@@ -0,0 +1,103 @@
...
+    public function validate(
+        ValueObject $object,
+        array $context = [],
+        ?array $fieldIdentifiers = null
+    ): array {
+        if (!$this->supports($object)) {
+            throw new InvalidArgumentException('$object', 'Not supported');
+        }
+
+        /** @var ContentCreateStruct $contentCreateStruct */
```
adamwojs, line 52: Nitpick: Please use FQCN

eZ/Publish/Core/Repository/Validator/ContentCreateStructValidator.php
```diff
@@ -0,0 +1,103 @@
...
+        $fields = $this->contentMapper->mapFieldsForCreate($contentCreateStruct);
+
+        $allFieldErrors = [];
+
+        foreach ($contentCreateStruct->contentType->getFieldDefinitions() as $fieldDefinition) {
+            if (isset($fieldIdentifiers) && !in_array($fieldDefinition->fieldTypeIdentifier, $fieldIdentifiers)) {
+                continue;
+            }
+
+            /** @var $fieldType \eZ\Publish\Core\FieldType\FieldType */
```
adamwojs, line 65:
```suggestion
            /** @var \eZ\Publish\Core\FieldType\FieldType $fieldType */
```

eZ/Publish/Core/Repository/Validator/VersionValidator.php
```diff
@@ -0,0 +1,103 @@
...
+    public function validate(
+        ValueObject $object,
+        array $context = [],
+        ?array $fieldIdentifiers = null
+    ): array {
+        if (!$this->supports($object)) {
+            throw new InvalidArgumentException('$object', 'Not supported');
+        }
+
+        /** @var VersionInfo $versionInfo */
```
adamwojs, line 52: Nitpick: Please use FQCN

eZ/Publish/SPI/Repository/Validator/ContentValidator.php
```diff
@@ -0,0 +1,25 @@
...
+    public function supports(ValueObject $object): bool;
+
+    /**
+     * @param string[]|null $fieldIdentifiers
+     */
+    public function validate(
+        ValueObject $object,
+        array $context = [],
+        ?array $fieldIdentifiers = null
+    ): array;
```
adamwojs, line 24: Q: Are there any requirements regarding result structure?

Nattfarinn: Yes, docblock updated.

eZ/Publish/API/Repository/ContentService.php
Nattfarinn: Docblock updated.

eZ/Publish/API/Repository/ContentService.php
Nattfarinn: Docblock updated.

PR #74, opened 2020-06-04
EZP-31671: Fix getRootLocation access to  private service

wizhippo:
| **JIRA issue**                          | [EZP-31671](https://jira.ez.no/browse/EZP-31671)
| **Tests pass**                          | yes/no

`eZ\Bundle\EzPublishCoreBundle\Controller::getGlobalHelper` tries to get `ezpublish.templating.global_helper` but this service is private.

#### Checklist:

### DIFF
--- a/eZ/Bundle/EzPublishCoreBundle/Resources/config/templating.yml
+++ b/eZ/Bundle/EzPublishCoreBundle/Resources/config/templating.yml
@@ -106,6 +106,7 @@ services:
         alias: ezpublish.templating.global_helper.core
 
     ezpublish.templating.global_helper:
+        public: true
         alias: ezpublish.templating.global_helper.core
 
     ezpublish.twig.extension.core:


alongosz:
@wizhippo This Controller is a [Service subscriber](https://symfony.com/doc/5.0/service_container/service_subscribers_locators.html) so making dependent subscribed services as public should not be required and is even implicitly discouraged.

What are your steps to reproduce? :)
For me this example works well: https://github.com/ezsystems/ezplatform/compare/v3.0.4...alongosz:base-v3.0.4/ezp-31671-test

Maybe your controller is not tagged properly? This can be achieved either by setting `autoconfigure: true` for controller Symfony Service Definition or by tagging it as: `container.service_subscriber` (and additionally as `controller.service_arguments`, but not relevant for the issue here).

Our meta-repository shipped with v3.0.x [does this by default](https://github.com/ezsystems/ezplatform/blob/v3.0.4/config/services.yaml#L11-L19) if conventions are followed.

wizhippo:
@alongosz Good catch.

wizhippo:
@alongosz Spoke too soon.  Migrating to ezplatform 3.0.4.

I still have the issue after I remove my redefinition of 
```
   ezpublish.templating.global_helper:
        public: true
        alias: ezpublish.templating.global_helper.core
```

The controller has a parent:

```
services:
    _defaults:
        autowire: true
        autoconfigure: true

    App\Controller\Controller:
        calls:
            - [setContainer, ["@service_container"]]
            - [performAccessCheck, []]
        tags: ['controller.service_arguments', 'container.service_subscriber']
```

```
services:
    App\Controller\ReportController:
        parent: App\Controller\Controller
        arguments:
            $jasperServerFormHelper: '@jasper_client.form_helper'
            $jasperClient: '@jasper_client.client'
```

I still get the error

```
An exception has been thrown during the rendering of a template ("The "ezpublish.templating.global_helper" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection instead.").
```

Which when I debug brings me to `$rootLocation = $this->getRootLocation();`

`App\Controller\Controller` extends `eZ\Bundle\EzPublishCoreBundle\Controller`

wizhippo:
@alongosz ok that it uses a `Service subscriber` but `setContatiner` is called with the container so the container is the container and not the locator that sue Service subscriber builds.

The locator needed to be injected and then used in `getGlobalHelper()` instead of the container.  Otherwise the service need to be public the use the container.

alongosz:
@wizhippo I think I found the culprit. Direct `[setContainer, ["@service_container"]]` call messes up Service Subscriber & Locator feature. This feature essentially injects an instance of a smaller container of sub-type `\Symfony\Component\DependencyInjection\Argument\ServiceLocator` with the subscribed services available for fetching via `get`.
If you explicitly set full `@service_container`, it behaves as it should in Symfony 4+ - not exposing private services.

The second issue is that `parent` does not inherit `tags`, thus Service Locator is not injected into `ReportController` instance in your case. At least `container.service_subscriber` needs to be added to every controller.

Given the eZ Platform v3.x base Controller uses the Service Subscriber & Locator feature, the behavior is correct and configuration needs to be migrated (look at this as a migration step to upgrade to eZ Platform 3.0 & Symfony 5). Naturally we need to upgrade our doc, we simply missed this use case.

To sum up, working diff: https://github.com/ezsystems/ezplatform/compare/v3.0.4...alongosz:f1ef07a

Please let me know if that works for you.

I still would recommend to:
* `autowire` and `autoconfigure` controllers[1],
* inject required services because Service Locator seems troublesome (e.g. needs an explicit tag for every service).

---

[1] Like [that](https://github.com/ezsystems/ezplatform/compare/v3.0.4...alongosz:a29de69) - please observe in the linked diff how little effort is required for the use case to work. This utilizes the already mentioned [default configuration](https://github.com/ezsystems/ezplatform/blob/v3.0.4/config/services.yaml#L11-L25) which can be seen in the most recent Symfony 5 applications created from Flex Skeleton.

If the default does not work for you, the required pieces for this to work are:
* `autowire: true`, 
* `autoconfigure: true`, 
* `tags: ['controller.service_arguments', 'container.service_subscriber']`

defined for each Controller.

alongosz:
> @alongosz ok that it uses a `Service subscriber` but `setContatiner` is called with the container so the container is the container and not the locator that the Service subscriber builds.
> 
> The locator needed to be injected and then used in `getGlobalHelper()` instead of the container. Otherwise the service need to be public the use the container.

@wizhippo sorry, missed that reply when writing my own. Yes, exactly that. Since you're using `autowire` from `_defaults` for `Controller` anyway, you don't need to add `setContainer` call. The another missing piece was Service Subscriber tag on `ReportController`.

wizhippo:
@alongosz Ok I can get it working if I do not follow the original `ezpublish.controller.base` as an example.

Does this not mean that this is incorrect as it does `[setContainer, ["@service_container"]]`?

```
# eZ/Bundle/EzPublishCoreBundle/Resources/config/services.yml
   ezpublish.controller.base:
        class: eZ\Publish\Core\MVC\Symfony\Controller\Controller
        abstract: true
        calls:
            - [ setContainer, ["@service_container"] ]
```

wizhippo:
This is really a mess because now if I want to use the container I can't as it is now the limited locator.

I will of course move to injecting the services instead but is a pain for migrating an old code base.

alongosz:
> @alongosz Ok I can get it working if I do not follow the original `ezpublish.controller.base` as an example.
> 
> Does this not mean that this is incorrect as it does `[setContainer, ["@service_container"]]`?
> 
> ```
> # eZ/Bundle/EzPublishCoreBundle/Resources/config/services.yml
>    ezpublish.controller.base:
>         class: eZ\Publish\Core\MVC\Symfony\Controller\Controller
>         abstract: true
>         calls:
>             - [ setContainer, ["@service_container"] ]
> ```

Why would you use `ezpublish.controller.base`? Those are two different controllers.

AFAICS `\eZ\Publish\Core\MVC\Symfony\Controller\Controller` is not related to `\eZ\Bundle\EzPublishCoreBundle\Controller`. Moreover I don't see in our doc any reference to neither the `Core\MVC\Symfony\Controller` nor `ezpublish.controller.base`. It's rather meant for eZ Platform core, not projects.

> This is really a mess because now if I want to use the container I can't as it is now the limited locator.
> 

Or use different base for the time being and migrate step by step. Actually the core container most likely will behave as it used to. This however won't make a difference to you since the method you're seeking - `getRootLocation` is available in `\eZ\Bundle\EzPublishCoreBundle\Controller` only.

> I will of course move to injecting the services instead but is a pain for migrating an old code base.

I'm sorry, but Symfony makes us to utilize the best practices. Migration is a price of progress. I haven't traced back the process during which we ended up with Symfony `AbstractController` as a base (which kinda forced Service Subscriber for our use case), but I'd guess it was one of the required upgrade steps. Feel free to blame Symfony :-)

_**Injecting entire Service Container and accessing services from that container is a bad practice. Not relying on that was enforced by Symfony itself**_. For some use cases where getting services would still be more convenient, Service Subscribers & Locators were introduced and can be used.

wizhippo:
Thanks for checking into this.  I blindly followed my editors code inspection back to the service definition. Took me to the wrong one.

wizhippo:
@alongosz Found my reference to the wrong base.

Note `EzSystems\EzPlatformContentFormsBundle\Controller\UserController` extends `eZ\Bundle\EzPublishCoreBundle\Controller` but in the service definition the parent is `ezpublish.controller.base` which is of `eZ\Publish\Core\MVC\Symfony\Controller\Controller` 

Guess I shouldn't assume the service parent would be the same as the parent class.

```
    ezplatform.content_forms.controller.user:
        class: "%ezplatform.content_forms.controller.user.class%"
        arguments:
            - "@ezpublish.api.service.content_type"
            - "@ezpublish.api.service.user"
            - "@ezpublish.api.service.location"
            - "@ezpublish.api.service.language"
            - "@ezplatform.content_forms.action_dispatcher.user"
            - '@eZ\Publish\API\Repository\PermissionResolver'
            - '@eZ\Publish\Core\MVC\Symfony\Locale\UserLanguagePreferenceProvider'
        parent: ezpublish.controller.base
        tags:
              - { name: controller.service_arguments }
```

wizhippo:
Just a follow still had a few odd issues.

Found using the bellow added the correct container
```
        calls:
            - [setContainer, ['@Psr\Container\ContainerInterface']]
            - [performAccessCheck, []]
```

PR #76, opened 2020-06-12
[WIP] EZP-31682: Support Argon2 password hashes

glye:
| **JIRA issue**                          | [EZP-31682](https://jira.ez.no/browse/EZP-31682)
| **Tests pass**                          | tbd

Add support for Argon2 password hashes, which are supported in PHP 7.2 / 7.3 but optional, PHP must be compiled with support for them.
https://www.php.net/manual/en/function.password-hash.php

Out of scope: Add support for the various cost/threads options. We could add this for bcrypt too.

#### Checklist:
- [ ] Tests are implemented.
- [ ] Added code follows Coding Standards (use `$ composer fix-cs`).
- [ ] isValidPassword() also needs to check if support is compiled in, refactor to do this all in one place.


glye:
Continued in https://github.com/ibexa/core/pull/581

PR #97, opened 2020-08-14
On ContentType copy clear the responding persistence cache

MarioBlazek:
| **JIRA issue**                          | [EZP-31800](https://jira.ez.no/browse/EZP-31800)

#### Checklist:
- [ ] PR description is updated.
- [ ] Tests are implemented.
- [ ] Added code follows Coding Standards (use `$ composer fix-cs`).

### DIFF
--- a/eZ/Publish/Core/Persistence/Cache/ContentTypeHandler.php
+++ b/eZ/Publish/Core/Persistence/Cache/ContentTypeHandler.php
@@ -332,8 +332,13 @@ class ContentTypeHandler extends AbstractInMemoryPersistenceHandler implements C
     public function copy($userId, $typeId, $status)
     {
         $this->logger->logCall(__METHOD__, ['user' => $userId, 'type' => $typeId, 'status' => $status]);
+        $contentType = $this->persistenceHandler->contentTypeHandler()->copy($userId, $typeId, $status);
 
-        return $this->persistenceHandler->contentTypeHandler()->copy($userId, $typeId, $status);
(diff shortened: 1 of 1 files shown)


alongosz (changes requested):
I don't understand why you'd want to clear cache of source content type. It should be unchanged after copying, shouldn't it?

MarioBlazek:
But the new content type is not in the listing cache yet. So, if you clear the cache for the new content type, the listing does not refresh.

On the other hand, the new content type is freshly created, why would you want to clear the cache for it?

MarioBlazek:
@alongosz what do you suggest here? Maybe I should load the group of this content type, and then clear the cache for group?

alongosz:
> But the new content type is not in the listing cache yet. So, if you clear the cache for the new content type, the listing does not refresh.
> 
> On the other hand, the new content type is freshly created, why would you want to clear the cache for it?

Both new type and old type are unchanged (the new one has actually cold cache). So the question is what the actual issue is? List not being refreshed? Then rather tags related to list(s) should be invalidated.

This PR misses integration test coverage (`\eZ\Publish\API\Repository\Tests\ContentTypeServiceTest`). Integration test setup takes into the account persistence cache, so the issue you're experiencing should be reproducible. It's gonna be easier to identify what actually we need to invalidate.

Moreover it probably applies to `ezpublish-kernel:7.5` as well, since the issue occurs in API, not UI :-)

alongosz:
Closing in favor of ezsystems/ezpublish-kernel#3069.

PR #108, opened 2020-09-17
EZP-31740: Introduced SiteAccess-aware NFS adapter

alongosz:
| **JIRA issue**                          | [EZP-31740](https://jira.ez.no/browse/EZP-31740)
| **Required by**                        | ezsystems/ezplatform#596
| **Target eZ Platform version** | `v3.1`+
| **Doc needed**                       | yes, will be covered in meta PR

## TL;DR;

This PR introduces our custom SiteAccess-aware local NFS adapter based on Flysystem's one. For configuration see the meta PR (ezsystems/ezplatform#596).

## Summary

We need to be able to process dynamic paths for NFS instance in the form of `%dfs_nfs_path%/$var_dir$/$storage_dir$`, mapped to proper Repository & SiteAccess settings based on a current request context. Prior Symfony 4 it was done by replacing injected service parameters on the fly when context had changed. Since Symfony 4+ it's no longer possible because the Container is frozen. Moreover in this case settings apply to 3rd party Semantic configuration, so usual simple fix with ConfigResolver injection is not possible.

To solve the issue we've introduced [custom adapter](https://github.com/1up-lab/OneupFlysystemBundle/blob/3.5.0/Resources/doc/adapter_custom.md) which resolves dynamically the mentioned path parameter.

Existing `ComplexConfigProcessor`, which previously was used to replace container parameters on the fly, has been refactored to process strings with multiple settings. Note that SiteAccess name resolution has been dropped there from the last step, because it's redundant (`getParameter` should use current scope if none given). In fact it can probably be dropped from the entire service, but further codebase adjustments are required for that.

## Review

See review notes in diff comments.

## QA

See ezsystems/ezplatform#596

## TODO
- [x] Add unit tests coverage.
- [x] Wait for Travis

#### Checklist:
- [x] Tests are implemented.

### DIFF
--- a/eZ/Bundle/EzPublishCoreBundle/SiteAccess/Config/ComplexConfigProcessor.php
+++ b/eZ/Bundle/EzPublishCoreBundle/SiteAccess/Config/ComplexConfigProcessor.php
@@ -12,9 +12,10 @@ use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\ComplexSetti
 use eZ\Publish\Core\MVC\Exception\ParameterNotFoundException;
 use eZ\Publish\Core\MVC\ConfigResolverInterface;
 use eZ\Publish\Core\MVC\Symfony\SiteAccess\SiteAccessService;
+use eZ\Publish\SPI\SiteAccess\ConfigProcessor;
 use function str_replace;
 
-final class ComplexConfigProcessor
+final class ComplexConfigProcessor implements ConfigProcessor
 {
     private const DEFAULT_NAMESPACE = 'ezsettings';
 
--- a/eZ/Bundle/EzPublishCoreBundle/SiteAccess/Config/ComplexConfigProcessor.php
+++ b/eZ/Bundle/EzPublishCoreBundle/SiteAccess/Config/ComplexConfigProcessor.php
@@ -12,6 +12,7 @@ use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\ComplexSetti
 use eZ\Publish\Core\MVC\Exception\ParameterNotFoundException;
 use eZ\Publish\Core\MVC\ConfigResolverInterface;
 use eZ\Publish\Core\MVC\Symfony\SiteAccess\SiteAccessService;
+use function str_replace;
 
 final class ComplexConfigProcessor
 {
@@ -23,12 +24,18 @@ final class ComplexConfigProcessor
     /** @var \eZ\Publish\Core\MVC\Symfony\SiteAccess\SiteAccessService */
     private $siteAccessService;
 
+    /** @var \eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\ComplexSettings\ComplexSettingParserInterface */
+    private $complexSettingParser;
+
     public function __construct(
         ConfigResolverInterface $configResolver,
         SiteAccessService $siteAccessService
     ) {
         $this->configResolver = $configResolver;
         $this->siteAccessService = $siteAccessService;
+
+        // instantiate non-injectable DI configuration parser
+        $this->complexSettingParser = new ComplexSettingParser();
     }
 
     public function processComplexSetting(string $setting): string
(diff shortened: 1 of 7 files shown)


eZ/Bundle/EzPublishCoreBundle/SiteAccess/Config/ComplexConfigProcessor.php
```diff
@@ -23,12 +24,18 @@ final class ComplexConfigProcessor
...
+
     public function __construct(
         ConfigResolverInterface $configResolver,
         SiteAccessService $siteAccessService
     ) {
         $this->configResolver = $configResolver;
         $this->siteAccessService = $siteAccessService;
+
+        // instantiate non-injectable DI configuration parser
+        $this->complexSettingParser = new ComplexSettingParser();
```
alongosz, line 38: IMHO not worth to make it a service and probably not entirely correct, since it's a part of DI extension config processors. Anyway, it's been like this before (see the line 42).

eZ/Bundle/EzPublishCoreBundle/SiteAccess/Config/ComplexConfigProcessor.php
```diff
@@ -39,31 +46,35 @@ public function processComplexSetting(string $setting): string
...
             }
-            if (!isset($parts['scope'])) {
-                $parts['scope'] = $siteAccessName;
-            }
 
-            $dynamicSettingValue = $this->configResolver->getParameter($parts['param'], $parts['namespace'], $parts['scope']);
+            $dynamicSettingValue = $this->configResolver->getParameter(
+                $parts['param'],
+                $parts['namespace'],
+                $parts['scope'] ?? $this->siteAccessService->getCurrent()->name
```
alongosz, line 76: Kept for BC of behavior, but this is indeed redundant (TBD if can be safely fixed).

eZ/Bundle/EzPublishIOBundle/DependencyInjection/ConfigurationFactory/Flysystem.php
```diff
@@ -62,7 +62,8 @@ public function configureHandler(ServiceDefinition $definition, array $config)
     private function createFilesystem(ContainerBuilder $container, $name, $adapter)
     {
         $adapterId = sprintf('oneup_flysystem.%s_adapter', $adapter);
-        if (!$container->hasDefinition($adapterId)) {
+        // has either definition or alias
+        if (!$container->has($adapterId)) {
```
alongosz, line 66: Custom adapter is exposed by Oneup as an alias, hence this adjustment is required.

eZ/Bundle/EzPublishIOBundle/Flysystem/Adapter/SiteAccessAwareLocalAdapter.php
```diff
@@ -0,0 +1,40 @@
...
+    public function __construct(
+        ComplexConfigProcessor $complexConfigProcessor,
+        array $config
+    ) {
+        parent::__construct(
+            $config['directory'],
+            $config['writeFlags'],
+            $config['linkHandling'],
+            $config['permissions']
+        );
```
alongosz, line 31: Done like that to simplify meta configuration.

eZ/Bundle/EzPublishIOBundle/Resources/config/io.yml
```diff
@@ -1,3 +1,10 @@
+parameters:
+    ibexa.platform.io.nfs.adapter.config:
+        directory: '%dfs_nfs_path%/$var_dir$/$storage_dir$'
+        writeFlags: ~
+        linkHandling: ~
+        permissions: [ ]
```
alongosz, line 6: Exposed also in meta, for custom project configuration. Seems to me array is more readable than a parameter per setting.

eZ/Bundle/EzPublishIOBundle/Resources/config/io.yml
```diff
@@ -103,6 +110,14 @@ services:
             - '@eZ\Publish\Core\IO\IOConfigProvider'
             - '@ezpublish.config.resolver'
 
+    eZ\Bundle\EzPublishIOBundle\Flysystem\Adapter\SiteAccessAwareLocalAdapter:
+        arguments:
+            $complexConfigProcessor: '@eZ\Bundle\EzPublishCoreBundle\SiteAccess\Config\ComplexConfigProcessor'
+            $config: '%ibexa.platform.io.nfs.adapter.config%'
+
+    ibexa.platform.io.nfs.adapter.site_access_aware:
```
alongosz, line 118: To be used by custom project in meta rather than FQCN.

alongosz (commented):
Review notes:

webhdx (approved):
Nothing to complain about. Even though I'm not a fan of keeping complex settings after we've removed dynamic settings, it's at least in line with IO configuration. We should try to get rid of it in next major completely.

webhdx (approved):

bdunogier (approved):

alongosz:
Update 1db293b: fixed the issue with initial directory being created at a Container compile time in the form of `<dfs_path>/$var_dir$/$storage_dir$`. Flysystem local adapter always tries to ensure directory exists (in adapter's constructor), before pre-processing its path.

To solve that I've separated static DFS/NFS root path setting, from dynamic path "pattern". SiteAccess-aware path part is applied now only when resolving concrete file path.

ViniTou (approved):

micszo (approved):
Test status in https://github.com/ezsystems/ezplatform/pull/596.

PR #109, opened 2020-09-22
EZP-31741: Detect proxy in eZ\Publish\Core\Repository\SearchService::supports

adamwojs:
| **JIRA issue**                          | [EZP-31741](https://jira.ez.no/browse/EZP-31741)

The following check 

https://github.com/ezsystems/ezplatform-kernel/blob/1.1/eZ/Publish/Core/Repository/SearchService.php#L347-L349

will always fail because injected  `\eZ\Publish\SPI\Search\Handler` is proxy object generated from interface instead of concrete class.  

See also https://jira.ez.no/browse/EZP-31701

#### Checklist:
- [ ] Tests are implemented.
- [X] PR is ready for a review.

### DIFF
--- a/eZ/Publish/Core/Repository/SearchService.php
+++ b/eZ/Publish/Core/Repository/SearchService.php
@@ -27,6 +27,8 @@ use eZ\Publish\Core\Repository\Mapper\ContentDomainMapper;
 use eZ\Publish\SPI\Search\Capable;
 use eZ\Publish\Core\Search\Common\BackgroundIndexer;
 use eZ\Publish\SPI\Search\Handler;
+use ProxyManager\Proxy\LazyLoadingInterface;
+use ProxyManager\Proxy\ValueHolderInterface;
 
 /**
  * Search service.
@@ -344,8 +346,18 @@ class SearchService implements SearchServiceInterface
 
     public function supports(int $capabilityFlag): bool
     {
-        if ($this->searchHandler instanceof Capable) {
-            return $this->searchHandler->supports($capabilityFlag);
+        $handler = $this->searchHandler;
+
+        if ($handler instanceof ValueHolderInterface) {
+            if ($handler instanceof LazyLoadingInterface && !$handler->isProxyInitialized()) {
+                $handler->initializeProxy();
+            }
+
+            $handler = $handler->getWrappedValueHolderValue();
+        }
+
+        if ($handler instanceof Capable) {
+            return $handler->supports($capabilityFlag);
         }
 
         return false;


eZ/Publish/Core/Repository/SearchService.php
```diff
@@ -344,8 +346,18 @@ protected function addPermissionsCriterion(Criterion &$criterion)
...
+            if ($handler instanceof LazyLoadingInterface && !$handler->isProxyInitialized()) {
+                $handler->initializeProxy();
+            }
+
+            $handler = $handler->getWrappedValueHolderValue();
+        }
+
+        if ($handler instanceof Capable) {
+            return $handler->supports($capabilityFlag);
         }
```
alongosz, line 361: This is gonna be an issue for a lot of cases due to BC policies we have. I was rather thinking to delegate this decision to compiler pass magic in this instance.
We need an interface documented to be non-bc-safe which extends both Handler and Capable.

So, back to the interface - compiler pass should use it to instruct lazy proxy how to define service, if that service is indeed capable, it should substitute it. Not sure if it's gonna work the way I want though.

_Capable should've been part of Handler since 3.0, but I didn't manage to find time to merge interfaces (I didn't find time to do a lot of things)._

adamwojs: > This is gonna be an issue for a lot of cases due to BC policies we have. I was rather thinking to delegate this decision to compiler pass magic in this instance.

I don't think it's possible. 

> We need an interface documented to be non-bc-safe which extends both Handler and Capable.

If you have in mind changing `ezpublish.spi.search_engine` definition to 

```
    ezpublish.spi.search_engine:
        class: eZ\Publish\SPI\Search\Engine
        factory: ["@ezpublish.api.search_engine.factory", buildSearchEngine]
        public: false
        lazy: true
```

where `eZ\Publish\SPI\Search\Engine` is

```
interface Engine extends Handler, Capable
{
}
```

it will not gonna work. If `\eZ\Publish\SPI\Search\Handler` implementation doesn't implement `\eZ\Publish\SPI\Search\Capable` (e.g. LSE) then calling \``\eZ\Publish\SPI\Search\Capable::supports` will throw **UndefinedMethodError**

```
Attempted to call an undefined method named "supports" of class "ContainerQXIx3RC\Handler_b009400".
```

> _Capable should've been part of Handler since 3.0, but I didn't manage to find time to merge interfaces (I didn't find time to do a lot of things)._

Agree.

adamwojs:
Closed in favor of #113

PR #110, opened 2020-09-22
EZP-31912: Fixed decorating binary files URLs with proper absolute prefix

alongosz:
| **JIRA issue**                          | EZP-31912
| **Target eZ Platform version** | `v3.1`+
| **Doc needed**                       | ezsystems/developer-documentation#1142

This PR fixes [decorating binary files (images, other media) URLs with custom `url_prefix`](https://doc.ezplatform.com/en/3.1/guide/file_management/#using-a-static-server-for-images).

While the JIRA ticket describes complicated AWS case, in fact URL prefix can be used with local (default) binary handler and it was the part which was broken.

In 2.5 [we used to fetch `io.url_prefix` from dynamic (SiteAccess-aware) configuration for AbsolutePrefix](https://github.com/ezsystems/ezpublish-kernel/blob/v7.5.11/eZ/Bundle/EzPublishIOBundle/Resources/config/io.yml#L131) which decorates media URLs.

Since dynamic settings feature was no longer possible due to Symfony 4+ Container changes, we've rewritten the decorating to make the implementation [fetch the proper setting](https://github.com/ezsystems/ezplatform-kernel/blob/v1.0.0/eZ/Publish/Core/IO/UrlDecorator/AbsolutePrefix.php#L20).

It seems however that what should be used at the line above is actually `getUrlPrefix` instead of `getLegacyUrlPrefix`. See [the parameter names fetched by both methods](https://github.com/ezsystems/ezplatform-kernel/blob/v1.0.0/eZ/Bundle/EzPublishCoreBundle/SiteAccess/Config/IOConfigResolver.php#L34-L39).

### TODO:

- [x] See if DI Container test coverage can be added for this case to avoid another regression in the future

#### Checklist:
- [x] Tests are implemented.

### DIFF
--- a/eZ/Bundle/EzPublishIOBundle/Tests/DependencyInjection/EzPublishIOExtensionTest.php
+++ b/eZ/Bundle/EzPublishIOBundle/Tests/DependencyInjection/EzPublishIOExtensionTest.php
@@ -6,12 +6,20 @@
  */
 namespace eZ\Bundle\EzPublishIOBundle\Tests\DependencyInjection;
 
+use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Compiler\ChainConfigResolverPass;
+use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\ComplexSettings\ComplexSettingParser;
+use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\Parser;
+use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\EzPublishCoreExtension;
 use eZ\Bundle\EzPublishIOBundle\DependencyInjection\EzPublishIOExtension;
 use eZ\Bundle\EzPublishIOBundle\DependencyInjection\ConfigurationFactory;
+use eZ\Publish\API\Repository\Tests\Container\Compiler\SetAllServicesPublicPass;
 use Matthias\SymfonyDependencyInjectionTest\PhpUnit\AbstractExtensionTestCase;
+use Symfony\Component\Yaml\Yaml;
 
 class EzPublishIOExtensionTest extends AbstractExtensionTestCase
 {
+    const FIXTURES_DIR = __DIR__ . '/../_fixtures';
+
     protected function getContainerExtensions(): array
     {
         $extension = new EzPublishIOExtension();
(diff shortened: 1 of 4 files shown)


alongosz:
@ezsystems/documentation-team qualification of "doc needed" changed, but I got you covered there ;)

kaff (approved):

lserwatka (approved):

mateuszdebinski (approved):

eZ/Bundle/EzPublishIOBundle/Tests/DependencyInjection/EzPublishIOExtensionTest.php
```diff
@@ -6,12 +6,20 @@
...
+use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\EzPublishCoreExtension;
 use eZ\Bundle\EzPublishIOBundle\DependencyInjection\EzPublishIOExtension;
 use eZ\Bundle\EzPublishIOBundle\DependencyInjection\ConfigurationFactory;
+use eZ\Publish\API\Repository\Tests\Container\Compiler\SetAllServicesPublicPass;
 use Matthias\SymfonyDependencyInjectionTest\PhpUnit\AbstractExtensionTestCase;
+use Symfony\Component\Yaml\Yaml;
 
 class EzPublishIOExtensionTest extends AbstractExtensionTestCase
 {
+    const FIXTURES_DIR = __DIR__ . '/../_fixtures';
```
adamwojs, line 21:
```suggestion
    private const FIXTURES_DIR = __DIR__ . '/../_fixtures';
```

adamwojs (approved):

ViniTou (approved):

mikadamczyk (approved):

eZ/Bundle/EzPublishIOBundle/Tests/DependencyInjection/EzPublishIOExtensionTest.php
alongosz: Good catch, fixed.

micszo (approved):
Retested successfully with single and multi repo setup using io.url_prefix.
Tested on eZ Platform EE v3.1.2 with diff.

PR #117, opened 2020-10-12
EZP-31825: Fixed user with user:password unable to change password

Steveb-p:
| **JIRA issue**                          | EZP-31825

Fixes regression introduced in 7901c9912a44ef5a58f57ddac97c463e539e92f1 ([EZP-30394](https://jira.ez.no/browse/EZP-30394)) that prevented users with `user/password` role, but without `user/read` or `user/edit`, from changing their own passwords.

15.10.2020 - **PR updated according to discussion with @alongosz.**

Basically, to keep backward compatibility with previous kernel version, `updateUser` method is not really changed other than refactoring duplicate code.
A new `updateUserPassword` is introduced that has the same arguments as `updateUser`. Reason for this is to allow developers to use the same events as previously (i.e. `\eZ\Publish\API\Repository\Events\User\BeforeUpdateUserEvent` and `\eZ\Publish\API\Repository\Events\User\UpdateUserEvent`) that were dispatched in the original method.

#### Checklist:


eZ/Publish/Core/Repository/UserService.php
```diff
@@ -731,40 +732,50 @@ public function updateUser(APIUser $user, UserUpdateStruct $userUpdateStruct): A
...
-            $this->repository->commit();
-        } catch (Exception $e) {
-            $this->repository->rollback();
-            throw $e;
+        if ($isUserProfileUpdateRequested) {
+            // Perform update in user context
+            $updateUserCallable();
+        } else {
+            // Perform update without checking other user permissions
+            $this->repository->sudo($updateUserCallable);
```
andrerom, line 778: :worried:

glye: If sudo is needed, it would look better imo to specifically check for password change requests, rather than `!$isUserProfileUpdateRequested`. In case there are, or could later be, more cases than these two (content edit and password change).

Steveb-p: :heavy_check_mark: 

We've discussed this and password change will probably be separated from the user update in general in this service, to prevent accidently introducing security issues.

Steveb-p:
Basically, to keep backward compatibility with previous kernel version, `updateUser` method is not really changed other than refactoring duplicate code.
A new `updateUserPassword` is introduced that has the same arguments as `updateUser`. Reason for this is to allow developers to use the same events as previously (i.e. `\eZ\Publish\API\Repository\Events\User\BeforeUpdateUserEvent` and `\eZ\Publish\API\Repository\Events\User\UpdateUserEvent`) that were dispatched in the original method.

eZ/Publish/Core/Repository/UserService.php
```diff
@@ -1348,6 +1320,46 @@ private function comparePasswordHashes(
         return $this->passwordHashService->isValidPassword($plainPassword, $passwordHash, $hashAlgorithm);
     }
 
+    private function executeUserUpdate(APIUser $loadedUser, $userUpdateStruct): void
```
adamwojs, line 1323: Missing type for `userUpdateStruct` parameter.

eZ/Publish/Core/Repository/UserService.php
```diff
@@ -61,6 +61,9 @@
  */
 class UserService implements UserServiceInterface
 {
+    private const MISSING_CONTENT_TYPE_EZUSER_FIELD_TYPE_ERROR_MESSAGE =
+        'The provided Content Type does not contain the ezuser Field Type';
+
```
adamwojs, line 66: Can we introduce dedicated exception class (extending ContentValidationException) for this case? e.g. MissingFieldException

eZ/Publish/Core/Repository/UserService.php
```diff
@@ -731,42 +690,55 @@ public function updateUser(APIUser $user, UserUpdateStruct $userUpdateStruct): A
...
-            $this->userHandler->update(
-                new SPIUser(
-                    [
-                        'id' => $loadedUser->id,
-                        'login' => $loadedUser->login,
-                        'email' => $userUpdateStruct->email ?: $loadedUser->email,
-                    ]
-                )
-            );
+    public function updateUserPassword(APIUser $user, UserUpdateStruct $userUpdateStruct): APIUser
```
adamwojs, line 698: IMHO is worth to introduce User Password Update Struct and not give false impression that it's possible to do more with this method.

adamwojs (commented):
You are missing integration test for API changes. See 
* `\eZ\Publish\API\Repository\Tests\UserServiceTest`
* `\eZ\Publish\API\Repository\Tests\UserServiceAuthorizationTest`

eZ/Publish/Core/Repository/UserService.php
Steveb-p: @adamwojs The reason a new structure was not introduced is because I would consider **not** dispatching `BeforeUserUpdateEvent` and `UserUpdateEvent` to be a BC break, and they rely on `UpdateUserStruct`. Because of this I was also hesitant to dispatch a new event for it.

Which approach is prefered? Should we still force the dispatch of the user update events? Should new `UserPasswordUpdateStruct` be somehow updated from `UserUpdateStruct` if it changes in the event handler?

adamwojs: IMHO you can provide BC in `\eZ\Publish\Core\Event\UserService::updateUserPassword` method. Create UpdateUserStruct from UserUpdatePasswordStruct and dispatch accordingly BeforeUpdateUserEvent/UpdateUserEvent. It's win-win scenario: we will keep BC promise and `updateUserPassword` method signature will be correct. WDYT @alongosz ?

eZ/Publish/Core/Repository/UserService.php
```diff
@@ -731,42 +688,55 @@ public function updateUser(APIUser $user, UserUpdateStruct $userUpdateStruct): A
...
+                'passwordHash' => $user->passwordHash,
+            ])
+        );
+
+        $updateUserCallable = function () use ($loadedUser, $userUpdateStruct): void {
+            $this->executeUserUpdate($loadedUser, $userUpdateStruct);
+        };
+
+        // Perform update without checking other user permissions
+        $this->repository->sudo($updateUserCallable);
```
Nattfarinn, line 738: Do we really need this `sudo`? I am aware of this discussion https://github.com/ezsystems/ezplatform-kernel/pull/117#discussion_r503732769 but still, `sudo` feels like a hack.

What about changes to `owner:self` limitation? Won't it solve the issue and remove the need of `sudo`?

Steveb-p: As discussed, limitation itself will remain unchanged, but instead of calling `$this->repository->sudo()` user password update will be handled by a dedicated SPI method.

eZ/Publish/Core/Repository/UserService.php
Steveb-p: Added `MissingUserFieldTypeException` in 3a0c98b

eZ/Publish/Core/Repository/UserService.php
Steveb-p: Implementation has been changed to omit using `ContentService`. New dependencies have been added to `UserService` as a result and new SPI method has been added.

eZ/Publish/Core/Repository/UserService.php
Steveb-p: `\eZ\Publish\Core\Event\UserService::updateUserPassword` now emits both `UserUpdateEvent` and `UserPasswordUpdateEvent` (as well as `Before...` variants). `UserUpdateEvent` is marked as deprecated inside the method and I think I'll open a separate PR to remove it in 4.0, so we do not forget about it @adamwojs?

eZ/Publish/Core/Repository/UserService.php
```diff
@@ -770,6 +748,108 @@ public function updateUser(APIUser $user, UserUpdateStruct $userUpdateStruct): A
...
+                )
+            );
+
+            $this->repository->commit();
+        } catch (Exception $e) {
+            $this->repository->rollback();
+            throw $e;
+        }
+
+        return $this->loadUser($loadedUser->id);
```
alongosz, line 850: This is a lot of work for a simple password update. Is the only reason for it the validation? Can the validation of password be extracted to a separate validator and used in Content update and here?
This code should read:
1. Validate password
2. Update password

That's it.

eZ/Publish/Core/Repository/UserService.php
```diff
@@ -770,6 +748,108 @@ public function updateUser(APIUser $user, UserUpdateStruct $userUpdateStruct): A
...
+     *
+     * @throws ContentFieldValidationException
+     * @throws MissingUserFieldTypeException
+     * @throws UnauthorizedException
+     * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
+     * @throws \eZ\Publish\API\Repository\Exceptions\ContentValidationException
+     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
+     */
+    public function updateUserPassword(APIUser $user, string $newPassword): APIUser
```
alongosz, line 765: Where's the check that the old password is valid? Expected to see `string $oldPassword` here.

eZ/Publish/SPI/Persistence/User/Handler.php
```diff
@@ -92,6 +92,13 @@ public function loadUserByToken($hash);
      */
     public function update(User $user);
 
+    /**
+     * Update the user password as specified by the user struct.
+     *
+     * @param \eZ\Publish\SPI\Persistence\User $user
+     */
+    public function updatePassword(User $user);
```
alongosz, line 100: missing return type hint

alongosz (changes requested):

eZ/Publish/API/Repository/Tests/UserServiceAuthorizationTest.php
```diff
@@ -297,6 +297,70 @@ public function testUpdateUserThrowsUnauthorizedException()
...
     }
 
+    /**
+     * Test for the updateUserPassword() method.
+     *
+     * @see \eZ\Publish\API\Repository\UserService::updateUserPassword()
+     */
+    public function testUpdateUserPasswordThrowsUnauthorizedException()
+    {
+        $this->expectException(\eZ\Publish\API\Repository\Exceptions\UnauthorizedException::class);
```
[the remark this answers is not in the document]
Steveb-p: Since I'm already working on this file I could do the update. Most of the comments regarding introduced tests result from me trying to keep to style consistent with other in the same file.

@alongosz :+1: / :-1: ?

eZ/Publish/Core/Persistence/Cache/UserHandler.php
```diff
@@ -215,6 +215,23 @@ public function update(User $user)
         return $user;
     }
 
+    public function updatePassword(User $user)
```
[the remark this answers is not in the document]
Steveb-p: Interface that this class implements is missing return type for this API. It should be added even if other methods of this interface do not have them?

eZ/Publish/Core/Persistence/Cache/UserHandler.php
```diff
@@ -215,6 +215,23 @@ public function update(User $user)
...
+    {
+        $this->logger->logCall(__METHOD__, ['struct' => $user]);
```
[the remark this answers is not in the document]
Steveb-p: I wasn't really sure if passing whole user structure into logger was important, so I've left it as-is (after copying). The same goes for `struct` key. I'll adjust accordingly, unless someone goes against it.

eZ/Publish/Core/Persistence/Legacy/User/Gateway/DoctrineDatabase.php
```diff
@@ -121,6 +122,33 @@ public function loadUserByToken(string $hash): array
...
+    {
+        $queryBuilder = $this->connection->createQueryBuilder();
+
+        $queryBuilder
+            ->update($this->connection->quoteIdentifier(self::USER_TABLE))
+            ->set('login', ':login')
+            ->set('email', ':email')
+            ->set('password_hash', ':passwordHash')
+            ->set('password_hash_type', ':passwordHashType')
+            ->set('password_updated_at', ':passwordUpdatedAt')
```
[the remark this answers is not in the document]
Steveb-p: My mistake, I've removed the other update commands but accidentaly left those :disappointed:  
Thanks for noticing.

eZ/Publish/Core/Repository/UserService.php
Steveb-p: IMO it's outside the scope of the user repository class if the developer checked the old password beforehand. In case where this method is called from - for example - password reset functionality, old password would not even be available.

From my point of view, it's the responsibility of the developer to check that kind of security. Even more so since it could not only be related to old password itself.

eZ/Publish/Core/Repository/UserService.php
Steveb-p: > Can the validation of password be extracted to a separate validator and used in Content update and here?

I agree.
Unfortunately I don't really think common code related to user update could be extracted and used in both - `ContentService` / `ContentValidator` do perform very specific checks and this version here already skips a few operations, mainly related to language of the content (which in this case isn't important).

Ideally I would only do checks against the new password, but at the point of writing I wasn't sure if password validation itself could be extracted / injected (circumventing the need to build the whole content update structure).

I'll look into it.

eZ/Publish/Core/Base/Exceptions/MissingUserFieldTypeException.php
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace eZ\Publish\Core\Base\Exceptions;
+
+use eZ\Publish\API\Repository\Values\ContentType\ContentType;
+
+class MissingUserFieldTypeException extends ContentValidationException
```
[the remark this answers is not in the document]
Steveb-p: I was considering which namespace to put this into, and decided to put it alongside `ContentValidationException`, which it extends.

In theory this exception only occurs when `ezuser` field type does not exist, so it should never really happen, I think?

The sole reason to introduce this exception class was to not store an error message constant in `UserService`. See https://github.com/ezsystems/ezplatform-kernel/pull/117#discussion_r505196613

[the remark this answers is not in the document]
Steveb-p: I've made the class `final` and `@internal`, and also made it a little more flexible.

eZ/Publish/API/Repository/Tests/UserServiceAuthorizationTest.php
```diff
@@ -297,6 +297,70 @@ public function testUpdateUserThrowsUnauthorizedException()
         /* END: Use Case */
     }
 
+    /**
+     * Test for the updateUserPassword() method.
+     *
+     * @see \eZ\Publish\API\Repository\UserService::updateUserPassword()
+     */
+    public function testUpdateUserPasswordThrowsUnauthorizedException()
```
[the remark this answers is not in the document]
Steveb-p: Changed.

eZ/Publish/API/Repository/Tests/UserServiceAuthorizationTest.php
```diff
@@ -297,6 +297,70 @@ public function testUpdateUserThrowsUnauthorizedException()
         /* END: Use Case */
     }
 
+    /**
+     * Test for the updateUserPassword() method.
+     *
```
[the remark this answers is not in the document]
Steveb-p: Removed.

eZ/Publish/API/Repository/Tests/UserServiceAuthorizationTest.php
```diff
@@ -297,6 +297,70 @@ public function testUpdateUserThrowsUnauthorizedException()
...
+     * @see \eZ\Publish\API\Repository\UserService::updateUserPassword()
```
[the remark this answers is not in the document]
Steveb-p: Changed.

eZ/Publish/API/Repository/Tests/UserServiceAuthorizationTest.php
```diff
@@ -297,6 +297,70 @@ public function testUpdateUserThrowsUnauthorizedException()
...
+     */
+    public function testUpdateUserPasswordWorksWithUserPasswordRole()
+    {
+        $repository = $this->getRepository();
+        $userService = $repository->getUserService();
+        $permissionResolver = $repository->getPermissionResolver();
+
+        /* BEGIN: Use Case */
```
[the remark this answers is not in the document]
Steveb-p: Removed.

eZ/Publish/API/Repository/Tests/UserServiceAuthorizationTest.php
```diff
@@ -297,6 +297,70 @@ public function testUpdateUserThrowsUnauthorizedException()
...
+            'Anons',
+            'CannotChangePassword'
+        );
+
+        // Now set the currently created "Editor" as current user
+        $permissionResolver->setCurrentUserReference($user);
+
+        // This call will fail with an "UnauthorizedException"
+        $userService->updateUserPassword($user, 'new password');
+        /* END: Use Case */
```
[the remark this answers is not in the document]
Steveb-p: Removed.

eZ/Publish/API/Repository/Tests/UserServiceAuthorizationTest.php
```diff
@@ -297,6 +297,70 @@ public function testUpdateUserThrowsUnauthorizedException()
...
+            'without_role_password',
+            'without_role_password@example.com',
+            'Anons',
+            'CannotChangePassword'
+        );
+
+        // Now set the currently created "Editor" as current user
+        $permissionResolver->setCurrentUserReference($user);
+
+        // This call will fail with an "UnauthorizedException"
```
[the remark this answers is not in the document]
Steveb-p: Moved.

eZ/Publish/API/Repository/UserService.php
```diff
@@ -218,21 +218,39 @@ public function deleteUser(User $user): iterable;
...
      * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException if a field value is not accepted by the field type
      * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException if the authenticated user is not allowed to update the user
      */
     public function updateUser(User $user, UserUpdateStruct $userUpdateStruct): User;
 
+    /**
+     * Updates user's password.
+     *
+     * 4.x: If the versionUpdateStruct is set in the user update structure, this method internally creates a content draft, updates it with the provided data
+     * and publishes the draft. If a draft is explicitly required, the user group can be updated via the content service methods.
```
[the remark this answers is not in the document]
Steveb-p: Comment removed.

eZ/Publish/API/Repository/UserService.php
```diff
@@ -218,21 +218,39 @@ public function deleteUser(User $user): iterable;
...
+     *
+     * @param \eZ\Publish\API\Repository\Values\User\User $user
+     * @param string $newPassword
+     *
+     * @return \eZ\Publish\API\Repository\Values\User\User
```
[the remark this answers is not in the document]
Steveb-p: Removed already explicitly typed parameters / return type.

eZ/Publish/API/Repository/UserService.php
```diff
@@ -218,21 +218,39 @@ public function deleteUser(User $user): iterable;
...
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\ContentFieldValidationException if a field in the $userUpdateStruct is not valid
+     * @throws \eZ\Publish\API\Repository\Exceptions\ContentValidationException if a required field is set empty
```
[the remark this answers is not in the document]
Steveb-p: Removed never thrown exceptions from phpdoc and added description for `ContentFieldValidationException`.

eZ/Publish/Core/Persistence/Cache/UserHandler.php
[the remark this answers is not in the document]
Steveb-p: Anyway, return type added.

eZ/Publish/Core/Persistence/Legacy/User/Handler.php
```diff
@@ -199,6 +199,16 @@ public function update(User $user)
         throw new NotImplementedException('This method should not be called, update is done via content handler.');
     }
 
+    /**
+     * Update the user password as specified by the user struct.
+     *
+     * @param User $user
+     */
+    public function updatePassword(User $user)
```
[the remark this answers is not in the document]
Steveb-p: phpdoc removed and return type added.

eZ/Publish/SPI/Persistence/User/Handler.php
Steveb-p: Added return type.

eZ/Publish/Core/Repository/UserService.php
```diff
@@ -781,52 +762,12 @@ public function updateUserPassword(APIUser $user, string $newPassword): APIUser
...
-                'passwordHash' => $user->passwordHash,
-            ])
-        );
-
-        $errors = $this->contentValidator->validate(
-            $contentUpdateStruct,
-            ['content' => $loadedUser],
-        );
-
+        $errors = $this->passwordValidator->validatePassword($newPassword, $userFieldDefinition);
```
Steveb-p, line 765: `$errors` returned here are instances of `eZ\Publish\SPI\FieldType\ValidationError` instead of `eZ\Publish\Core\FieldType\ValidationError`, but this validator is used internally in the User Type and I don't see any other exception class that would fit. WDYT?

eZ/Publish/Core/Repository/UserService.php
Steveb-p: @alongosz code has been revised in https://github.com/ezsystems/ezplatform-kernel/pull/117/commits/d2dec1a15187e202862d31b62d8341cb28b29df4. Now `updateUserPassword` does not rely on `ContentService` at all. I'm having doubts about how password hashing is done (it's a tad different from `UserPasswordEncoder` I'm more familiar with) so I'd ask you to see if I haven't made a mistake regarding selection of hash algorithm.

eZ/Publish/Core/Repository/UserService.php
alongosz: > IMO it's outside the scope of the user repository class if the developer checked the old password beforehand. In case where this method is called from - for example - password reset functionality, old password would not even be available.
> 
> From my point of view, it's the responsibility of the developer to check that kind of security. Even more so since it could not only be related to old password itself.

But is it really used by reset password or is it just an idea? This is Public API. Public API provides security layer.

eZ/Publish/Core/Repository/UserService.php
alongosz: > `$errors` returned here are instances of `eZ\Publish\SPI\FieldType\ValidationError` instead of `eZ\Publish\Core\FieldType\ValidationError`, but this validator is used internally in the User Type and I don't see any other exception class that would fit. WDYT?

If it's already provided like that, then ok, we can re-use it. Ideally there should be a separate set of exceptions for this, but maybe an overkill...

eZ/Publish/SPI/Persistence/User/Handler.php
```diff
@@ -92,6 +92,13 @@ public function loadUserByToken($hash);
      */
     public function update(User $user);
 
+    /**
+     * Update the user password as specified by the user struct.
+     *
+     * @param \eZ\Publish\SPI\Persistence\User $user
```
alongosz, line 98: can be skipped

eZ/Publish/Core/Repository/UserService.php
Steveb-p: Original `updateUser` method, while allowing the password to be changed, didn't require previous password to be passed, nor had any checks related to it.

At this point in time I password reset still uses `updateUser` method. I think it should also be adjusted once this is integrated into kernel, since both `PasswordChangeController` and `PasswordResetController` relied on the same set of roles and policies, and will be subject the same bug as reported (i.e. user unable to reset password despite having `user/password` policy).

The security check we're discussing here is not really related to user's password - it's about confirming that the logged in user is still the same person making the request. This can be done in more ways than just the old password check - for example by sending a text message with code to one's phone or sending an email with confirmation code. It's only that our eZPlatform meta package uses old password as a security mechanism here.
Limiting `updateUserPassword` to old password check only would make it a liability in the future IMO.

eZ/Publish/Core/Repository/UserService.php
Steveb-p: @alongosz 
From what I've seen Field Types create Core Validation Errors, even though they specify that they would return arrays of SPI Validation Errors (which are not translatable). 
Exception class itself defines Core Validation Errors as acceptable (which excludes their parent SPI variant), but does not complain / does any checks against types. `DomainContentMutationResolver` that is one class that I've found that uses this exception does relies on specific array structure inside the exception (namely, by language codes first) and needs them to be translatable.

What approach do you imagine would be best?

eZ/Publish/API/Repository/UserService.php
```diff
@@ -218,21 +218,39 @@ public function deleteUser(User $user): iterable;
     /**
      * Updates a user.
      *
-     * 4.x: If the versionUpdateStruct is set in the user update structure, this method internally creates a content draft, updates ts with the provided data
+     * 4.x: If the versionUpdateStruct is set in the user update structure, this method internally creates a content draft, updates it with the provided data
```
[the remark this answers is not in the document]
Steveb-p: I'll remove this change as it's out of scope for the PR.

eZ/Publish/API/Repository/Tests/UserServiceAuthorizationTest.php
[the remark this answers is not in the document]
alongosz: > Since I'm already working on this file I could do the update. Most of the comments regarding introduced tests result from me trying to keep to style consistent with other in the same file.

Depends on the diff. If it's too big -> separate PR. However new code should be aligned already, hence my comment

eZ/Publish/Core/Persistence/Cache/UserHandler.php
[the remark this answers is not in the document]
alongosz: > Interface that this class implements is missing return type for this API. It should be added even if other methods of this interface do not have them?

It depends. If we have control over the code, then definitely. In other cases it needs to be assessed (for instance if 3rd party type-hints in PHPDoc e.g. `@return bool` then it can be safely applied as well).

eZ/Publish/Core/Repository/UserService.php
alongosz: > Original `updateUser` method, while allowing the password to be changed, didn't require previous password to be passed, nor had any checks related to it.

Hmm, this is a design inconsistent with what other API methods offer. However...

> The security check we're discussing here is not really related to user's password - it's about confirming that the logged in user is still the same person making the request (authentication), in contrast to what service does (authorization). This can be done in more ways than just the old password check - for example by sending a text message with code to one's phone or sending an email with confirmation code. It's only that our eZPlatform meta package uses old password as a security mechanism here.

This is a good point. It still should be supported by extension point so API is consistent, but this is more complicated topic, out of scope here.

+1

alongosz (approved):

eZ/Publish/API/Repository/Tests/UserServiceAuthorizationTest.php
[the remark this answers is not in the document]
Steveb-p: A separate PR it is then.

tomaszszopinski (approved):
QA approved on eZPlatform-ee 3.1.2 with diff.

lserwatka:
You can merge it up.

PR #123, opened 2020-10-19
EZP-32037: Replaced RenderMethods with Symfony built-in FragmentRenderers

Nattfarinn:
| **JIRA issue**                          | https://jira.ez.no/browse/EZP-32037
| **Type**                                   | bug & improvement
| **Target eZ Platform version** | `v3.2`

No need for custom RenderMethods as we can reuse FragmentRenderer feature from Symfony (no more reinventing the wheel). This also fixes issues with custom solution for ESI rendering. 

#### Checklist:


eZ/Bundle/EzPublishCoreBundle/Fragment/DirectFragmentRenderer.php
```diff
@@ -86,4 +97,38 @@ protected function getArguments(callable $controller, ControllerEvent $event): a
...
+    public function render(
+        $uri,
+        Request $request,
+        array $options = []
+    ): Response {
+        $event = $this->getControllerEvent($request);
+        $controller = $this->getController($event);
+        $arguments = $this->getArguments($controller, $event);
+
+        $response = call_user_func_array($controller, $arguments);
```
Steveb-p, line 115: I'm pretty sure @alongosz would point this out, just as he did recently.

`call_user_func_array` is heavier than using calling a callable directly.

```suggestion
        $response = $controller(...$arguments);
```

eZ/Publish/Core/MVC/Symfony/Templating/RenderContentStrategy.php
```diff
@@ -39,20 +40,28 @@ public function render(ValueObject $valueObject, RenderOptions $options): string
         $currentRequest = $this->requestStack->getCurrentRequest();
         $surrogateCapability = $currentRequest->headers->get('Surrogate-Capability');
 
-        $request = new Request();
+        $request = Request::createFromGlobals();
```
Steveb-p, line 43: If we're actually going to need all the request information, why aren't we using a clone of current request from request stack instead? Possibly also adjusting it so it is identified as a subrequest, instead of master request?

eZ/Publish/SPI/MVC/Templating/BaseRenderStrategy.php
```diff
@@ -29,30 +28,28 @@ abstract class BaseRenderStrategy implements RenderStrategy
...
-                "Method '%s' is not supported for %s.", $method, get_class($valueObject)
-            ));
+        if (!empty($this->fragmentRenderers[$name])) {
+            return $this->fragmentRenderers[$name];
         }
 
-        return $this->renderMethods[$method];
+        throw new InvalidArgumentException('method', sprintf(
+            'Fragment renderer "%s" does not exist.', $name
+        ));
```
Steveb-p, line 53: I actually prefer the original version with exception being thrown as a special condition and return being the result of normal execution flow. This allows new exception conditions to be added down the line, if needed.

eZ/Bundle/EzPublishCoreBundle/Resources/config/templating.yml
```diff
@@ -184,46 +184,22 @@ services:
 
     eZ\Publish\Core\MVC\Symfony\Templating\RenderContentStrategy:
         arguments:
-            $defaultMethod: 'inline'
-            $renderMethods: !tagged_iterator ibexa.platform.render.method
+            $fragmentRenderers: !tagged_iterator kernel.fragment_renderer
+            $defaultRenderer: 'direct'
```
Steveb-p, line 188: We can use class constant instead (since we already have it declared). Double check if my syntax is correct.

```suggestion
            $defaultRenderer: !php/const eZ\Bundle\EzPublishCoreBundle\Fragment\DirectFragmentRenderer::NAME
```

eZ/Bundle/EzPublishCoreBundle/Resources/config/services.yml
```diff
@@ -165,6 +165,17 @@ services:
...
+    eZ\Bundle\EzPublishCoreBundle\Fragment\DirectFragmentRenderer:
+        arguments:
+            $kernel: '@kernel'
+            $controllerListener: '@ezpublish.view_controller_listener'
+            $controllerResolver: '@controller_resolver'
+            $argumentMetadataFactory: '@argument_metadata_factory'
+            $argumentValueResolver: '@argument_resolver.request_attribute'
+            $viewTemplateRenderer: '@ezpublish.view.template_renderer'
+        tags:
+            - { name: kernel.fragment_renderer, alias: direct }
```
Steveb-p, line 177: We can use class constant instead (since we already have it declared). Double check if my syntax is correct.

```suggestion
            - { name: kernel.fragment_renderer, alias: !php/const eZ\Bundle\EzPublishCoreBundle\Fragment\DirectFragmentRenderer::NAME }
```

eZ/Bundle/EzPublishCoreBundle/Resources/config/templating.yml
```diff
@@ -184,46 +184,22 @@ services:
...
             $requestStack: '@request_stack'
         tags:
             - { name: ibexa.platform.render.strategy }
 
     eZ\Publish\Core\MVC\Symfony\Templating\RenderLocationStrategy:
         arguments:
-            $defaultMethod: 'inline'
-            $renderMethods: !tagged_iterator ibexa.platform.render.method
+            $fragmentRenderers: !tagged_iterator kernel.fragment_renderer
+            $defaultRenderer: 'direct'
```
Steveb-p, line 197: We can use class constant instead (since we already have it declared). Double check if my syntax is correct.

```suggestion
            $defaultRenderer: !php/const eZ\Bundle\EzPublishCoreBundle\Fragment\DirectFragmentRenderer::NAME
```

eZ/Bundle/EzPublishCoreBundle/Fragment/DirectFragmentRenderer.php
```diff
@@ -86,4 +97,38 @@ protected function getArguments(callable $controller, ControllerEvent $event): a
...
+        if ($response instanceof Response) {
+            return $response;
+        } elseif ($response instanceof View) {
+            return new Response($this->viewTemplateRenderer->render($response));
+        } elseif (is_string($response)) {
+            return new Response($response);
+        }
+
+        throw new InvalidResponseException(
+            sprintf('Unsupported type (%s)', is_object($response) ? get_class($response) : gettype($response))
```
Steveb-p, line 126: Just would like to add that in PHP 8.0 (and in Symfony polyfill for PHP8: https://github.com/symfony/polyfill-php80) a `get_debug_type` function exists that does this.

eZ/Bundle/EzPublishCoreBundle/Fragment/DirectFragmentRenderer.php
```diff
@@ -86,4 +97,38 @@ protected function getArguments(callable $controller, ControllerEvent $event): a
...
+
+    /**
+     * @param string|\Symfony\Component\HttpKernel\Controller\ControllerReference $uri
+     *
+     * @throws \eZ\Publish\Core\MVC\Symfony\Templating\Exception\InvalidResponseException
+     */
+    public function render(
+        $uri,
+        Request $request,
+        array $options = []
```
Steveb-p, line 109: Is this our own method parameter? It doesn't seem to be used.

Nattfarinn: It comes from interface we use as well. Currently unused in this implementation.

eZ/Publish/Core/MVC/Symfony/Templating/RenderContentStrategy.php
Nattfarinn: All we actually need is base path (whatever would it be) and there is no way to manually add it to the request. Clone would work here as well, but not sure is it worth doing it.

eZ/Bundle/EzPublishCoreBundle/Fragment/DirectFragmentRenderer.php
Nattfarinn: All Symfony polyfills are marked as `internal`. Are you sure it is safe to use them?

Steveb-p: I'm advocating for their use since functions they provide are part of new PHP versions. Cost of their introduction is low (since they are not loaded into memory if they're not used, and if functions already exist - because we're already on new PHP version - they are not loaded at all).

I don't know the reasoning for them being marked internal, but clearly Symfony intends for them to be used.

However, I do understand that adding them here would mean introducing a dependency, even if small, to all projects using ezplatform-kernel.

eZ/Bundle/EzPublishCoreBundle/Fragment/DirectFragmentRenderer.php
```diff
@@ -86,4 +99,69 @@ protected function getArguments(callable $controller, ControllerEvent $event): a
...
+            '_format' => $reference->attributes['_format'] ?? null,
+            '_locale' => $reference->attributes['_locale'] ?? null,
+        ];
+
+        return $this->generateFragmentUri($simplifiedReference, $request, false, false);
+    }
+
+    private function getSubRequest(Request $request): Request
+    {
+    }
```
Steveb-p, line 118: This method is empty and should be removed.

eZ/Bundle/EzPublishCoreBundle/Fragment/DirectFragmentRenderer.php
```diff
@@ -86,4 +99,69 @@ protected function getArguments(callable $controller, ControllerEvent $event): a
...
+        if ($response instanceof Response) {
+            return $response;
+        } elseif ($response instanceof View) {
+            return new Response($this->viewTemplateRenderer->render($response));
+        } elseif (is_string($response)) {
+            return new Response($response);
+        }
+
+        throw new InvalidResponseException(
+            sprintf('Unsupported type (%s)', Php80::get_debug_type($response))
```
Steveb-p, line 159: Just call it exactly as you would in PHP8, otherwise there's no point in using a polyfill.

```suggestion
            sprintf('Unsupported type (%s)', \get_debug_type($response))
```

Steveb-p (changes requested):

eZ/Bundle/EzPublishCoreBundle/Fragment/DirectFragmentRenderer.php
Steveb-p: And, if we're not having `symfony/polyfill-php80` in `composer.json`, then we cannot use it. I was just noting that it's a possibility to use those polyfills.

mnocon (approved):
Tested all rendering methods with custom templates using Symfony proxy and varnish, looks good now!

PR #141, opened 2020-12-02
EZP-31462: Removed password hash types break login

glye:
| **JIRA issue**                          | EZP-31462
| **Target eZ Platform version** | `v3.2`+

The previous fixes are incomplete. When an invalid hash type is detected, we expire the password and ask the user to enter a new password. But when we save this new password, we don't update the hash type, which means the password isn't saved, and the user ends up in a redirect loop.

However, cache can mask the issue. If you test by manually changing hash type to an invalid one in the db, you have to clear caches before you go through the test steps.

This fix catches when the type is invalid, and sets it to the default type in that case.

DOC: Once fixed, I suggest the relevant doc should mention that you have to upgrade to at least version x for this to work.
https://github.com/ezsystems/developer-documentation/blob/03a7cc0ea04f3e6bf7ebf22a8bea53840123196e/docs/releases/ez_platform_v3.0_deprecations.md#password-hashes

#### Checklist:

### DIFF
--- a/eZ/Publish/Core/Repository/UserService.php
+++ b/eZ/Publish/Core/Repository/UserService.php
@@ -740,13 +740,13 @@ class UserService implements UserServiceInterface
     /**
      * Validates and updates just the user's password.
      *
-     * @param \eZ\Publish\API\Repository\Values\User\User $user
-     * @param string $newPassword
+     * @return \eZ\Publish\API\Repository\Values\User\User
      *
-     * @throws \eZ\Publish\Core\Base\Exceptions\ContentFieldValidationException
-     * @throws \eZ\Publish\Core\Base\Exceptions\MissingUserFieldTypeException
-     * @throws \eZ\Publish\Core\Base\Exceptions\UnauthorizedException
+     * @throws \eZ\Publish\API\Repository\Exceptions\ForbiddenException
+     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
+     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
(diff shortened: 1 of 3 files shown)


glye:
Fixup: Get the default hash type from the password hash service, rather than hardcoding it.

NB: It's not at all certain that the final fix will look like this. It may be the wrong place to fix it.

alongosz (changes requested):
Could be, but I'm not gonna accept this w/o either integration or unit tests.

glye:
Fair point. Set as work in progress, and I've asked for Eng help on it. I need to focus on the growing security issue list, and this isn't security. Maybe @kaff will be able to look at this later?  🏃💨

alongosz (approved):

eZ/Publish/Core/Repository/UserService.php
```diff
@@ -739,13 +740,13 @@ public function updateUser(APIUser $user, UserUpdateStruct $userUpdateStruct): A
     /**
      * Validates and updates just the user's password.
      *
-     * @param \eZ\Publish\API\Repository\Values\User\User $user
-     * @param string $newPassword
+     * @return \eZ\Publish\API\Repository\Values\User\User
```
Steveb-p, line 743: nitpick: Unnecessary `@return` declaration - it matches the method's return type afaik.

Steveb-p (approved):

adamwojs (approved):

ciastektk (approved):

micszo (approved):
Retested successfully.
QA Approved on Ibexa Experience v3.2.4 with diff.

PR #148, opened 2020-12-16
EZP-32214: Made purge type configuration using env vars resolvable at compile time

webhdx:
| **JIRA issue**                          | EZP-32214
| **Target eZ Platform version** | `v3.3` - please update `x` accordingly

Since we are using ENV variables to setup HTTP Cache purge driver, it's needed to resolve all envs at compile time. This is rather bad solution because ENVs are not designed to be used this way. In order to make it work we need to call `Container::resolveEnvPlaceholders` to get real values. Previously it was working because `config/packages/overrides/*.php` files were used and they resolved values themselves.

#### Checklist:


ViniTou (approved):

adamwojs (approved):

alongosz (approved):
#goodChange

mnocon (approved):
QA approved.

Please remember to remove TMP changes in https://github.com/ezsystems/ezplatform/pull/632 before merging.

PR #150, opened 2020-12-22
EZP-32250: Exposed PasswordHashService APIs

adamwojs:
| **JIRA issue**                          | EZP-32250

Added methods

* `getSupportedHashTypes(): array`
* `isHashTypeSupported(int $hashType): bool`

to `\eZ\Publish\Core\Repository\User\PasswordHashServiceInterface` and refactor existing `\eZ\Publish\API\Repository\Values\User\User::SUPPORTED_PASSWORD_HASHES` usages

This allows to decorate `PasswordHashService` in 3rd-party bundles and add support for custom password hash types.

#### Checklist:
- [X] Provided PR description.

### DIFF
--- a/eZ/Publish/API/Repository/Values/User/User.php
+++ b/eZ/Publish/API/Repository/Values/User/User.php
@@ -24,6 +24,9 @@ use eZ\Publish\API\Repository\Values\Content\Content;
  */
 abstract class User extends Content implements UserReference
 {
+    /**
+     * @var int[] List of supported (by default) hash types.
+     */
     public const SUPPORTED_PASSWORD_HASHES = [
         self::PASSWORD_HASH_BCRYPT,
         self::PASSWORD_HASH_PHP_DEFAULT,
--- a/eZ/Publish/Core/Repository/User/PasswordHashServiceInterface.php
+++ b/eZ/Publish/Core/Repository/User/PasswordHashServiceInterface.php
@@ -8,12 +8,13 @@ declare(strict_types=1);
 
 namespace eZ\Publish\Core\Repository\User;
 
-/**
- * @internal
- */
 interface PasswordHashServiceInterface
 {
     public function getDefaultHashType(): int;
+
+    public function getSupportedHashTypes(): array;
+
+    public function isHashTypeSupported(int $hashType): bool;
 
     public function createPasswordHash(string $password, ?int $hashType = null): string;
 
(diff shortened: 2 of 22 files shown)


eZ/Publish/Core/Repository/User/PasswordHashServiceInterface.php
```diff
@@ -8,13 +8,14 @@
 
 namespace eZ\Publish\Core\Repository\User;
 
-/**
- * @internal
- */
```
alongosz, line 13: If you don't want it internal, I'd prefer to see it in API.

alongosz (changes requested):

eZ/Publish/Core/Repository/User/PasswordHashServiceInterface.php
```diff
@@ -8,13 +8,14 @@
...
 interface PasswordHashServiceInterface
 {
     public function getDefaultHashType(): int;
 
+    public function getSupportedHashTypes(): array;
```
Steveb-p, line 15: Since other methods specifically ask for `int`.

```suggestion
    /**
     * @return int[]
     */
    public function getSupportedHashTypes(): array;
```

eZ/Publish/Core/Repository/User/PasswordHashServiceInterface.php
Steveb-p: To prevent BC we could `class_alias` it in API repository / namespace to occupy both class names until 4.0?

eZ/Publish/API/Repository/Values/User/User.php
```diff
@@ -24,6 +24,9 @@
  */
 abstract class User extends Content implements UserReference
 {
+    /**
+     * @var int[] List of supported (by default) hash types.
+     */
```
Steveb-p, line 29: Side note: to speed up checks if hash is supported this can be turned into `array<int, int>`. Then to check support for hash you'd just see if:
```php
isset(User::SUPPORTED_PASSWORD_HASHES[$hashType]);
```

adamwojs: > Premature optimization is the root of all evil

D. Knuth

eZ/Publish/Core/Repository/User/PasswordHashServiceInterface.php
adamwojs: Done in e9c6d71. I've marked PasswordHashServiceInterface as deprecated in v3.3 and change it's definition to extend PasswordHashService from PAPI. This should prevent any BC breaks.

alongosz (approved):

Steveb-p (approved):

PR #163, opened 2021-01-22
Fixed EZP-32302: Kosovo and South Sudan are missing in the Country FieldType

konradoboza:
| **JIRA issue**                          | EZP-32302
| **Target eZ Platform version** | `v3.2`, `v3.3`

Excerpt from JIRA:

>The customer observed that Kosovo is missing on Country FieldType. We should also add South Sudan and update Macedonia.

This PR target versions 3.2 and 3.3.

#### Checklist:


konradoboza:
@alongosz not sure whether this change should also be reflected in https://github.com/ezsystems/ezcommerce-base-design/blob/master/ezpublish_legacy/silver.content/settings/selections.ini.append.php.

glye (approved):

alongosz:
> @alongosz not sure whether this change should also be reflected in https://github.com/ezsystems/ezcommerce-base-design/blob/master/ezpublish_legacy/silver.content/settings/selections.ini.append.php.

No, looks like unused Legacy config and Legacy is past EOM.

Regardless, this needs to be handled as a horizontal merge from ezsystems/ezpublish-kernel#3082, not a PR. Feel free to use this branch as a reference, so I'm not going to delete it yet.

PR #181, opened 2021-05-06
IBX-300: Added timezone offset to timestamp

mateuszdebinski:
| **JIRA issue**                          | [IBX-300](https://issues.ibexa.co/browse/IBX-300)
| **Target eZ Platform version** | `v3.2`

Related PR: https://github.com/ezsystems/ezplatform-admin-ui/pull/1752

#### Checklist:

### DIFF
--- a/eZ/Publish/API/Repository/Tests/FieldType/DateIntegrationTest.php
+++ b/eZ/Publish/API/Repository/Tests/FieldType/DateIntegrationTest.php
@@ -375,7 +375,6 @@ class DateIntegrationTest extends SearchBaseIntegrationTest
     protected function getValidSearchValueTwo()
     {
         $dateTime = new DateTime('1970-01-03');
-        $dateTime->setTime(0, 0, 0);
 
         return $dateTime->getTimestamp();
     }
@@ -402,9 +401,6 @@ class DateIntegrationTest extends SearchBaseIntegrationTest
 
     protected function getValueOneDate(): DateTime
     {
-        $dateTime = new DateTime('1970-01-02');
-        $dateTime->setTime(0, 0, 0);
-
-        return $dateTime;
+        return new DateTime('1970-01-02');
     }
 }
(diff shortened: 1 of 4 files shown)


Steveb-p:
I've noticed that `Value::fromTimestamp()` method constructs `DateTime` objects by passing timestamps directly into constructor method. Shouldn't this be changed into:
```diff
    public static function fromTimestamp($timestamp)
    {
        try {
-            return new static(new DateTime("@{$timestamp}"));
+            $datetime = new DateTime();
+            $datetime->setTimestamp($timestamp);
+
+            return new static($datetime);
        } catch (Exception $e) {
            throw new InvalidArgumentValue('$timestamp', $timestamp, __CLASS__, $e);
        }
    }
```

Reason for this is that `DateTime` objects created using timestamps passed to constructor are actually initialized with offsets +0000, instead of offsets coming from `date_default_timezone_get`. This means that `DateTime` objects created this way will have potentially different offsets, therefore generating different values for, for example, `DateTime::format` method.

This might be the reason why those datetimes differ in the first place. Adding default offset would cause that an actually invalid timestamp would be stored (!).

Steveb-p (changes requested):
Please address the issue raised in the comment.

mateuszdebinski:
> Please address the issue raised in the comment.

The proposed change actually works as you describe, but then there is a problem on the front side, so I need to verify the JS files that are responsible for it

alongosz (changes requested):
Looks like we're missing test coverage

Steveb-p (approved):
LGTM. @alongosz do you think we need some more tests for this?

alongosz (approved):

konradoboza (approved):

Steveb-p:
One thing to note: do we need to migrate anyone? Can any customer have timestamps stored with an offset?

mateuszdebinski:
> One thing to note: do we need to migrate anyone? Can any customer have timestamps stored with an offset?

I think no, there shouldn't be any incorrect data saved anywhere, because when creating this item we changed the timestamp (add offset timezone) via JS now we do it in PHP

bogusez (approved):

mateuszdebinski:
merge up - done

PR #194, opened 2021-06-08
IBX-154: Allowed to define if the alternative text for image field is required

adamwojs:
| **JIRA issue**                          | https://issues.ibexa.co/browse/IBX-154
| **Target eZ Platform version** | `v4.0`

#### Checklist:

### DIFF
--- a/eZ/Publish/Core/FieldType/Tests/ImageTest.php
+++ b/eZ/Publish/Core/FieldType/Tests/ImageTest.php
@@ -116,6 +116,12 @@ class ImageTest extends FieldTypeTest
                     'default' => null,
                 ],
             ],
+            'AlternativeTextValidator' => [
+                'required' => [
+                    'type' => 'bool',
+                    'default' => false,
+                ],
+            ],
         ];
     }
 
@@ -814,6 +820,56 @@ class ImageTest extends FieldTypeTest
                     ),
                     new ValidationError(
                         'A valid image file is required.', null, [], 'id'
+                    ),
+                ],
+            ],
+
+            // alternative text is null
+            [
+                [
+                    'validatorConfiguration' => [
+                        'AlternativeTextValidator' => [
+                            'required' => true,
+                        ],
+                    ],
+                ],
+                new ImageValue(
+                    [
+                        'id' => $this->getImageInputPath(),
+                        'fileName' => basename($this->getImageInputPath()),
+                        'fileSize' => filesize($this->getImageInputPath()),
+                        'alternativeText' => null,
+                        'uri' => '',
+                    ]
+                ),
+                [
+                    new ValidationError(
+                        'Alternative text is required.', null, [], 'alternativeText'
+                    ),
+                ],
+            ],
+
+            // alternative text is empty string
+            [
+                [
+                    'validatorConfiguration' => [
+                        'AlternativeTextValidator' => [
+                            'required' => true,
+                        ],
+                    ],
+                ],
+                new ImageValue(
+                    [
+                        'id' => $this->getImageInputPath(),
+                        'fileName' => basename($this->getImageInputPath()),
+                        'fileSize' => filesize($this->getImageInputPath()),
+                        'alternativeText' => '',
+                        'uri' => '',
+                    ]
+                ),
+                [
+                    new ValidationError(
+                        'Alternative text is required.', null, [], 'alternativeText'
                     ),
                 ],
             ],
--- a/eZ/Publish/Core/FieldType/Tests/ImageTest.php
+++ b/eZ/Publish/Core/FieldType/Tests/ImageTest.php
@@ -664,8 +664,7 @@ class ImageTest extends FieldTypeTest
     public function provideInvalidDataForValidate()
     {
         return [
-            // File is too large
-            [
+            'file is too large' => [
                 [
                     'validatorConfiguration' => [
                         'FileSizeValidator' => [
@@ -693,9 +692,7 @@ class ImageTest extends FieldTypeTest
                     ),
                 ],
             ],
-
-            // file is not an image file
-            [
+            'file is not an image file' => [
                 [
                     'validatorConfiguration' => [
                         'FileSizeValidator' => [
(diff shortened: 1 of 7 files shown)


adamwojs:
CI issues seems to be not related to PR changes.

eZ/Publish/Core/Persistence/Legacy/Content/FieldValue/Converter/ImageConverter.php
```diff
@@ -176,6 +179,32 @@ public function toFieldValue(StorageFieldValue $value, FieldValue $fieldValue)
...
+        $storageDef->dataInt2 = (int)($validators['AlternativeTextValidator']['required'] ?? 0);
+    }
+
+    public function toFieldDefinition(StorageFieldDefinition $storageDef, FieldDefinition $fieldDef): void
+    {
+        $fieldDef->fieldTypeConstraints = new FieldTypeConstraints(
+            [
+                'validators' => [
+                    'FileSizeValidator' => [
+                        'maxFileSize' => ($storageDef->dataInt1 != 0
```
ViniTou, line 196: What about strong comparison here?
```suggestion
                        'maxFileSize' => ($storageDef->dataInt1 !== 0
```
Not sure if it's only me but those parentheses are adding this feeling that there is something more here, but they seems redundant?

eZ/Publish/Core/FieldType/Tests/ImageTest.php
```diff
@@ -817,6 +823,56 @@ public function provideInvalidDataForValidate()
...
+                ),
+                [
+                    new ValidationError(
+                        'Alternative text is required.', null, [], 'alternativeText'
+                    ),
+                ],
+            ],
+
+            // alternative text is empty string
+            [
```
Steveb-p, line 853: I think it makes sense for these cases to have this comment as part of the key for the array. Even if the previous are not "keyed", it's still possible PHP-wise to add set keys explicitly here, so previous ones do not need to be updated if you don't want to.

```suggestion
            'alternative text is empty string' => [
```

eZ/Publish/Core/Persistence/Legacy/Content/FieldValue/Converter/ImageConverter.php
```diff
@@ -176,6 +179,32 @@ public function toFieldValue(StorageFieldValue $value, FieldValue $fieldValue)
...
 
+    public function toStorageFieldDefinition(FieldDefinition $fieldDef, StorageFieldDefinition $storageDef): void
+    {
+        $validators = $fieldDef->fieldTypeConstraints->validators;
+
+        $storageDef->dataInt1 = $validators['FileSizeValidator']['maxFileSize'] ?? 0;
+        $storageDef->dataInt2 = (int)($validators['AlternativeTextValidator']['required'] ?? 0);
+    }
+
+    public function toFieldDefinition(StorageFieldDefinition $storageDef, FieldDefinition $fieldDef): void
```
Steveb-p, line 190: As a general remark, those "to*" methods would probably be better if they could return the `FieldDefinition` and `StorageFieldDefinition` respectively. I'm not sure what context they are used in, but I think usually those objects are constructed just before this method and are mutated here?

I.e.
```php
    public function toFieldDefinition(StorageFieldDefinition $storageDef, ?FieldDefintion $fieldDef = null): FieldDefinition
    {
        $fieldDef = $fieldDef ?? new FieldDefinition();
        // ...
        return $fieldDef;
    }
```

I understand that this is part of some sort of an interface, so only wanted to comment on it, nothing more.

Steveb-p (approved):

eZ/Publish/Core/Persistence/Legacy/Content/FieldValue/Converter/ImageConverter.php
Steveb-p: If that was supposed to be a "fix" for a float comparison then `abs($storageDef->dataInt1) < PHP_FLOAT_EPSILON` should be used instead?

ViniTou: I am almost sure that `maxFileSize` is int not float.
And you cant store floats in `dataIntX` properties, right? We have `dataFloatX` for that.

alongosz: > I am almost sure that `maxFileSize` is int not float.
> And you cant store floats in `dataIntX` properties, right? We have `dataFloatX` for that.

`dataIntX` should never contain anything other than int or null.

eZ/Publish/Core/Persistence/Tests/Legacy/FieldValue/Converter/ImageConverterTest.php
```diff
@@ -225,4 +228,142 @@ public function xmlToFieldValueProvider(): array
...
+     */
+    public function testToStorageFieldDefinition(
+        FieldDefinition $fieldDefinition,
+        StorageFieldDefinition $expectedStorageDef
+    ): void {
+        $storageFieldDefinition = new StorageFieldDefinition();
+
+        $this->imageConverter->toStorageFieldDefinition($fieldDefinition, $storageFieldDefinition);
+
+        $this->assertEquals(
```
alongosz, line 243:
```suggestion
        self::assertEquals(
```

eZ/Publish/Core/Persistence/Tests/Legacy/FieldValue/Converter/ImageConverterTest.php
```diff
@@ -225,4 +228,142 @@ public function xmlToFieldValueProvider(): array
...
+     */
+    public function testToFieldDefinition(
+        StorageFieldDefinition $storageDef,
+        FieldDefinition $expectedFieldDefinition
+    ): void {
+        $fieldDefinition = new FieldDefinition();
+
+        $this->imageConverter->toFieldDefinition($storageDef, $fieldDefinition);
+
+        $this->assertEquals(
```
alongosz, line 323:
```suggestion
        self::assertEquals(
```

alongosz (changes requested):
Please format PR title & commit according to the existing convention

adamwojs:
PR updated according to code review suggestions.

ViniTou (approved):

alongosz (approved):

adamwojs:
Changes integrated manually.

PR #197, opened 2021-06-11
IBX-368: Added image path normalization before storing it into the DB

mateuszbieniek:
Merge up PR for https://github.com/ezsystems/ezpublish-kernel/pull/3102


eZ/Publish/SPI/Tests/FieldType/ImageIntegrationTest.php
```diff
@@ -124,6 +128,16 @@ public function getAliasCleanerMock(): MockObject
         return $this->aliasCleanerMock;
     }
 
+    private function getFilePathNormalizerMock(): IO\FilePathNormalizerInterface
+    {
+        if (!isset($this->filePathNormalizer)) {
```
alongosz, line 133: Must have missed it in the initial PR. This is a bad habit semantically, the class member is always "set". What you need to check is if it's `null ===`

alongosz (changes requested):

eZ/Publish/SPI/Tests/FieldType/ImageIntegrationTest.php
Steveb-p: This will not be the case for PHP 7.4 with typed properties.

For
```php
private FilePathNormalizer $filePathNormalizer;
```
those two checks result in different results:
```php
$this->filePathNormalizer === null;
// An error occurs - we're accessing uninitialized property

isset($this->filePathNormalizer);
// Works normally
```

So no, in this case `isset` usage is preferred.

adamwojs (approved):

eZ/Publish/SPI/Tests/FieldType/ImageIntegrationTest.php
alongosz: > This will not be the case for PHP 7.4 with typed properties.
> 
> For
> 
> ```
> private FilePathNormalizer $filePathNormalizer;
> ```
> 
> those two checks result in different results:
> 
> ```
> $this->filePathNormalizer === null;
> // An error occurs - we're accessing uninitialized property
> 
> isset($this->filePathNormalizer);
> // Works normally
> ```
> 
> So no, in this case `isset` usage is preferred.

Interesting change. Thanks for pointing this out @Steveb-p.

alongosz (approved):

PR #210, opened 2021-07-02
Added \eZ\Publish\API\Repository\Values\Filter\Filter::{withOffset,withLimit} methods

adamwojs:
| **JIRA issue**                          | -
| **Type**                                   | i
| **Doc needed**                       | ?

Added `withOffset` and `withLimit` metods to `\eZ\Publish\API\Repository\Values\Filter\Filter`. 

Primary use case is batch processing using Filtering API. In such scenario limit (batch size) is set at the beginning of data processing and only offset (data pointer) is changing over iteration. 

Currently with `\eZ\Publish\API\Repository\Values\Filter\Filter::sliceBy(...)`  developer is forced to set both limit and offset at once.

#### Checklist:
- [X] Provided PR description.
- [X] Provided automated test coverage.
- [X] Asked for a review (ping `@ezsystems/php-dev-team`).

### DIFF
--- a/eZ/Publish/API/Repository/Values/Filter/Filter.php
+++ b/eZ/Publish/API/Repository/Values/Filter/Filter.php
@@ -62,7 +62,7 @@ final class Filter
     /**
      * Reset Filter so it can be built from scratch.
      */
-    public function reset(): Filter
+    public function reset(): self
     {
         $this->criterion = null;
         $this->sortClauses = [];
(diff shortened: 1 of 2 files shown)


eZ/Publish/API/Repository/Values/Filter/Filter.php
```diff
@@ -143,6 +143,20 @@ public function withSortClause(FilteringSortClause $sortClause): Filter
         return $this;
     }
 
+    public function withOffset(int $offset): Filter
```
Steveb-p, line 146: We don't use it in the other methods, so I don't mind, but why not 

```suggestion
    public function withOffset(int $offset): self
```

? :smile:

ciastektk (approved):

Steveb-p (approved):

alongosz (approved):

lserwatka (approved):

eZ/Publish/API/Repository/Values/Filter/Filter.php
adamwojs: Fixed in https://github.com/ezsystems/ezplatform-kernel/pull/210/commits/7f792aee07fdec45bac26e42c7e56a67de6a4aeb

PR #211, opened 2021-07-06
IBX-719: Introduced BatchIterator

adamwojs:
| **JIRA issue**                          | https://issues.ibexa.co/browse/IBX-719

`BatchIterator`  allows iterate over result sets (e.g. content objects, locations) which cannot be loaded into memory at once.

Examples from existing code base:

https://github.com/ezsystems/ezplatform-kernel/blob/4c840747ef4a75782becc7dac987ff615775d242/eZ/Bundle/EzPublishCoreBundle/Command/CheckURLsCommand.php#L90-L108

https://github.com/ezsystems/ezplatform-kernel/blob/4c840747ef4a75782becc7dac987ff615775d242/eZ/Bundle/EzPublishCoreBundle/Command/ResizeOriginalImagesCommand.php#L210-L220
 
### Usage

```php
$filter = new Filter();
// ... add criteria, sort clauses etc.

$iterator = new BatchIterator(new ContentFilteringAdapter($contentService, $filter));
foreach ($iterator as $content) {
   // process content object
}
```

### Available adapters

All built-in adapters are places under `\eZ\Publish\API\Repository\Iterator\BatchIteratorAdapter\` namespace.

| Adapter                    | Method                                                      |
|----------------------------|-------------------------------------------------------------|
| `ContentFilteringAdapter`  | `\eZ\Publish\API\Repository\ContentService::find`           |
| `ContentInfoSearchAdapter` | `\eZ\Publish\API\Repository\SearchService::findContentInfo` |
| `ContentSearchAdapter`     | `\eZ\Publish\API\Repository\SearchService::findContent`     |
| `LocationFilteringAdapter` | `\eZ\Publish\API\Repository\LocationService::find`          |
| `LocationSearchAdapter`    | `\eZ\Publish\API\Repository\SearchService::findLocations`   |

#### Checklist:

### DIFF
--- a/eZ/Publish/API/Repository/Iterator/BatchIterator.php
+++ b/eZ/Publish/API/Repository/Iterator/BatchIterator.php
@@ -89,6 +89,6 @@ final class BatchIterator implements Iterator
 
     private function isInitialized(): bool
     {
-        return $this->innerIterator !== null;
+        return isset($this->innerIterator);
     }
 }
--- a/eZ/Publish/API/Repository/Iterator/BatchIterator.php
+++ b/eZ/Publish/API/Repository/Iterator/BatchIterator.php
@@ -53,7 +53,7 @@ final class BatchIterator implements Iterator
         ++$this->position;
         $this->innerIterator->next();
         if (!$this->innerIterator->valid() && ($this->position % $this->batchSize) === 0) {
-            $this->innerIterator = $this->adapter->fetch($this->position, $this->batchSize);
+            $this->innerIterator = $this->fetch();
         }
     }
 
@@ -83,12 +83,17 @@ final class BatchIterator implements Iterator
 
     private function initialize(): void
     {
-        $this->innerIterator = $this->adapter->fetch(0, $this->batchSize);
         $this->position = 0;
+        $this->innerIterator = $this->fetch();
     }
 
     private function isInitialized(): bool
     {
         return isset($this->innerIterator);
+    }
+
+    private function fetch(): Iterator
+    {
+        return $this->adapter->fetch($this->position, $this->batchSize);
     }
 }
--- a/eZ/Publish/API/Repository/Iterator/BatchIterator.php
+++ b/eZ/Publish/API/Repository/Iterator/BatchIterator.php
@@ -1,0 +1,94 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace eZ\Publish\API\Repository\Iterator;
+
+use Iterator;
+
+final class BatchIterator implements Iterator
+{
+    public const DEFAULT_BATCH_SIZE = 25;
+
+    /** @var \eZ\Publish\API\Repository\Iterator\BatchIteratorAdapter */
+    private $adapter;
+
+    /** @var Iterator|null */
+    private $innerIterator;
+
+    /** @var int */
+    private $batchSize;
+
+    /** @var int */
+    private $position;
+
+    public function __construct(
+        BatchIteratorAdapter $adapter,
+        int $batchSize = self::DEFAULT_BATCH_SIZE
+    ) {
+        $this->adapter = $adapter;
+        $this->batchSize = $batchSize;
+        $this->position = 0;
+    }
+
+    public function current()
+    {
+        if (!$this->isInitialized()) {
+            $this->initialize();
+        }
+
+        return $this->innerIterator->current();
+    }
+
+    public function next(): void
+    {
+        if (!$this->isInitialized()) {
+            $this->initialize();
+        }
+
+        ++$this->position;
+        $this->innerIterator->next();
+        if (!$this->innerIterator->valid()) {
+            $this->innerIterator = $this->adapter->fetch($this->position, $this->batchSize);
+        }
+    }
+
+    public function key(): int
+    {
+        return $this->position;
+    }
+
+    public function valid(): bool
+    {
+        if (!$this->isInitialized()) {
+            $this->initialize();
+        }
+
+        return $this->innerIterator->valid();
+    }
+
+    public function rewind(): void
+    {
+        $this->initialize();
+    }
+
+    public function setBatchSize(int $batchSize): void
+    {
+        $this->batchSize = $batchSize;
+    }
+
+    private function initialize(): void
+    {
+        $this->innerIterator = $this->adapter->fetch(0, $this->batchSize);
+        $this->position = 0;
+    }
+
+    private function isInitialized(): bool
+    {
+        return $this->innerIterator !== null;
+    }
+}
--- a/eZ/Publish/API/Repository/Iterator/BatchIterator.php
+++ b/eZ/Publish/API/Repository/Iterator/BatchIterator.php
@@ -52,7 +52,7 @@ final class BatchIterator implements Iterator
 
         ++$this->position;
         $this->innerIterator->next();
-        if (!$this->innerIterator->valid()) {
+        if (!$this->innerIterator->valid() && ($this->position % $this->batchSize) === 0) {
             $this->innerIterator = $this->adapter->fetch($this->position, $this->batchSize);
         }
     }
--- a/eZ/Publish/API/Repository/Tests/Iterator/BatchIteratorAdapter/AbstractSearchAdapterTest.php
+++ b/eZ/Publish/API/Repository/Tests/Iterator/BatchIteratorAdapter/AbstractSearchAdapterTest.php
@@ -1,0 +1,75 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace eZ\Publish\API\Repository\Tests\Iterator\BatchIteratorAdapter;
+
+use eZ\Publish\API\Repository\Iterator\BatchIteratorAdapter\AbstractSearchAdapter;
+use eZ\Publish\API\Repository\SearchService;
+use eZ\Publish\API\Repository\Values\Content\Query;
+use eZ\Publish\API\Repository\Values\Content\Query\Criterion\MatchAll;
+use eZ\Publish\API\Repository\Values\Content\Search\SearchResult;
+use Iterator;
+use PHPUnit\Framework\TestCase;
+
+abstract class AbstractSearchAdapterTest extends TestCase
+{
+    protected const EXAMPLE_LANGUAGE_FILTER = [
+        'languages' => ['eng-GB', 'pol-PL'],
+        'useAlwaysAvailable' => true,
+    ];
+
+    protected const EXAMPLE_OFFSET = 7;
+    protected const EXAMPLE_LIMIT = 13;
+
+    final public function testFetch(): void
+    {
+        $expectedIterator = $this->createMock(Iterator::class);
+
+        $searchResults = $this->createMock(SearchResult::class);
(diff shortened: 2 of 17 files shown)


eZ/Publish/API/Repository/Iterator/BatchIterator.php
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public function current()
+    {
+        return $this->current;
+    }
+
+    public function next(): void
+    {
+        if ($this->current === null) {
```
Steveb-p, line 47:
```suggestion
        if ($this->innerIterator === null || $this->innerIterator->valid() === false) {
```

eZ/Publish/API/Repository/Iterator/BatchIterator.php
```diff
@@ -0,0 +1,76 @@
...
+            $this->innerIterator = $this->adapter->fetch($this->position, $this->batchSize);
```
Steveb-p, line 48: Moving iterator initialization (or re-initialization) into a separate method would allow you to ensure that `IteratorAggregate` (the second half of `Traversable` union) is converted into `Iterator` here.

This would remove the need to reduce the scope of `SearchResult::getIterator()` (currently `Traversable`, this PR limits it to `Iterator`).

eZ/Publish/API/Repository/Iterator/BatchIterator.php
```diff
@@ -0,0 +1,76 @@
...
+        $this->adapter = $adapter;
+        $this->batchSize = $batchSize;
+    }
+
+    public function current()
+    {
+        return $this->current;
+    }
+
+    public function next(): void
```
Steveb-p, line 45: `Iterator::next()` is not always called first.

See https://www.php.net/manual/en/class.iterator.php for `foreach` loop:

```php
string(18) "myIterator::rewind"
string(17) "myIterator::valid"
string(19) "myIterator::current"
string(15) "myIterator::key"
int(0)
string(12) "firstelement"

string(16) "myIterator::next"
string(17) "myIterator::valid"
string(19) "myIterator::current"
string(15) "myIterator::key"
int(1)
string(13) "secondelement"

string(16) "myIterator::next"
string(17) "myIterator::valid"
string(19) "myIterator::current"
string(15) "myIterator::key"
int(2)
string(11) "lastelement"

string(16) "myIterator::next"
string(17) "myIterator::valid"
```

So you have to assume that if `Iterator::current()`, `Iterator::key()` or `Iterator::valid()` methods are called, then inner iterator should be implicitly initialized if it's not already.

adamwojs:
PR updated according to code review suggestions.

eZ/Publish/API/Repository/Tests/Iterator/BatchIteratorAdapter/AbstractSearchAdapterTest.php
```diff
@@ -0,0 +1,75 @@
...
+        $searchService = $this->createMock(SearchService::class);
+        $searchService
+            ->expects($this->once())
+            ->method($this->getExpectedFindMethod())
+            ->with($expectedQuery, self::EXAMPLE_LANGUAGE_FILTER, true)
+            ->willReturn($searchResults);
+
+        $adapter = $this->createAdapterUnderTest($searchService, $originalQuery, self::EXAMPLE_LANGUAGE_FILTER, true);
+
+        $this->assertEquals(
```
Steveb-p, line 53: I'm almost certain that this will be the same `Iterator` instance, so we can use `assertSame` in this case (and it makes more sense since you're comparing an empty, mocked iterator).

adamwojs: Fixed in https://github.com/ezsystems/ezplatform-kernel/pull/211/commits/dec2587157cb1b24fd7ebd111daa9b80bd7145e8

eZ/Publish/API/Repository/Iterator/BatchIterator.php
```diff
@@ -0,0 +1,94 @@
...
+    {
+        if (!$this->isInitialized()) {
+            $this->initialize();
+        }
+
+        ++$this->position;
+        $this->innerIterator->next();
+        if (!$this->innerIterator->valid() && ($this->position % $this->batchSize) === 0) {
```
Steveb-p, line 55: This seems a little odd. The second check assumes that in some cases the `Iterator` supplied by the `adapter::fetch` method may contain more items than batch size?

Why do we even care about it? If we somehow ended up with Iterator that contains more items than it's supposed to, we would continue iterating until it's exhausted - and then we would simple fetch using the `position` property, just like it's already done.

eZ/Publish/API/Repository/Iterator/BatchIterator.php
```diff
@@ -0,0 +1,94 @@
...
+
+    public function setBatchSize(int $batchSize): void
+    {
+        $this->batchSize = $batchSize;
+    }
+
+    private function initialize(): void
+    {
+        $this->innerIterator = $this->adapter->fetch(0, $this->batchSize);
+        $this->position = 0;
```
Steveb-p, line 87:
```suggestion
        $this->position = 0;
        $this->innerIterator = $this->adapter->fetch($this->position, $this->batchSize);
```

Steveb-p: You could also unpack the `IteratorAggregate` here:
```php
$iterator = $this->adapter->fetch(...);

if ($iterator instanceof IteratorAggregate) {
    $iterator = $iterator->getIterator();
}

$this->innerIterator = $iterator;
```

This way change to `SearchResult` would become unnecessary and we could accept `Traversable`s.

adamwojs: > You could also unpack the IteratorAggregate here:

This is responsibility of \eZ\Publish\API\Repository\Iterator\BatchIteratorAdapter implementation.

eZ/Publish/API/Repository/Iterator/BatchIterator.php
adamwojs: > The second check assumes that in some cases the Iterator supplied by the adapter::fetch method may contain more items than batch size?

No, `($this->position % $this->batchSize) === 0` allows to avoid additional `\eZ\Publish\API\Repository\Iterator\BatchIteratorAdapter::fetch` call (in average case) when we reach last item (number of items on last page < batch size).

Steveb-p: Makes sense. Now I get it. :+1:

eZ/Publish/API/Repository/Tests/Iterator/BatchIteratorTest.php
```diff
@@ -0,0 +1,43 @@
...
+use eZ\Publish\API\Repository\Iterator\BatchIterator;
+use PHPUnit\Framework\TestCase;
+
+final class BatchIteratorTest extends TestCase
+{
+    public function testIterateOverDummyResultSet(): void
+    {
+        $expectedData = range(1, 100);
+
+        $iterator = new BatchIterator(new BatchIteratorTestAdapter($expectedData));
```
Steveb-p, line 20: Mock the adapter instead. We're interested in counting `fetch` calls as well, since you've got that optimization to prevent `fetch` calls if `batchSize` wasn't reached and Iterator became exhausted.

Steveb-p: Maybe something along the lines:

```php
$mock = $this->createMock(BatchIteratorAdapter::class);
$mock->method('fetch')->willReturnCallback(function ($offset, $limit): \Iterator {
    return function () use ($offset, $limit) {
        $realLimit = $offset + $limit;
        while ($offset < $realLimit) {
            yield $offset;
            $offset++;
        }
    }
})
```

So you get the iterator splitting for fetches without extra effort?

adamwojs: Applied in https://github.com/ezsystems/ezplatform-kernel/pull/211/commits/a0df5d8c4e7a4f9e9d23f7a349aa150c45ab7f07 by introducing fetch coutner on test adapter

Steveb-p (approved):
Added language phpdocs based on usage found in PR.

:+1:

Steveb-p:
I think sonar code style checks can be discarded: it complains that a local variable has the same name as class property.

ViniTou (approved):

eZ/Publish/API/Repository/Iterator/BatchIterator.php
Steveb-p: Ok.

If we could use the suggested code variant as above (with reversed property assignment and `fetch` using `$this->position`, then
```php
$this->innerIterator = $this->adapter->fetch($this->position, $this->batchSize);
```
Could be moved to a separate method (like `doFetch`) that will do exactly as above, since all invocations of `Adapter::fetch` will be the same.

eZ/Publish/API/Repository/Iterator/BatchIterator.php
```diff
@@ -0,0 +1,94 @@
...
+ */
+declare(strict_types=1);
+
+namespace eZ\Publish\API\Repository\Iterator;
+
+use Iterator;
+
+final class BatchIterator implements Iterator
+{
+    public const DEFAULT_BATCH_SIZE = 25;
```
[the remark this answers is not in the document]
adamwojs: No, but there is valid use case to keep it public. See https://github.com/ibexa/installer/pull/30/commits/50841b962bc2e666cf0036b11738a5503ad44d1e

eZ/Publish/API/Repository/Iterator/BatchIterator.php
adamwojs: Fixed in https://github.com/ezsystems/ezplatform-kernel/pull/211/commits/80a96f0ebb2df56009d300d9365220555e58ac76

eZ/Publish/API/Repository/Tests/Iterator/BatchIteratorAdapter/ContentFilteringAdapterTest.php
```diff
@@ -0,0 +1,58 @@
...
+        $contentService = $this->createMock(ContentService::class);
+        $contentService
+            ->expects($this->once())
+            ->method('find')
+            ->with($expectedFilter, self::EXAMPLE_LANGUAGE_FILTER)
+            ->willReturn($contentList);
+
+        $adapter = new ContentFilteringAdapter($contentService, $originalFilter, self::EXAMPLE_LANGUAGE_FILTER);
+
+        $this->assertEquals(
```
Steveb-p, line 49: I think those iterators will be the same instances.

```suggestion
        $this->assertSame(
```

eZ/Publish/API/Repository/Tests/Iterator/BatchIteratorAdapter/LocationFilteringAdapterTest.php
```diff
@@ -0,0 +1,61 @@
...
+        $locationService = $this->createMock(LocationService::class);
+        $locationService
+            ->expects($this->once())
+            ->method('find')
+            ->with($expectedFilter, self::EXAMPLE_LANGUAGE_FILTER)
+            ->willReturn($locationList);
+
+        $adapter = new LocationFilteringAdapter($locationService, $originalFilter, self::EXAMPLE_LANGUAGE_FILTER);
+
+        $this->assertEquals(
```
Steveb-p, line 52: I think those iterators will be the same instances.

```suggestion
        $this->assertSame(
```

Steveb-p:
Both code smells detected by SonarCloud should probably be marked as "won't fix", since they are about variable that shares the name with property that they are `clone`d from.

micszo (approved):
QA Approved on Ibexa Experience 3.3.6-dev with branch.

PR #212, opened 2021-07-06
IBX-764: Added Kernel / Repository Test API

Steveb-p:
| **JIRA issue**                          | [IBX-764](https://issues.ibexa.co/browse/IBX-764)
| **Doc needed**                       | ~yes/no - _depending on whether we feel ok with exposing the test base right now_~ **No, only internal use**

Adds `IbexaTestKernel` & `IbexaKernelTestCase` as base for other packages for wholesome integration tests.

The main selling point of those tests is that they also test our service declarations. They are also easy to expand upon by adding more bundles / services into the test Kernel, and makes use of built-in service compilation (which makes the tests faster).

Potentially, by using `WebTestCase` from Symfony, even routes can be tested.

**~Addition of MySQL and Postgres tests will be done in a following PR~**

~Current integration tests remain unchanged and will be adjusted in following PRs.~

### Supports Postgres integration tests

Due to issues with MySQL / MariaDB images those tests can be enabled in a separate PR.

### Security hotspots reported by Sonar

False-positives. They're related to `md5` usage when creating a hash for temporary directory that stores compiled Kernel.

### Other changes

To allow tests with different environments to produce different Kernels, whole environment is used to calculate cache hash. This allows for multiple tests to be run on a local machine, without issues coming from Kernel cache file re-use.

#### Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -66,7 +66,8 @@
         "ezsystems/ezplatform-code-style": "^0.2",
         "phpunit/phpunit": "^8.2",
         "matthiasnoback/symfony-dependency-injection-test": "^4.1",
-        "symfony/phpunit-bridge": "^5.1"
+        "symfony/phpunit-bridge": "^5.1",
+        "symfony/proxy-manager-bridge": "^5.3"
     },
     "conflict": {
         "symfony/security-core": "5.3.0",
--- a/src/contracts/Test/IbexaKernelTestCase.php
+++ b/src/contracts/Test/IbexaKernelTestCase.php
@@ -35,10 +35,6 @@ abstract class IbexaKernelTestCase extends KernelTestCase
 
     final protected static function loadSchema(): void
     {
-        if (!self::$booted) {
-            self::bootKernel();
-        }
-
         /** @var \eZ\Publish\API\Repository\Tests\LegacySchemaImporter $schemaImporter */
         $schemaImporter = self::getContainer()->get(LegacySchemaImporter::class);
         foreach (static::getSchemaFiles() as $schemaFile) {
@@ -56,10 +52,6 @@ abstract class IbexaKernelTestCase extends KernelTestCase
 
     final protected static function loadFixtures(): void
     {
-        if (!self::$booted) {
-            self::bootKernel();
-        }
-
         /** @var \eZ\Publish\SPI\Tests\Persistence\FixtureImporter $fixtureImporter */
         $fixtureImporter = self::getContainer()->get(FixtureImporter::class);
         foreach (static::getFixtures() as $fixture) {
--- a/src/contracts/Test/IbexaKernelTestCase.php
+++ b/src/contracts/Test/IbexaKernelTestCase.php
@@ -26,6 +26,9 @@ use eZ\Publish\SPI\Tests\Persistence\YamlFixture;
 use RuntimeException;
 use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
 
+/**
+ * @experimental
+ */
 abstract class IbexaKernelTestCase extends KernelTestCase
 {
     protected static function getKernelClass(): string
--- a/src/contracts/Test/IbexaKernelTestCase.php
+++ b/src/contracts/Test/IbexaKernelTestCase.php
@@ -89,22 +89,31 @@ abstract class IbexaKernelTestCase extends KernelTestCase
      */
     final protected static function getServiceByClassName(string $className, ?string $id = null): object
     {
-        $kernelClass = static::getKernelClass();
-        if (!is_a($kernelClass, IbexaTestKernel::class, true)) {
+        if (!self::$booted) {
+            static::bootKernel();
+        }
+
+        $serviceId = self::getTestServiceId($id, $className);
+        $service = self::getContainer()->get($serviceId);
+        assert(is_object($service) && is_a($service, $className));
+
+        return $service;
+    }
+
+    protected static function getTestServiceId(?string $id, string $className): string
+    {
+        $kernel = self::$kernel;
+        if (!$kernel instanceof IbexaTestKernel) {
             throw new RuntimeException(sprintf(
                 'Expected %s to be an instance of %s.',
-                $kernelClass,
+                get_class($kernel),
                 IbexaTestKernel::class,
             ));
         }
 
         $id = $id ?? $className;
-        $serviceId = $kernelClass::getAliasServiceId($id);
 
-        $service = self::getContainer()->get($serviceId);
-        assert(is_object($service) && is_a($service, $className));
-
-        return $service;
+        return $kernel->getAliasServiceId($id);
     }
 
     protected static function getDoctrineConnection(): Connection
--- a/src/contracts/Test/IbexaKernelTestCase.php
+++ b/src/contracts/Test/IbexaKernelTestCase.php
@@ -1,0 +1,169 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace Ibexa\Contracts\Core\Test;
+
+use Doctrine\DBAL\Connection;
+use eZ\Publish\API\Repository\ContentService;
+use eZ\Publish\API\Repository\ContentTypeService;
+use eZ\Publish\API\Repository\LanguageService;
+use eZ\Publish\API\Repository\LocationService;
+use eZ\Publish\API\Repository\ObjectStateService;
+use eZ\Publish\API\Repository\PermissionResolver;
+use eZ\Publish\API\Repository\RoleService;
+use eZ\Publish\API\Repository\SearchService;
+use eZ\Publish\API\Repository\SectionService;
+use eZ\Publish\API\Repository\Tests\LegacySchemaImporter;
+use eZ\Publish\API\Repository\UserService;
+use eZ\Publish\SPI\Persistence\TransactionHandler;
+use eZ\Publish\SPI\Tests\Persistence\FixtureImporter;
+use eZ\Publish\SPI\Tests\Persistence\YamlFixture;
+use RuntimeException;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+abstract class IbexaKernelTestCase extends KernelTestCase
+{
+    protected static function getKernelClass(): string
+    {
+        return IbexaTestKernel::class;
+    }
+
+    final protected static function loadSchema(): void
+    {
+        if (!self::$booted) {
+            self::bootKernel();
+        }
+
+        /** @var \eZ\Publish\API\Repository\Tests\LegacySchemaImporter $schemaImporter */
+        $schemaImporter = self::getContainer()->get(LegacySchemaImporter::class);
+        foreach (static::getSchemaFiles() as $schemaFile) {
+            $schemaImporter->importSchema($schemaFile);
+        }
+    }
+
+    /**
+     * @return array<string>
+     */
+    protected static function getSchemaFiles(): iterable
+    {
+        yield self::$kernel->locateResource('@EzPublishCoreBundle/Resources/config/storage/legacy/schema.yaml');
+    }
+
+    final protected static function loadFixtures(): void
+    {
+        if (!self::$booted) {
+            self::bootKernel();
+        }
+
+        /** @var \eZ\Publish\SPI\Tests\Persistence\FixtureImporter $fixtureImporter */
+        $fixtureImporter = self::getContainer()->get(FixtureImporter::class);
+        foreach (static::getFixtures() as $fixture) {
+            $fixtureImporter->import($fixture);
+        }
+
+        static::postLoadFixtures();
+    }
+
+    protected static function postLoadFixtures(): void
+    {
+    }
+
+    /**
+     * @return iterable<\eZ\Publish\SPI\Tests\Persistence\Fixture>
+     */
+    protected static function getFixtures(): iterable
+    {
+        yield new YamlFixture(dirname(__DIR__, 3) . '/eZ/Publish/API/Repository/Tests/_fixtures/Legacy/data/test_data.yaml');
+    }
+
+    /**
+     * @template T of object
+     * @phpstan-param class-string<T> $className
+     *
+     * @return T
+     */
+    final protected static function getServiceByClassName(string $className, ?string $id = null): object
+    {
+        $kernelClass = static::getKernelClass();
+        if (!is_a($kernelClass, IbexaTestKernel::class, true)) {
+            throw new RuntimeException(sprintf(
+                'Expected %s to be an instance of %s.',
+                $kernelClass,
+                IbexaTestKernel::class,
+            ));
+        }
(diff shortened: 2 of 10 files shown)


src/contracts/Test/IbexaTestKernel.php
```diff
@@ -0,0 +1,184 @@
...
+    private static function prepareDatabaseConnection(ContainerBuilder $container): void
+    {
+        $container->loadFromExtension('doctrine', [
+            'dbal' => [
+                'memory' => true,
+                'driver' => 'pdo_sqlite',
+                'platform_service' => SqliteDbPlatform::class,
+                'logging' => false,
+            ],
+        ]);
```
alongosz, line 94: SQLite integration exists only to speed up tests. The real value is an ability to test on MySQL and PostgreSQL.

src/contracts/Test/IbexaTestKernel.php
```diff
@@ -0,0 +1,184 @@
...
+
+        $definition = new Definition(Repository\Tests\LegacySchemaImporter::class);
+        $definition->setPublic(true);
+        $definition->setArgument(0, new Reference('doctrine.dbal.default_connection'));
+        $container->setDefinition('test.ibexa.migrations.schema_importer', $definition);
+
+        $definition = new Definition(FixtureImporter::class);
+        $definition->setPublic(true);
+        $definition->setArgument(0, new Reference('doctrine.dbal.default_connection'));
+        $container->setDefinition('test.ibexa.migrations.fixture_importer', $definition);
```
alongosz, line 104: Is it possible to replace this with Yaml configuration?

src/contracts/Test/IbexaTestKernel.php
```diff
@@ -0,0 +1,184 @@
...
+                'app' => 'cache.adapter.array',
+            ],
+            'router' => [
+                'resource' => 'foo',
+            ],
+        ]);
+
+        $container->setAlias(
+            'eZ\Publish\Core\MVC\ConfigResolverInterface',
+            'eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\ChainConfigResolver'
```
alongosz, line 168: Can't we use `::class` here?

src/contracts/Test/IbexaTestKernel.php
```diff
@@ -0,0 +1,184 @@
...
+            'session' => [
+                'storage_id' => 'session.storage.mock_file',
+            ],
+            'cache' => [
+                'app' => 'cache.adapter.array',
+            ],
+            'router' => [
+                'resource' => 'foo',
+            ],
+        ]);
```
alongosz, line 164: It's best that we load this configuration from Yaml files. It would allow to optimize the spaghetti code here.

alongosz (changes requested):
I need to see a consumption example by 1st party package.
The major issue however I have with this solution is [a cyclic dependency on `ezplatform-core`](https://github.com/ezsystems/ezplatform-core/blob/v2.3.5/composer.json#L28). While it might not be visible atm, it always causes composer installation issues at some point.

src/contracts/Test/IbexaKernelTestCase.php
```diff
@@ -0,0 +1,101 @@
...
+use eZ\Publish\API\Repository\RoleService;
+use eZ\Publish\API\Repository\SearchService;
+use eZ\Publish\API\Repository\SectionService;
+use eZ\Publish\API\Repository\UserService;
+use eZ\Publish\SPI\Persistence\TransactionHandler;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+abstract class IbexaKernelTestCase extends KernelTestCase
+{
+    protected static function getKernelClass()
```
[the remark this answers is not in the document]
Steveb-p: I was always reluctant to add types to methods that only declare their intent via phpdoc, just to be on the safe side.

:+1:

composer.json
```diff
@@ -67,7 +67,9 @@
         "ezsystems/ezplatform-code-style": "^0.2",
         "phpunit/phpunit": "^8.2",
         "matthiasnoback/symfony-dependency-injection-test": "^4.1",
-        "symfony/phpunit-bridge": "^5.1"
+        "symfony/phpunit-bridge": "^5.1",
+        "ezsystems/ezplatform-core": "^2.3@dev",
```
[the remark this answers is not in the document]
Steveb-p: Issue here is that without bundle coming from this package, trying to compile the kernel results in the following error:

```
1) Ibexa\Tests\Integration\Core\BasicKernelTest::testBasicKernelCompiles
Symfony\Component\DependencyInjection\Exception\LogicException: Container extension "ezplatform" is not registered.
```

src/contracts/Test/IbexaKernelTestCase.php
```diff
@@ -0,0 +1,101 @@
...
+    {
+        return IbexaTestKernel::class;
+    }
+
+    /**
+     * @template T
+     * @phpstan-param class-string<T> $className
+     * @phpstan-return T
+     */
+    protected static function getServiceByClassName(string $className, ?string $id = null)
```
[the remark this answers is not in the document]
Steveb-p: Fixed in 942076ca.

Note that this also required a change in phpstan annotation (phpstan is not added to ezplatform-kernel yet, but for the future compatibility annotation is added).

```diff + php
    /**
-     * @template T of object
+     * @template T of object
     * @phpstan-param class-string<T> $className
     * @phpstan-return T
     */
-    protected static function getServiceByClassName(string $className, ?string $id = null)
+    protected static function getServiceByClassName(string $className, ?string $id = null): object
```

src/contracts/Test/IbexaKernelTestCase.php
```diff
@@ -0,0 +1,101 @@
...
+    {
+        $serviceId = IbexaTestKernel::SERVICE_ID_TO_TEST_SERVICE_MAP[$id ?? $className] ?? $id;
+        assert(is_string($serviceId));
+
+        $service = self::$container->get($serviceId);
```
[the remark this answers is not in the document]
Steveb-p: Fixed in 2ec6106f.

src/contracts/Test/IbexaTestKernel.php
```diff
@@ -0,0 +1,184 @@
...
+            new FrameworkBundle(),
+            new LiipImagineBundle(),
+            new TwigBundle(),
+            new DoctrineBundle(),
+        ];
+    }
+
+    public function registerContainerConfiguration(LoaderInterface $loader): void
+    {
+        $loader->load(static function (ContainerBuilder $container): void {
```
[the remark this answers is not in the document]
Steveb-p: Fixed in a956623f.

src/contracts/Test/IbexaTestKernel.php
Steveb-p: Which is possible, just `driver` would need to be changed into an `env(...)` declaration (similarly in PC, we have legacy/solr/ES tests done like that).

For simplicity I left it as is.

src/contracts/Test/IbexaTestKernel.php
```diff
@@ -0,0 +1,184 @@
...
+        $container->setDefinition('test.ibexa.migrations.fixture_importer', $definition);
+
+        $definition = new Definition(SqliteDbPlatform::class);
+        $definition->addMethodCall('setEventManager', [
+            new Reference('doctrine.dbal.default_connection.event_manager'),
+        ]);
+        $container->setDefinition(SqliteDbPlatform::class, $definition);
+    }
+
+    private static function prepareEzPlatformFramework(ContainerBuilder $container): void
```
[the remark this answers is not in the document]
Steveb-p: Fixed in 086ccb23.

src/contracts/Test/IbexaTestKernel.php
Steveb-p: Fixed in 3664c1ae.

composer.json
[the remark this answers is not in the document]
Steveb-p: Using `ezpublish` solves the issue.

src/contracts/Test/IbexaTestKernel.php
Steveb-p: Replaced.

src/contracts/Test/IbexaTestKernel.php
Steveb-p: Replaced with YAML files.

tests/integration/BasicKernelTest.php
```diff
@@ -0,0 +1,24 @@
+<?php
+declare(strict_types=1);
```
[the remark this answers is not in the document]
Steveb-p: Fixed in 84e916c9.

This also adds new directories to be checked by PHP CS Fixer.

alongosz (changes requested):
In general this needs to be split into 2 or 3 parts:
1. Namespaces should be introduced to ezpublish-kernel:7.5 since EOM is March 2022, so there's 100% chance that we will need it there anyway
2. Fixer changes (maybe with what I mentioned in the diff comment that follows) need to be separated
3. Repository Test API changes need to be separated ([IBX-764](https://issues.ibexa.co/browse/IBX-764))

2 & 3 can remain in this PR but as separate commits. In this case mark the PR as `[FAST FORWARD MERGE]` and add relevant label.

.php_cs
```diff
@@ -3,7 +3,11 @@
 return EzSystems\EzPlatformCodeStyle\PhpCsFixer\EzPlatformInternalConfigFactory::build()
     ->setFinder(
         PhpCsFixer\Finder::create()
-            ->in([__DIR__ . '/eZ'])
+            ->in([
+                __DIR__ . '/eZ',
+                __DIR__ . '/src',
+                __DIR__ . '/tests',
+            ])
```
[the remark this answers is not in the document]
Steveb-p: As discussed I've reverted the change to code style command and will remove all deprecations in a separate PR.

However, adding those directories is necessary for repository integrity, so it should be either added here or in a separate PR adding namespaces (which is the most probable outcome, as we expect those namespaces in 2.5 version of the platform as well).

Steveb-p:
@alongosz extracted PHP namespace changes to https://github.com/ezsystems/ezpublish-kernel/pull/3111, since they should be merged up from ezplatform `2.5`.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,111 @@
...
+            -   uses: "ramsey/composer-install@v1"
+                with:
+                    dependency-versions: "highest"
+                    composer-options: "${{ matrix.composer_options }}"
+
+            -   name: Setup problem matchers for PHPUnit
+                run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
+
+            -   name: Run integration test suite vs Postgresql
+                run: composer exec phpunit tests
```
alongosz, line 109: Can we already have https://github.com/ibexa/core/pull/3/commits/02cde5fbea8180f84aeaba6129b1659e1301ae34 and
```suggestion
                run: composer run-script test
```

Side note: this is not a correct way to execute integration tests. Those are split into test suites which are configured via different `phpunit*.xml` files.

phpunit.xml
```diff
@@ -11,6 +11,7 @@
   >
   <php>
     <ini name="error_reporting" value="-1" />
+    <env name="DATABASE_URL" value="sqlite:///:memory:" />
```
alongosz, line 14: Unit tests should not rely on any database. phpunit.xml is dedicated to unit tests only

alongosz (changes requested):

phpunit.xml
Steveb-p: Fixed in fdfec9f0.

.github/workflows/ci.yaml
Steveb-p: Fixed in fdfec9f0.

Steveb-p: The initial idea was to keep those tests separate, because currently Travis does all the other tests.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,106 @@
...
+on:
+    push:
+        branches:
+            - master
+            - '[0-9]+.[0-9]+'
+    pull_request: ~
+
+jobs:
+    tests:
+        name: Tests
```
alongosz:
```suggestion
    tests:
        name: Unit tests & SQLite integration tests
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,106 @@
...
+            - name: Run code style check
+              run: composer run-script check-cs -- --format=checkstyle | cs2pr
+
+            - name: Run unit test suite
+              run: composer run-script unit
+
+            - name: Run integration test suite
+              run: composer run-script integration
+
+    postgres-tests:
```
alongosz:
```suggestion
    integration-tests-postgres:
        name: PostgreSQL integration tests
```

alongosz (changes requested):

alongosz (approved):

src/contracts/Test/IbexaTestKernel.php
```diff
@@ -0,0 +1,207 @@
...
+ *
+ * IbexaTestKernel::getExposedServicesByClass() is a simpler variant provided for services that are registered in
+ * service container using their FQCN.
+ *
+ * IbexaTestKernel::getExposedServicesById() is useful if your service is not registered as it's FQCN (for example,
+ * if you have multiple services for the same class / interface).
+ *
+ * If don't need the repository services (or not all), you can replace the IbexaTestKernel::EXPOSED_SERVICES_BY_CLASS and
+ * IbexaTestKernel::EXPOSED_SERVICES_BY_ID consts in extending class, without changing the methods above.
+ */
```
adamwojs, line 65: Please move it to documentation if it's meant to be used in public or into our internal wiki.

src/contracts/Test/IbexaKernelTestCase.php
```diff
@@ -0,0 +1,178 @@
...
+                'Expected %s to be an instance of %s.',
+                get_class($kernel),
+                IbexaTestKernel::class,
+            ));
+        }
+
+        $id = $id ?? $className;
+
+        return $kernel->getAliasServiceId($id);
+    }
```
Steveb-p, line 117: @adamwojs @alongosz `getServiceByClassName()` method has been changed to use the actual Kernel instance, and allow changing how the service ID is acquired by descending classes.

This will make it possible to use this class without `IbexaTestKernel`, and in cases where `createKernel` method is overriden entirely (which is normally the method that calls `getKernelClass()`, and may provide a different way of supplying a Kernel instance).

src/contracts/Test/IbexaTestKernel.php
Steveb-p: I've discussed this a bit with @alongosz and I think we should mark this class as `@experimental` (since it's possible something will change) and the phpdoc description should be copied into internal wiki.

I'd rather keep the class usage description in the class and consider it the source of truth for any documentation that follows, be it internal or official. WDYT?

Steveb-p: @adamwojs documentation added to internal wiki.

adamwojs (approved):

Steveb-p:
Merged into `master` in 57c50999.

PR #221, opened 2021-08-02
IBX-843: Skipped unnecessary relation building during content update

barw4:
| **JIRA issue**                          | [IBX-843](https://issues.ibexa.co/browse/IBX-843)
| **BC breaks**                          | ?

This PR is explained in detail in the JIRA ticket and is rather proof of concept. The discussion here is advisable.

#### Checklist:

### DIFF
--- a/eZ/Publish/Core/Repository/ContentService.php
+++ b/eZ/Publish/Core/Repository/ContentService.php
@@ -1338,13 +1338,16 @@ class ContentService implements ContentServiceInterface
                     $isEmpty = true;
                 }
 
-                $this->relationProcessor->appendFieldRelations(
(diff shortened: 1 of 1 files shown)


adamwojs (changes requested):
Proposed changes breaks relation extraction. Please look on the following scenario:

1. Create content A, B, C
2. Embed B in A and publish
3. Add translation to A , embed C instead of B and publish

Result: 

B is no longer considered as relation to A. This is an issue because B is still embedded in A (but in different translation). What you can do here is to try copy relations from untouched translations. 

Also please keep in mind that PAPI is designed to be work on multiple translations at once even if GUI doesn't offer such capabilities.

PR #231, opened 2021-09-06
IBX-947: Fixed deleteVersion method to remove Content if necessary

barw4:
| **JIRA issue**                          | [IBX-947](https://issues.ibexa.co/browse/IBX-947)

There is a problem with providing a test for such a case. Actually, [this particular test](https://github.com/ezsystems/ezplatform-kernel/blob/master/eZ/Publish/API/Repository/Tests/ContentServiceTest.php#L3125) should handle this problem. However, it doesn't check for records to exist in `ezcontentobject` table which may be orphaned. We don't have an API method to load a single record from `ezcontentobject` table. The `loadContent` method loads content using the join and that's why it throws the `NotFoundException`, however, this record still exists and causes problems elsewhere.

ping @ezsystems/php-dev-team 

#### Checklist:

### DIFF
--- a/eZ/Publish/Core/Repository/ContentService.php
+++ b/eZ/Publish/Core/Repository/ContentService.php
@@ -1717,8 +1717,9 @@ class ContentService implements ContentServiceInterface
             null,
             2
         );
+        $versionsCount = count($versionList);
 
-        if (count($versionList) === 1 && !$versionInfo->isDraft()) {
+        if ($versionsCount === 1 && !$versionInfo->isDraft()) {
             throw new BadStateException(
                 '$versionInfo',
                 'The Version is the last version of the Content item and cannot be removed'
@@ -1727,10 +1728,12 @@ class ContentService implements ContentServiceInterface
 
         $this->repository->beginTransaction();
         try {
-            $this->persistenceHandler->contentHandler()->deleteVersion(
+            $versionsCount === 1 && $versionInfo->isDraft()
+            ? $this->persistenceHandler->contentHandler()->deleteContent($versionInfo->contentInfo->id)
+            : $this->persistenceHandler->contentHandler()->deleteVersion(
                 $versionInfo->getContentInfo()->id,
                 $versionInfo->versionNo
-            );
+              );
             $this->repository->commit();
         } catch (Exception $e) {
             $this->repository->rollback();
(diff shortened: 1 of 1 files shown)


alongosz (changes requested):
> There is a problem with providing a test for such a case. Actually, [this particular test](https://github.com/ezsystems/ezplatform-kernel/blob/master/eZ/Publish/API/Repository/Tests/ContentServiceTest.php#L3125) should handle this problem. However, it doesn't check for records to exist in `ezcontentobject` table which may be orphaned. We don't have an API method to load a single record from `ezcontentobject` table. The `loadContent` method loads content using the join and that's why it throws the `NotFoundException`, however, this record still exists and causes problems elsewhere.

Uh... Kernel... :upside_down_face: 
What is gonna happen if instead of loading Content you'd try to load `ContentInfo`?

barw4:
I don't think it will help, we are joining there an `ezobjectcontent_tree` table in which record doesn't exist at that point.

alongosz:
@barw4 but did you check that specific test case?

barw4:
@alongosz yes, the test passes as well.

alongosz:
@barw4 ok, I'll try to check it today, because locally I got different results, but done with haste.

alongosz (changes requested):
@barw4 for me this simple change on the stable `1.3` branch:
```diff
diff --git a/eZ/Publish/API/Repository/Tests/ContentServiceTest.php b/eZ/Publish/API/Repository/Tests/ContentServiceTest.php
index d0d609cc0..f638f5bc1 100644
--- a/eZ/Publish/API/Repository/Tests/ContentServiceTest.php
+++ b/eZ/Publish/API/Repository/Tests/ContentServiceTest.php
@@ -3132,7 +3132,7 @@ class ContentServiceTest extends BaseContentServiceTest
 
         // This call will fail with a "NotFound", because we allow to delete content if remaining version is draft.
         // Can normally only happen if there where always only a draft to begin with, simplifies UI edit API usage.
-        $this->contentService->loadContent($draft->id);
+        $this->contentService->loadContentInfo($draft->id);
     }
 
     /**
```
causes the test to fail. After applying your fix, the test starts to pass. Note that you have to disable `@depends` annotations for the test to be not skipped.

Question 2: It seems to me this applies to v2.5, which is also supported, so all the changes should be applied on `ezpublish-kernel:7.5`

barw4:
Closed in favor of https://github.com/ezsystems/ezpublish-kernel/pull/3118

PR #245, opened 2021-10-14
IBX-1258: Fixed multiple `FixtureImporter::import()` calls not resetting sequences

Steveb-p:
| **JIRA issue**                          | [IBX-1258](https://issues.ibexa.co/browse/IBX-1258)

This changes the internal cache for `FixtureImporter`, and how it is regenerated when new fixtures are imported.

Previously, calling `FixtureImporter::import()` multiple times (with different `Fixture`s) was possible, but sequence keys for `Fixture`s after the first one were not regenerated properly. This results in issues when different tests receive different fixtures, usually concluding in primary key constraint validations.

#### Checklist:
- [x] ~Tested the solution manually.~


Steveb-p:
@alongosz codebase has been changed to include the fact that some tables might not have an autoincrement column - which in previous code would cause them to be constantly queried.

This causes the `FixtureImporter::$resetSequenceStatements` property to change it's declaration to `array<string, string|null>`, where `null` value indicates a table without autoincrement.

PR #264, opened 2021-11-23
IBX-1489: Provided integration test for Installers for PHP8

Steveb-p:
| **JIRA issue**                          | [IBX-1489](https://issues.ibexa.co/browse/IBX-1489)

Provides integration tests & fixes for PHP8 compatibility.

# Notable changes:

## Disabling transactions during schema import / install.

Due to a change in PHP's MySQL driver https://github.com/php/php-src/commit/990bb34 for PHP8 it becomes necessary to handle a case where `Connection::commit()` throws an exception that there is no transaction (which happens after DDL SQL commands).

> DDL statements on mysql result in an implicit commit (see [mysql documentation](https://dev.mysql.com/doc/refman/8.0/en/implicit-commit.html)). This causes the explicit commit in transactional mode to fail. But only starting with php 8.0 an error is raised (probably due to php/php-src@990bb34).

## Removal of PHP8 deprecation warnings

This includes:
 * Reordering of some test methods, so that optional argument are at the end of the list.
 * Converting values to `string`, where other data types are possible. Especially `null` or `int`.

## Issues that will require addressing in the near future
 * One test fails specifically on PHP 8.1 due to change in `DateTime` interval behavior. It is therefore disabled on this platform.
 * MySQL 8.0 tests have issues with collation conversion (which are **probably** silently ignored on earlier versions).

#### Checklist:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -138,9 +138,9 @@ jobs:
         timeout-minutes: 20
 
         strategy:
+            fail-fast: false
             matrix:
                 php:
-                    - '7.3'
                     - '7.4'
                 composer_options: [ "" ]
                 include:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -58,7 +58,7 @@ jobs:
 
     integration-tests-postgres:
         name: PostgreSQL integration tests
-        needs: tests
+        # needs: tests
         services:
             postgres:
                 image: postgres:10
@@ -117,7 +117,7 @@ jobs:
 
     integration-tests-mysql:
         name: MySQL integration tests
-        needs: tests
+        # needs: tests
         services:
             mysql:
                 image: mysql:8.0
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -135,7 +135,7 @@ jobs:
                     --health-retries=5
                     --tmpfs=/var/lib/mysql
         runs-on: "ubuntu-20.04"
-        timeout-minutes: 20
+        timeout-minutes: 60
 
         strategy:
             fail-fast: false
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -80,12 +80,10 @@ jobs:
                     --tmpfs /var/lib/postgres
         runs-on: "ubuntu-20.04"
         timeout-minutes: 60
-        continue-on-error: ${{ matrix.experimental }}
 
         strategy:
             fail-fast: false
             matrix:
-                experimental: [ false ]
                 php:
                     - '7.3'
                     - '7.4'
@@ -127,7 +125,7 @@ jobs:
         # needs: tests
         services:
             mysql:
-                image: mysql:8.0
+                image: mysql:5.7
                 ports:
                     - 3306/tcp
                 env:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,6 +21,9 @@ jobs:
                     - '7.3'
                     - '7.4'
                 composer_options: [ "" ]
+                include:
+                    -   php: '8.0'
+                        composer_options: "--ignore-platform-req php"
 
         steps:
             - uses: actions/checkout@v2
@@ -82,6 +85,9 @@ jobs:
                     - '7.3'
                     - '7.4'
                 composer_options: [ "" ]
+                include:
+                    -   php: '8.0'
+                        composer_options: "--ignore-platform-req php"
 
         steps:
             -   uses: actions/checkout@v2
@@ -108,3 +114,60 @@ jobs:
                     DATABASE_URL: "pgsql://postgres:postgres@localhost:${{ job.services.postgres.ports[5432] }}/testdb?server_version=10"
                     # Required by old repository tests
                     DATABASE: "pgsql://postgres:postgres@localhost:${{ job.services.postgres.ports[5432] }}/testdb"
+
+    integration-tests-mysql:
+        name: MySQL integration tests
+        needs: tests
+        services:
+            mysql:
+                image: mysql:8.0
+                ports:
+                    - 3306/tcp
+                env:
+                    MYSQL_RANDOM_ROOT_PASSWORD: true
+                    MYSQL_USER: mysql
+                    MYSQL_PASSWORD: mysql
+                    MYSQL_DATABASE: testdb
+                options: >-
+                    --health-cmd="mysqladmin ping"
+                    --health-interval=10s
+                    --health-timeout=5s
+                    --health-retries=5
+                    --tmpfs=/var/lib/mysql
+        runs-on: "ubuntu-20.04"
+        timeout-minutes: 20
+
+        strategy:
+            matrix:
+                php:
+                    - '7.3'
+                    - '7.4'
+                composer_options: [ "" ]
+                include:
+                    -   php: '8.0'
+                        composer_options: "--ignore-platform-req php"
+
+        steps:
+            -   uses: actions/checkout@v2
+
+            -   name: Setup PHP Action
+                uses: shivammathur/setup-php@v2
+                with:
+                    php-version: ${{ matrix.php }}
+                    coverage: none
+                    extensions: pdo_mysql, gd
+                    tools: cs2pr
+
+            -   uses: "ramsey/composer-install@v1"
+                with:
+                    dependency-versions: "highest"
+                    composer-options: "${{ matrix.composer_options }}"
+
+            -   name: Setup problem matchers for PHPUnit
+                run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
+
+            -   name: Run integration test suite vs MySQL
+                run: composer run-script integration
+                env:
+                    DATABASE_URL: "mysql://mysql:mysql@127.0.0.1:${{ job.services.mysql.ports[3306] }}/testdb"
+                    DATABASE: "mysql://mysql:mysql@127.0.0.1:${{ job.services.mysql.ports[3306] }}/testdb"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,7 +17,6 @@ jobs:
             fail-fast: false
             matrix:
                 experimental: [ false ]
-                code_style_check: [ true ]
                 php:
                     - '7.3'
                     - '7.4'
@@ -27,7 +26,7 @@ jobs:
                         composer_options: "--ignore-platform-req php"
                     -   php: '8.1'
                         composer_options: "--ignore-platform-req php"
-                        code_style_check: false
+                        skip_code_style: true
 
         steps:
             - uses: actions/checkout@v2
@@ -52,7 +51,7 @@ jobs:
             #   run: composer run-script phpstan
 
             - name: Run code style check
-              if: matrix.code_style_check
+              if: matrix.skip_code_style != true
               run: composer run-script check-cs -- --format=checkstyle | cs2pr
 
             - name: Run unit test suite
@@ -63,7 +62,7 @@ jobs:
 
     integration-tests-postgres:
         name: PostgreSQL integration tests
-        # needs: tests
+        needs: tests
         services:
             postgres:
                 image: postgres:10
@@ -122,7 +121,7 @@ jobs:
 
     integration-tests-mysql:
         name: MySQL integration tests
-        # needs: tests
+        needs: tests
         services:
             mysql:
                 image: mysql:5.7
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,6 +17,7 @@ jobs:
             fail-fast: false
             matrix:
                 experimental: [ false ]
+                code_style_check: [ true ]
                 php:
                     - '7.3'
                     - '7.4'
@@ -26,6 +27,7 @@ jobs:
                         composer_options: "--ignore-platform-req php"
                     -   php: '8.1'
                         composer_options: "--ignore-platform-req php"
+                        code_style_check: false
 
         steps:
             - uses: actions/checkout@v2
@@ -50,6 +52,7 @@ jobs:
             #   run: composer run-script phpstan
 
             - name: Run code style check
+              if: matrix.code_style_check
               run: composer run-script check-cs -- --format=checkstyle | cs2pr
 
             - name: Run unit test suite
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,6 +24,8 @@ jobs:
                 include:
                     -   php: '8.0'
                         composer_options: "--ignore-platform-req php"
+                    -   php: '8.1'
+                        composer_options: "--ignore-platform-req php"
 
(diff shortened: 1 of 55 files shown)


.github/workflows/ci.yaml
```diff
@@ -21,6 +21,9 @@ jobs:
                     - '7.3'
                     - '7.4'
                 composer_options: [ "" ]
+                include:
+                    -   php: '8.0'
```
konradoboza, line 25: We might consider switching to 8.1 since it was released today.

Steveb-p: On kernel/core we can safely add both, because open source projects have unlimited GH Action minutes.

.github/workflows/ci.yaml
```diff
@@ -17,10 +17,17 @@ jobs:
             fail-fast: false
             matrix:
                 experimental: [ false ]
+                code_style_check: [ true ]
```
Steveb-p:
```suggestion
```

.github/workflows/ci.yaml
```diff
@@ -17,10 +17,17 @@ jobs:
...
                 php:
                     - '7.3'
                     - '7.4'
                 composer_options: [ "" ]
+                include:
+                    -   php: '8.0'
+                        composer_options: "--ignore-platform-req php"
+                    -   php: '8.1'
+                        composer_options: "--ignore-platform-req php"
+                        code_style_check: false
```
Steveb-p:
```suggestion
                        skip_code_style: true
```

.github/workflows/ci.yaml
```diff
@@ -45,6 +52,7 @@ jobs:
             #   run: composer run-script phpstan
 
             - name: Run code style check
+              if: matrix.code_style_check
```
Steveb-p: This change will make it so we don't need to explicitly declare in matrix that we want code style to run. It will run automatically unless we explicitly say otherwise.
```suggestion
              if: !matrix.skip_code_style
```

eZ/Publish/Core/FieldType/Media/Type.php
```diff
@@ -246,21 +246,11 @@ public function fromPersistenceValue(FieldValue $fieldValue)
...
-            ? $fieldValue->externalData['height']
-            : 0);
-        $result->width = (isset($fieldValue->externalData['width'])
-            ? $fieldValue->externalData['width']
-            : 0);
+        $result->hasController = ($fieldValue->externalData['hasController'] ?? false);
+        $result->autoplay = ($fieldValue->externalData['autoplay'] ?? false);
+        $result->loop = ($fieldValue->externalData['loop'] ?? false);
+        $result->height = ($fieldValue->externalData['height'] ?? 0);
+        $result->width = ($fieldValue->externalData['width'] ?? 0);
```
Steveb-p:
```suggestion
        $result->hasController = $fieldValue->externalData['hasController'] ?? false;
        $result->autoplay = $fieldValue->externalData['autoplay'] ?? false;
        $result->loop = $fieldValue->externalData['loop'] ?? false;
        $result->height = $fieldValue->externalData['height'] ?? 0;
        $result->width = $fieldValue->externalData['width'] ?? 0;
```

.github/workflows/ci.yaml
```diff
@@ -55,7 +63,7 @@ jobs:
 
     integration-tests-postgres:
         name: PostgreSQL integration tests
-        needs: tests
+        # needs: tests
```
Steveb-p: We should remember to revert this one (although technically it's OSS, so we could run integration tests against databases immediately - github pays :dollar: )

eZ/Publish/Core/MVC/Symfony/FieldType/Tests/User/ParameterProviderTest.php
```diff
@@ -44,6 +44,9 @@ protected function setUp(): void
         $this->parameterProvider = new ParameterProvider($this->userService);
     }
 
+    /**
+     * @requires PHP < 8.1
```
konradoboza, line 48: The reason for that was most likely some regression coming from PHP 8.1, ref: https://github.com/ezsystems/ezplatform-kernel/runs/4354252098?check_suite_focus=true#step:7:156.

.github/workflows/ci.yaml
```diff
@@ -72,16 +79,19 @@ jobs:
...
             fail-fast: false
             matrix:
-                experimental: [ false ]
                 php:
                     - '7.3'
                     - '7.4'
                 composer_options: [ "" ]
+                include:
+                    -   php: '8.0'
+                        composer_options: "--ignore-platform-req php"
```
alongosz, line 92: If you have:
```json
"php": "^7.3 || ^8.0",
```
Do you still need `--ignore-platform-req php`?

konradoboza:
The failing test case seems to be random, given it passes on all other PHP versions. Perhaps tests order comes into play.

eZ/Bundle/PlatformInstallerBundle/src/Installer/CoreInstaller.php
```diff
@@ -67,16 +66,9 @@ public function importSchema()
...
-                $this->db->exec($query);
-                $progressBar->advance(1);
-            }
-            $this->db->commit();
-        } catch (DBALException $e) {
-            $this->db->rollBack();
-            throw $e;
+        foreach ($queries as $query) {
+            $this->db->exec($query);
+            $progressBar->advance(1);
```
alongosz, line 71: Why do you need to get rid of this? If there's no transaction, a transaction is per commit which means that in case of any failure, partial installation will be persisted.

eZ/Publish/Core/Persistence/Tests/Legacy/FieldValue/Converter/ImageConverterTest.php
```diff
@@ -48,7 +48,7 @@ public function testToStorageValue(FieldValue $fieldValue, string $expectedXml):
         ClockMock::withClockMock(true);
 
         $time = ClockMock::time();
-        $expectedXml = str_replace('{timestampToReplace}', $time, $expectedXml);
+        $expectedXml = str_replace('{timestampToReplace}', (string) $time, $expectedXml);
```
alongosz: inconsistent CS, so far you had no space between a cast and a variable.

eZ/Bundle/EzPublishCoreBundle/Imagine/AliasGenerator.php
```diff
@@ -94,7 +94,7 @@ public function getVariation(Field $field, VersionInfo $versionInfo, $variationN
             try {
                 $originalBinary = $this->dataLoader->find($originalPath);
             } catch (NotLoadableException $e) {
-                throw new SourceImageNotFoundException($originalPath, 0, $e);
+                throw new SourceImageNotFoundException((string)$originalPath, 0, $e);
```
alongosz, line 97: Why is this necessary and how it's related to PHP 8, since `SourceImageNotFoundException` is our own class?

eZ/Publish/Core/FieldType/Tests/AuthorTest.php
```diff
@@ -561,8 +561,8 @@ public function provideDataForGetName(): array
         );
 
         return [
-            [$this->getEmptyValueExpectation(), [], 'en_GB', ''],
-            [$authorList, [], 'en_GB', 'Boba Fett'],
+            [$this->getEmptyValueExpectation(), '', [], 'en_GB'],
+            [$authorList, 'Boba Fett', [], 'en_GB'],
```
alongosz, line 565: Why do we change the order of arguments generated by the provider and how is it related to PHP 8?

alongosz (changes requested):
It's a bit unclear to me why some changes were made and how they're related to what the PR is supposed to be fixing. Unfortunately the PR description is rather empty.

.github/workflows/ci.yaml
Steveb-p: Other dependencies prevent us from not using this option.

```
composer update
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires php ^7.3 but your php version (8.0.11) does not satisfy that requirement.
  Problem 2
    - Root composer.json requires ibexa/ci-scripts ^0.1@dev -> satisfiable by ibexa/ci-scripts[0.1.x-dev].
    - ibexa/ci-scripts 0.1.x-dev requires php ^7.3 -> your php version (8.0.11) does not satisfy that requirement.
  Problem 3
    - ezsystems/doctrine-dbal-schema[v1.0.0-beta1, ..., 1.0.x-dev] require php ^7.3 -> your php version (8.0.11) does not satisfy that requirement.
    - Root composer.json requires ezsystems/doctrine-dbal-schema ^1.0@dev -> satisfiable by ezsystems/doctrine-dbal-schema[v1.0.0-beta1, ..., 1.0.x-dev].
```

eZ/Bundle/PlatformInstallerBundle/src/Installer/CoreInstaller.php
Steveb-p: MySQL driver received a change in PHP 8.0 that makes attempting to commit a transaction when there isn't any an error. Because DDL operations do not allow transactions in MySQL this means that transaction is implicitly committed. This results in an error.

A similar solution was made for doctrine/migrations, where it was suggested that mysql Migrations should explicitly be made outside of transaction. See https://github.com/doctrine/migrations/issues/1104.

Note that technically only Postgres supports transactions in DDL operations, which means that in reality on MySQL those were in fact partial installations.

Another option is acquiring internal `PDO` object from Doctrine DBAL's `Connection` (via `getWrappedConnection()`) and using `inTransaction()` method which asks the database if there is a transaction in progress (which is also one of the suggested solutions to the issue, and which is cached for subsequent calls if there were no SQL commands).

eZ/Bundle/EzPublishCoreBundle/Imagine/AliasGenerator.php
Steveb-p: I'm almost certain that this intends to fix an issue when `null` value is in `$originalPath`. Possibly some tests do not configure a mock I guess?

eZ/Bundle/PlatformInstallerBundle/src/Installer/CoreInstaller.php
konradoboza: As of PHP 8, `PDO::commit()` and `rollBack()` will throw a `There is no active transaction` PDOException if the transaction had been autocommitted, Previously they would just fail silently.

eZ/Bundle/EzPublishCoreBundle/Imagine/AliasGenerator.php
konradoboza: This fix aims to suppress deprecation warnings like: `PHP Deprecated:  Exception::__construct(): Passing null to parameter #1 ($message) of type string is deprecated`.

eZ/Publish/Core/FieldType/Tests/AuthorTest.php
konradoboza: Another deprecation warning: `PHP Deprecated:  Optional parameter [xxx] declared before required parameter [yyy] is implicitly treated as a required parameter`, ref: https://www.php.net/manual/en/migration80.deprecated.php#migration80.deprecated.core.

Steveb-p: From what I see the changes are limited to test classes only. If it were otherwise, I would say see we should make the arguments mandatory instead and remove the default value (as it was impossible to use the optional parameter anyway - it always had to be specified).

That is, unless those test classes were used in other libraries, which I know might have been the case sometimes?

konradoboza: >From what I see the changes are limited to test classes only

This is the main reason why I decided to just switch arguments order. I am not sure if we should care about providing any BC for tests anyway.

eZ/Bundle/EzPublishCoreBundle/Imagine/VariationPurger/LegacyStorageImageFileList.php
```diff
@@ -96,7 +97,7 @@ private function fetchRow(): void
         ++$this->cursor;
         $imageId = $this->rowReader->getRow();
 
-        if (substr($imageId, 0, strlen($prefix)) === $prefix) {
+        if (strpos((string)$imageId, $prefix) === 0) {
```
Steveb-p: As far as I know the previous variant would be _slightly_ more efficient (but that's micro). What was the reason for change?

konradoboza: This was one of the SonarLint plugin suggestions. To me, readability benefits here, but this isn't something I would have any strong opinion on.

Steveb-p: Symfony polyfill for PHP 7.4 proposes the following implementation of `str_starts_with`:
```php
public static function str_starts_with(string $haystack, string $needle): bool
{
    return 0 === strncmp($haystack, $needle, strlen($needle));
}
```
Benefits include:
 * binary-safe comparison
 * only beginning of the string is compared, which is not the case for `strpos` (which looks up through the whole string)

.github/workflows/ci.yaml
```diff
@@ -108,3 +118,62 @@ jobs:
                     DATABASE_URL: "pgsql://postgres:postgres@localhost:${{ job.services.postgres.ports[5432] }}/testdb?server_version=10"
                     # Required by old repository tests
                     DATABASE: "pgsql://postgres:postgres@localhost:${{ job.services.postgres.ports[5432] }}/testdb"
+
+    integration-tests-mysql:
+        name: MySQL integration tests
+        needs: tests
+        services:
+            mysql:
+                image: mysql:5.7
```
Steveb-p, line 127: Note: A follow up draft PR with `mysql:8.0` is necessary to show that there are tests that will fail on MySQL 8.

eZ/Bundle/PlatformInstallerBundle/src/Installer/CoreInstaller.php
alongosz: Fair enough, can't do much about that, +1

eZ/Bundle/EzPublishCoreBundle/Imagine/AliasGenerator.php
alongosz: Thx for the explanation! +1

eZ/Publish/Core/FieldType/Tests/AuthorTest.php
alongosz: +1

alongosz (approved):
+1 but Steve has some valid remarks: https://github.com/ezsystems/ezplatform-kernel/pull/264#pullrequestreview-818744103 ;)

konradoboza (approved):

micszo (approved):
Tested with sanities on Ibexa Commerce 3.3.12-dev with PHP8.1.

QA Approved.

PR #265, opened 2021-11-23
Provided PHP8 compatibility for moving content

konradoboza:
| **JIRA issue**                          | [IBX-1521](https://issues.ibexa.co/browse/IBX-1521)
| **Target Ibexa version** | `v3.3`, `v4.0`

After moving content we end up with the following warning:
`Warning: reset(): Argument #1 ($array) must be passed by reference, value given`

`reset` function is waiting for a variable reference which isn't the case while passing it to `array_map`.

#### Checklist:

### DIFF
--- a/eZ/Publish/Core/Persistence/Legacy/Content/Location/Gateway/DoctrineDatabase.php
+++ b/eZ/Publish/Core/Persistence/Legacy/Content/Location/Gateway/DoctrineDatabase.php
@@ -231,7 +231,7 @@ final class DoctrineDatabase extends Gateway
         $results = $statement->fetchAll($onlyIds ? (FetchMode::COLUMN | PDO::FETCH_GROUP) : FetchMode::ASSOCIATIVE);
         // array_map() is used to map all elements stored as $results[$i][0] to $results[$i]
         return $onlyIds
-            ? array_map(static function ($result) {
+            ? array_map(static function (array $result) {
                 return $result[0];
             }, $results)
             : $results;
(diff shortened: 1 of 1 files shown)


Steveb-p (approved):
Even if it does not solve the main issues reported by phpunit on PHP 8, it's still ok for fixing the warning that comes from PHP.

eZ/Publish/Core/Persistence/Legacy/Content/Location/Gateway/DoctrineDatabase.php
```diff
@@ -229,8 +229,12 @@ public function getSubtreeContent(int $sourceId, bool $onlyIds = false): array
         $statement = $query->execute();
 
         $results = $statement->fetchAll($onlyIds ? (FetchMode::COLUMN | PDO::FETCH_GROUP) : FetchMode::ASSOCIATIVE);
-        // array_map() is used to to map all elements stored as $results[$i][0] to $results[$i]
-        return $onlyIds ? array_map('reset', $results) : $results;
+        // array_map() is used to map all elements stored as $results[$i][0] to $results[$i]
+        return $onlyIds
+            ? array_map(static function ($result) {
```
adamwojs:
```suggestion
            ? array_map(static function (array $result): int {
```

but please double check return type declaration

adamwojs (approved):

eZ/Publish/Core/Persistence/Legacy/Content/Location/Gateway/DoctrineDatabase.php
Steveb-p: Return type depends on PDO variant. We should not assume `int` here.

As for the argument, :+1:

konradoboza: Alternatively, we might add return type and casting to `int`. Your call @adamwojs. ;)

konradoboza: After internal sync, we went only for specifying parameter type.

PR #268, opened 2021-11-30
Added PHPStan analysis

Steveb-p:

## NOTE: Temporarily set `fix-php8` as base, since it already includes fixes for some errors detected by PHPStan.

This PR introduces static analysis on the lowest possible level, with multiple low-hanging errors put into baseline to be fixed in the near future.

As a noteworthy mention, since 0.12 PHPStan version some best practices changed. Now PHPStan loads either `phpstan.neon` or `phpstan.neon.dist` configuration files (in that order) when performing `analyze` (see https://phpstan.org/config-reference#config-file).
Since this allows user-customization, this practice is followed, and configuration file option is no longer used in composer script declaration.

#### Checklist:
- [x] ~Provided automated test coverage.~ _No behavior changes_

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,0 +1,537 @@
+parameters:
+	ignoreErrors:
+		-
+			message: "#^Access to undefined constant static\\(eZ\\\\Bundle\\\\EzPublishCoreBundle\\\\DependencyInjection\\\\Compiler\\\\ViewManagerPass\\)\\:\\:VIEW_PROVIDER_IDENTIFIER\\.$#"
+			count: 2
+			path: eZ/Bundle/EzPublishCoreBundle/DependencyInjection/Compiler/ViewManagerPass.php
+
+		-
+			message: "#^Access to undefined constant static\\(eZ\\\\Bundle\\\\EzPublishCoreBundle\\\\DependencyInjection\\\\Compiler\\\\ViewManagerPass\\)\\:\\:VIEW_TYPE\\.$#"
+			count: 1
+			path: eZ/Bundle/EzPublishCoreBundle/DependencyInjection/Compiler/ViewManagerPass.php
+
+		-
+			message: "#^Method eZ\\\\Bundle\\\\EzPublishCoreBundle\\\\DependencyInjection\\\\Configuration\\\\ConfigResolver\\:\\:logTooEarlyLoadedListIfNeeded\\(\\) should return string but return statement is missing\\.$#"
+			count: 1
+			path: eZ/Bundle/EzPublishCoreBundle/DependencyInjection/Configuration/ConfigResolver.php
+
+		-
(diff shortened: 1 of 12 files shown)


konradoboza (approved):

phpstan-baseline.neon
```diff
@@ -0,0 +1,532 @@
...
+		-
+			message: "#^Class eZ\\\\Publish\\\\SPI\\\\Persistence\\\\Content\\\\UrlAlias referenced with incorrect case\\: eZ\\\\Publish\\\\SPI\\\\Persistence\\\\Content\\\\URLAlias\\.$#"
+			count: 2
+			path: eZ/Publish/Core/Repository/URLAliasService.php
+
+		-
+			message: "#^Class Symfony\\\\Component\\\\Validator\\\\ConstraintViolation constructor invoked with 1 parameter, 6\\-10 required\\.$#"
+			count: 1
+			path: eZ/Publish/SPI/FieldType/Generic/Tests/GenericTest.php
+
```
alongosz, line 532: That baseline is suspiciously short :trollface:

alongosz (approved):

phpstan-baseline.neon
Steveb-p: It's the lowest level of analysis possible (`0`). Only the most glaring and obvious issues are reported.

Steveb-p: See [PHPStan rule levels](https://phpstan.org/user-guide/rule-levels) (we're using 0):

> Here’s a brief overview of what’s checked on each level. Levels are cumulative - for example running level 5 also gives you all the checks from levels 0-4.
> 
> 0. basic checks, unknown classes, unknown functions, unknown methods called on $this, wrong number of arguments passed to those methods and functions, always undefined variables
> 1. possibly undefined variables, unknown magic methods and properties on classes with __call and __get
> 2. unknown methods checked on all expressions (not just $this), validating PHPDocs
> 3. return types, types assigned to properties
> 4. basic dead code checking - always false instanceof and other type checks, dead else branches, unreachable code after return; etc.
> 5. checking types of arguments passed to methods and functions
> 6. report missing typehints
> 7. report partially wrong union types - if you call a method that only exists on some types in a union type, level 7 starts to report that; other possibly incorrect situations
> 8. report calling methods and accessing properties on nullable types
> 9. be strict about the mixed type - the only allowed operation you can do with it is to pass it to another mixed

konradoboza: I am a little anxious already by reading points 1 and 2. 😜 Nice improvement though!

Steveb-p: FYI
```
composer phpstan -- --level=1
...
[ERROR] Found 69 errors
```
```
composer phpstan -- --level=2
...
[ERROR] Found 1444 errors
```
Which is to be expected: a lot of PHPDoc's do not match their actual contents.

adamwojs (approved):

Steveb-p:
@adamwojs @alongosz @konradoboza PR has been rebased without introducing new changes.

PR #272, opened 2021-12-17
IBX-1392: Fixed listening for encoded URI in clustered setup

barw4:
| **JIRA issue**                          | [IBX-1392](https://issues.ibexa.co/browse/IBX-1392)

As the title suggests. 

#### Checklist:


eZ/Bundle/EzPublishIOBundle/EventListener/StreamFileListener.php
```diff
@@ -62,7 +62,11 @@ public function onKernelRequest(RequestEvent $event)
 
         $binaryFile = $this->ioService->loadBinaryFileByUri($uri);
         if ($binaryFile instanceof MissingBinaryFile) {
-            throw new NotFoundHttpException("Could not find 'BinaryFile' with identifier '$uri'");
+            // Try again with decoded URI
+            $binaryFile = $this->ioService->loadBinaryFileByUri(urldecode($uri));
+            if ($binaryFile instanceof MissingBinaryFile) {
+                throw new NotFoundHttpException("Could not find 'BinaryFile' with identifier '$uri'");
+            }
```
adamwojs: Q: What will happen if my file is named `obraze%20z spacja.jpg`?

adamwojs (changes requested):
"Guessing" correct file name it's an workaround, not proper bugfix. 

IMHO the source of the regression is https://github.com/ezsystems/ezpublish-kernel/pull/3126 and this patch should be reverted (on v2.5 as well).

eZ/Bundle/EzPublishIOBundle/EventListener/StreamFileListener.php
glye: The proper solution for all these troubles is to rename ALL uploaded files to random hashes. Problem solved, and it has other benefits (unguessable). But that's a bigger change. So do we want a quick solution/workaround now or not? A real fix should come later in any case.

mateuszbieniek: @adamwojs https://github.com/ezsystems/ezpublish-kernel/pull/3126 was fixing a different issue, and with or without these changes ibx-1392 is still valid.

barw4: Both @adamwojs and @mateuszbieniek are right, to some extent. Mateusz's PR fixed the following problematic filenames:
`some%25special+chars.jpg`. 
However, https://github.com/ezsystems/ezpublish-kernel/pull/3126 broke the following:
`test test.png`.

The current PR handles both these cases,  we can also go as a temporary solution with just:
`        $binaryFile = $this->ioService->loadBinaryFileByUri(urldecode($uri));` - that would mean just a single load. Obviously, this isn't ideal but handles most of the cases.

Without guessing: https://github.com/ezsystems/ezplatform-kernel/pull/272/commits/610e6f02598ce67c4e1ef49aca8a5ead1148c94e

barw4: > Q: What will happen if my file is named `obraze%20z spacja.jpg`?

@adamwojs This will not work in any of those PRs or even before those PRs were introduced.

alongosz: > > Q: What will happen if my file is named `obraze%20z spacja.jpg`?
> 
> This will not work in any of those PRs or even before those PRs were introduced.

But do we have a solution for such case?

barw4: Besides random hashes, I don't know tbh. The problem is that the browser converts the space to `%20` before a request is made so we cannot guess which `%20` is a space or which is just `%20`. 

However, this seems like an edge case, there will always be some names that won't be parsed properly like `test$%^&_ds.jpg`. This file for example breaks on the webserver level, even before reaching PHP. Until the proper solution, which probably utilizes hashes, is made there will be always some problematic filenames.

mateuszbieniek: @glye I agree but we can't apply this solution here. In most cases images are served directly from the webserver, therefore introducing hashing will either force everyone to stream files via index.php, or at least will require a change in generated URLs, which I think both are out of the question.

glye: (Continuing the renaming/streaming discussion elsewhere)

YohannsMonnier: > bigger

I think we should only sluggify it instead of changing it to a random name, for SEO perpective having name with keyword in it could be a good point.

Yohann

alongosz (approved):

YohannsMonnier:
@barw4 Do you have a date in mind for the final fix ? 

We have patched our installation, by modifying the vendor manually, but we really would like to have a real fix.

Thank you very much, 
Yohann

barw4:
> @barw4 Do you have a date in mind for the final fix ?
> 
> We have patched our installation, by modifying the vendor manually, but we really would like to have a real fix.
> 
> Thank you very much, Yohann

Hi Yohann, can we stick to the ticket, please? I have responded there to you.

eZ/Bundle/EzPublishIOBundle/Tests/EventListener/StreamFileListenerTest.php
```diff
@@ -97,29 +97,39 @@ public function testRespondsToIoUri()
...
+
+        $finalUri = sprintf('http://%s%s', $host, $uri);
 
         $this->ioServiceMock
-            ->expects($this->once())
+            ->expects(self::once())
             ->method('loadBinaryFileByUri')
-            ->with(sprintf('http://%s%s', $host, $uri))
-            ->will($this->returnValue($binaryFile));
+            ->with(urldecode($finalUri))
```
Steveb-p, line 122: Do not use the decoding mechanism as the implementation uses. Instead, use raw value, so we can be certain about inputs and outputs, and that we do not encode / decode twice by mistake.

Steveb-p (changes requested):

barw4:
Closed as another solution was chosen.

PR #273, opened 2021-12-20
IBX-1549: changed validation content to only one language

mateuszdebinski:
| **JIRA issue**                          | [IBX-1549](https://issues.ibexa.co/browse/IBX-1549), [IBX-1741](https://issues.ibexa.co/browse/IBX-1741) and [IBX-175](https://issues.ibexa.co/browse/IBX-175)

#### Checklist:

### DIFF
--- a/eZ/Publish/Core/Repository/Validator/VersionValidator.php
+++ b/eZ/Publish/Core/Repository/Validator/VersionValidator.php
@@ -68,7 +68,7 @@ final class VersionValidator implements ContentValidator
             );
 
             foreach ($languageCodes as $languageCode) {
-                $fieldValue = $content->getField($fieldDefinition->identifier)->value ?? $fieldDefinition->defaultValue;
+                $fieldValue = $content->getField($fieldDefinition->identifier, $languageCode)->value ?? $fieldDefinition->defaultValue;
                 $fieldValue = $fieldType->acceptValue($fieldValue);
 
                 if ($fieldType->isEmptyValue($fieldValue)) {
--- a/eZ/Publish/Core/Repository/Validator/VersionValidator.php
+++ b/eZ/Publish/Core/Repository/Validator/VersionValidator.php
@@ -54,7 +54,7 @@ final class VersionValidator implements ContentValidator
         $content = $context['content'];
 
         $contentType = $content->getContentType();
-        $languageCodes = $content->versionInfo->languageCodes;
+        $languageCode = $content->versionInfo->initialLanguageCode;
 
         $allFieldErrors = [];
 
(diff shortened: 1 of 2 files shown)


Steveb-p (approved):

alongosz (changes requested):
This change contains potential regression, which is hidden by `ContentUpdateStructValidator` and `ContentCreateStructValidator` because the validation happens also there and there's no easy way to manipulate content after `updateContent($draft)` but before `publishVersion($draft->versionInfo)`.

In other words, skipping validation here makes the Validator incoherent, looking from its independent and self-contained perspective. The business logic of API allows modifying multiple translations in a single update, so all translations should be valid. That said, I didn't manage to find a use case which would clearly uncover the bug introduced, because of the existence of other validators.

What I could agree on is for this Validator to validate only modified languages, determined by the `$translations` parameter of `\eZ\Publish\Core\Repository\ContentService::publishVersion`. If publish action occurred for all translations, it should behave as prior the change in this PR. Otherwise it should validate modified language(s) passed by `$translation`.

alongosz (approved):

Steveb-p (approved):

eZ/Publish/Core/Repository/Validator/VersionValidator.php
```diff
@@ -68,7 +68,7 @@ public function validate(
             );
 
             foreach ($languageCodes as $languageCode) {
-                $fieldValue = $content->getField($fieldDefinition->identifier)->value ?? $fieldDefinition->defaultValue;
+                $fieldValue = $content->getField($fieldDefinition->identifier, $languageCode)->value ?? $fieldDefinition->defaultValue;
```
konradoboza, line 71: Unfortunately, this change would introduce a regression, ref: https://github.com/ezsystems/ezplatform-kernel/pull/246.

konradoboza (changes requested):

eZ/Publish/Core/Repository/Validator/VersionValidator.php
konradoboza: Well, to be more precise: if the above steps don't lead to the issue anymore concerning the scope of this PR, we should be good to go.

mateuszdebinski: I tested this ticket with steps from your comment, and the problem does not occur anymore

mateuszdebinski:
@bogusez new ticket to test (with steps from comment): https://issues.ibexa.co/browse/IBX-175

konradoboza (approved):

alongosz (approved):

bogusez:
> @bogusez new ticket to test (with steps from comment): https://issues.ibexa.co/browse/IBX-175

Already tested the scenario using changes from this PR and no exception occurs.

bogusez (approved):

PR #275, opened 2021-12-29
IBX-1809: Allowed override of test Kernel via environment

Steveb-p:
| **JIRA issue**                          | [IBX-1809](https://issues.ibexa.co/browse/IBX-1809)

This PR allows developers to change the base Kernel class used in their tests without adding the `getKernelClass()` method.
In practice this allows exposition & re-use of any abstract test classes without creating intermediary classes just to overwrite the Kernel class, or overwriting the method in all descendants.

Additionally, methods to switch user context are added, because they repeat in every package that needs to take user into account.

#### Checklist:
- ~Provided automated test coverage.~

### DIFF
--- a/src/contracts/Test/IbexaKernelTestCase.php
+++ b/src/contracts/Test/IbexaKernelTestCase.php
@@ -41,13 +41,6 @@ abstract class IbexaKernelTestCase extends KernelTestCase
         try {
             return parent::getKernelClass();
         } catch (LogicException $e) {
-            @trigger_error(sprintf(
-                'You should set the KERNEL_CLASS environment variable to the fully-qualified class name of your '
-                . 'Kernel in phpunit.xml / phpunit.xml.dist or override the "%1$s::createKernel()" or '
-                . '"%1$s::getKernelClass()" method.',
-                static::class
-            ), E_USER_DEPRECATED);
-
             return IbexaTestKernel::class;
         }
     }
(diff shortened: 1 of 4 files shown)


src/contracts/Test/IbexaKernelTestCase.php
```diff
@@ -31,9 +33,23 @@
...
     {
-        return IbexaTestKernel::class;
+        try {
+            return parent::getKernelClass();
+        } catch (LogicException $e) {
+            @trigger_error(sprintf(
+                'You should set the KERNEL_CLASS environment variable to the fully-qualified class name of your '
+                . 'Kernel in phpunit.xml / phpunit.xml.dist or override the "%1$s::createKernel()" or '
+                . '"%1$s::getKernelClass()" method.',
+                static::class
```
alongosz: Why not simply `__METHOD__`?

alongosz (changes requested):

src/contracts/Test/IbexaKernelTestCase.php
Steveb-p: If you look on the message you'll see that it's not always the `__METHOD__` that is suggested. To be completely frank I've copied a similar message in the parent `KernelTestCase`.

adamwojs (approved):

alongosz (approved):

ViniTou (approved):

Steveb-p:
Merged up into `ibexa/core` in https://github.com/ibexa/core/commit/47c681616bc788a6136645770f68aaa1dd40a002

PR #283, opened 2022-02-03
IBX-2041: Made LocalAdapter lazy loaded to fix commands having wrong adapter set

barw4:
| **JIRA issue**                          | [IBX-2041](https://issues.ibexa.co/browse/IBX-2041)

As the title states. Without this change, all of the console commands that use different `--siteaccess` option other than `default` have wrongly set Site Access because the `ConsoleListener` which changes Site Access is executed **after** the services are set.

Not sure if the solution is proper but it fixes the issue with all of the commands.

#### Checklist:

### DIFF
--- a/eZ/Bundle/EzPublishIOBundle/Resources/config/io.yml
+++ b/eZ/Bundle/EzPublishIOBundle/Resources/config/io.yml
@@ -99,6 +99,7 @@ services:
     # Inject the siteaccess config into a few io services
     ezpublish.core.io.flysystem.default_adapter:
         class: eZ\Publish\Core\IO\Adapter\LocalAdapter
+        lazy: true
         arguments:
             - '@eZ\Publish\Core\IO\IOConfigProvider'
             - '@ezpublish.config.resolver'
--- a/eZ/Publish/Core/IO/Adapter/LocalAdapter.php
+++ b/eZ/Publish/Core/IO/Adapter/LocalAdapter.php
@@ -18,7 +18,7 @@ use LogicException;
 /**
  * @internal
  */
-final class LocalAdapter extends Local implements ConfigScopeChangeSubscriber
+class LocalAdapter extends Local implements ConfigScopeChangeSubscriber
 {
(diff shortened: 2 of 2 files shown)


alongosz (commented):
@barw4 can you give an example of such command? I see steps to reproduce in the ticket, but I need deeper technical analysis of what is happening. Which variable(s) and where are not properly set so the issue occurs?

barw4:
sure, you can easily reproduce this by changing some `var_dir` on certain site access and then dumping the root dir here: https://github.com/ezsystems/ezplatform-kernel/blob/1.3/eZ/Publish/Core/IO/Adapter/LocalAdapter.php#L35.

now it's just a matter of running any command with the `--siteaccess=certain_site_access` parameter in `prod` environment. You can see that `LocalAdapter` has been constructed before `ConsoleListener` changed the site access from `default` to `new_site_acccess`.

alongosz (changes requested):
> sure, you can easily reproduce this by changing some var_dir on certain site access and then dumping the root dir here: https://github.com/ezsystems/ezplatform-kernel/blob/1.3/eZ/Publish/Core/IO/Adapter/LocalAdapter.php#L35.

Okay, so there's a bug there, which just got hidden by `lazy`. What needs to be really fixed here is config resolver usage in the constructor. This is incorrect because of the Symfony Container architecture. Refactor please both the class and its parent to lazily fetch both properties, so there are not configResolver calls in the constructor.

This could be achieved by inline-ing the call, but given the batch performance it would be better to use lazy property pattern, sth like:
```php
if (!isset($this->property)) {
    $this->property = $this->configResolver->getParameter('property');
}

return $this->property;
```
You might also need to react to `\eZ\Publish\Core\MVC\Symfony\MVCEvents::CONFIG_SCOPE_CHANGE` event and update those properties in this case.

adamwojs (changes requested):
Agree with @alongosz

barw4:
Closed due to https://issues.ibexa.co/browse/IBX-2295

PR #288, opened 2022-02-14
IBX-2264: Handled missing thumbnail images

barw4:
| **JIRA issue**                          | [IBX-2264](https://issues.ibexa.co/browse/IBX-2264)

As the title states.

Required `ezplatform-rest` PR: https://github.com/ezsystems/ezplatform-rest/pull/91

#### Checklist:

The related bundles fixes will be listed here:

### DIFF
--- a/eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
+++ b/eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
@@ -100,7 +100,7 @@ class ImageThumbnailStrategy implements FieldTypeBasedThumbnailStrategy, LoggerA
         ]);
     }
 
-    private function generateContentDetailsMessage(?APIVersionInfo $versionInfo)
+    private function generateContentDetailsMessage(?APIVersionInfo $versionInfo): string
     {
         return $versionInfo !== null
             ? sprintf('Content: %d, Version No: %d', $versionInfo->getContentInfo()->id, $versionInfo->versionNo)
--- a/eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
+++ b/eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
@@ -61,34 +61,26 @@ class ImageThumbnailStrategy implements FieldTypeBasedThumbnailStrategy, LoggerA
                 $this->variationName
             );
         } catch (SourceImageNotFoundException $e) {
-            $contentDetails = isset($versionInfo->contentInfo) && isset($versionInfo->versionNo)
-                ? sprintf(' Content: %d, Version No: %d', $versionInfo->contentInfo->id, $versionInfo->versionNo)
-                : '';
-
             $this->logger->warning(
                 sprintf(
-                    'Thumbnail source image generated for %s field and %s variation could not be found (%s).',
+                    'Thumbnail source image generated for %s field and %s variation could not be found (%s). %s',
                     $field->fieldTypeIdentifier,
                     $this->variationName,
-                    $e->getMessage()
+                    $e->getMessage(),
+                    $this->generateContentDetailsMessage($versionInfo)
                 )
-                . $contentDetails
             );
 
             return null;
         } catch (Exception $e) {
-            $contentDetails = isset($versionInfo->contentInfo) && isset($versionInfo->versionNo)
-                ? sprintf(' Content: %d, Version No: %d', $versionInfo->contentInfo->id, $versionInfo->versionNo)
-                : '';
-
             $this->logger->warning(
                 sprintf(
-                    'Thumbnail could not be generated for %s field and %s variation due to %s.',
+                    'Thumbnail could not be generated for %s field and %s variation due to %s. %s',
                     $field->fieldTypeIdentifier,
                     $this->variationName,
-                    $e->getMessage()
+                    $e->getMessage(),
+                    $this->generateContentDetailsMessage($versionInfo)
                 )
-                . $contentDetails
             );
 
             return null;
@@ -100,5 +92,12 @@ class ImageThumbnailStrategy implements FieldTypeBasedThumbnailStrategy, LoggerA
             'height' => $variation->height,
             'mimeType' => $variation->mimeType,
         ]);
+    }
+
+    private function generateContentDetailsMessage(?APIVersionInfo $versionInfo)
+    {
+        return $versionInfo !== null
+            ? sprintf(' Content: %d, Version No: %d', $versionInfo->getContentInfo()->id, $versionInfo->versionNo)
+            : '';
     }
 }
--- a/eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
+++ b/eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
@@ -97,7 +97,7 @@ class ImageThumbnailStrategy implements FieldTypeBasedThumbnailStrategy, LoggerA
     private function generateContentDetailsMessage(?APIVersionInfo $versionInfo)
     {
         return $versionInfo !== null
-            ? sprintf(' Content: %d, Version No: %d', $versionInfo->getContentInfo()->id, $versionInfo->versionNo)
+            ? sprintf('Content: %d, Version No: %d', $versionInfo->getContentInfo()->id, $versionInfo->versionNo)
             : '';
     }
 }
--- a/eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
+++ b/eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
@@ -51,7 +51,7 @@ class ImageThumbnailStrategy implements FieldTypeBasedThumbnailStrategy
                 $this->variationName
             );
         } catch (\Exception $e) {
-            return new Thumbnail();
+            return null;
         }
 
         return new Thumbnail([
--- a/eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
+++ b/eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
@@ -8,12 +8,16 @@ declare(strict_types=1);
 
 namespace eZ\Publish\Core\FieldType\Image;
 
+use Exception;
 use eZ\Publish\API\Repository\Values\Content\Field;
 use eZ\Publish\API\Repository\Values\Content\Thumbnail;
 use eZ\Publish\API\Repository\Values\Content\VersionInfo as APIVersionInfo;
+use eZ\Publish\Core\MVC\Exception\SourceImageNotFoundException;
 use eZ\Publish\Core\Repository\Values\Content\VersionInfo;
 use eZ\Publish\SPI\Repository\Strategy\ContentThumbnail\Field\FieldTypeBasedThumbnailStrategy;
 use eZ\Publish\SPI\Variation\VariationHandler;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
 
 class ImageThumbnailStrategy implements FieldTypeBasedThumbnailStrategy
 {
@@ -26,14 +30,19 @@ class ImageThumbnailStrategy implements FieldTypeBasedThumbnailStrategy
     /** @var string */
     private $variationName;
 
+    /** @var \Psr\Log\LoggerInterface */
+    private $logger;
+
     public function __construct(
         string $fieldTypeIdentifier,
         VariationHandler $variationHandler,
-        string $variationName
+        string $variationName,
+        LoggerInterface $logger = null
     ) {
         $this->fieldTypeIdentifier = $fieldTypeIdentifier;
         $this->variationHandler = $variationHandler;
         $this->variationName = $variationName;
+        $this->logger = null !== $logger ? $logger : new NullLogger();
     }
 
     public function getFieldTypeIdentifier(): string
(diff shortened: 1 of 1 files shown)


eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
```diff
@@ -43,12 +43,16 @@ public function getFieldTypeIdentifier(): string
...
-        );
+        try {
+            /** @var \eZ\Publish\SPI\Variation\Values\ImageVariation $variation */
+            $variation = $this->variationHandler->getVariation(
+                $field,
+                $versionInfo ?? new VersionInfo(),
+                $this->variationName
+            );
+        } catch (\Exception $e) {
+            return new Thumbnail();
```
adamwojs: According to Thumbnail contract, resource is guaranteed to be a string:

```
class Thumbnail extends ValueObject
{
    /**
     * Can be target URL or Base64 data (or anything else).
     *
     * @var string
     */
    protected $resource;
    
    # ...
}
    
```

eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
```diff
@@ -43,12 +43,16 @@ public function getFieldTypeIdentifier(): string
...
-            $this->variationName
-        );
+        try {
+            /** @var \eZ\Publish\SPI\Variation\Values\ImageVariation $variation */
+            $variation = $this->variationHandler->getVariation(
+                $field,
+                $versionInfo ?? new VersionInfo(),
+                $this->variationName
+            );
+        } catch (\Exception $e) {
```
adamwojs: Missing import.

adamwojs (changes requested):

eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
ViniTou: Is there also any reason why we cant return null in that case as it should be handled according to typehint?

barw4: @ViniTou well, we could use `null` as I have done originally but it crashes in the other bundles after doing so, for example here: https://github.com/ezsystems/ezplatform-rest/blob/1.3/src/lib/Server/Output/ValueObjectVisitor/Version.php#L140

as the `Thumbnail` is a proxy object at this point, not sure about other places.

ViniTou: Dont get it, we return null but we somehow ends with Proxy object? If yes I would say that needs a fix. And if our bundles are not ready to handle `null`s when this is in contract that also needs a fix.

barw4: That's exactly the case, I'll change it to `null` in this case and handle the remaining bundles.

eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
barw4: We have a `\` here. That import would be unused.

ViniTou: Its about replacing `\` with import.

barw4: What's wrong with `\Exception $e` though? We have plenty of such notations in the kernel :(

alongosz: > What's wrong with `\Exception $e` though? We have plenty of such notations in the kernel :(

It's a matter of convention. We don't perform bulk replace for the old codebase, but try to remember about those rules for the new code.

alongosz (changes requested):
Is a missing thumbnail a common occurrence?
Is the `\Exception` the most narrow one that can be handled?

I would expect to see here:
* Handling of the most narrow exception (it should come from API)
* Handling generic exception with some error logging

If the exception for missing thumbnail is not at all common, maybe it should be logged too? Only in that case handling can be unified as-is (with logging).

barw4:
> Is a missing thumbnail a common occurrence? Is the `\Exception` the most narrow one that can be handled?
> 
> I would expect to see here:
> 
> * Handling of the most narrow exception (it should come from API)
> * Handling generic exception with some error logging
> 
> If the exception for missing thumbnail is not at all common, maybe it should be logged too? Only in that case handling can be unified as-is (with logging).

@alongosz I think this can be quite common, especially on dev installations with missing image files. The most common exception would be `SourceImageNotFoundException` and this has been narrowed down and logged, besides logging for rest of the exceptions has been added. Please, see https://github.com/ezsystems/ezplatform-kernel/pull/288/commits/4903a02146ce6784d3a0abde135d31da0853b566

eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
```diff
@@ -26,14 +30,19 @@ class ImageThumbnailStrategy implements FieldTypeBasedThumbnailStrategy
...
         string $fieldTypeIdentifier,
         VariationHandler $variationHandler,
-        string $variationName
+        string $variationName,
+        LoggerInterface $logger = null
     ) {
         $this->fieldTypeIdentifier = $fieldTypeIdentifier;
         $this->variationHandler = $variationHandler;
         $this->variationName = $variationName;
+        $this->logger = null !== $logger ? $logger : new NullLogger();
```
alongosz: Without the extra configuration this logger is always gonna be a null logger.
Use `LoggerAwareTrait` and implement `LoggerAwareInterface` instead. Those combined together are [automatically configured](https://symfony.com/doc/5.4/logging.html#using-a-logger-inside-a-service) to inject an actual logger

eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
```diff
@@ -43,12 +52,34 @@ public function getFieldTypeIdentifier(): string
...
+            );
+
+            return null;
+        } catch (Exception $e) {
+            $this->logger->warning(
+                sprintf(
+                    'Thumbnail could not be generated for %s field and %s variation. Error: %s',
+                    $field->fieldTypeIdentifier,
+                    $this->variationName,
+                    $e->getMessage()
```
alongosz: The message in both cases should contain all the information needed to identify the Content item which causes the issue.
* if `$versionInfo` is not null: log Content Id + Version No
* `$field->fieldTypeIdentifier` - missing only for the first message
* `$this->variationName`  - missing only for the first message

alongosz (changes requested):

eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
```diff
@@ -29,11 +37,13 @@ class ImageThumbnailStrategy implements FieldTypeBasedThumbnailStrategy
...
         string $fieldTypeIdentifier,
         VariationHandler $variationHandler,
-        string $variationName
+        string $variationName,
+        ?LoggerInterface $logger = null
     ) {
         $this->fieldTypeIdentifier = $fieldTypeIdentifier;
         $this->variationHandler = $variationHandler;
         $this->variationName = $variationName;
+        $this->logger = $logger ?: new NullLogger();
```
alongosz: CS
```suggestion
        $this->logger = $logger ?? new NullLogger();
```

eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
```diff
@@ -43,12 +53,46 @@ public function getFieldTypeIdentifier(): string
...
+            /** @var \eZ\Publish\SPI\Variation\Values\ImageVariation $variation */
+            $variation = $this->variationHandler->getVariation(
+                $field,
+                $versionInfo ?? new VersionInfo(),
+                $this->variationName
+            );
+        } catch (SourceImageNotFoundException $e) {
+            $contentDetails = isset($versionInfo->contentInfo) && isset($versionInfo->versionNo)
+                ? sprintf(' Content: %d, Version No: %d', $versionInfo->contentInfo->id, $versionInfo->versionNo)
+                : '';
```
alongosz: It's not necessary to check properties here. Also we tend not to use magic properties if there's an alternative. What is simply enough:
```suggestion
            $contentDetails = null !== $versionInfo
                ? sprintf(' Content: %d, Version No: %d', $versionInfo->getContentInfo()->id, $versionInfo->versionNo)
                : '';
```

eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
```diff
@@ -43,12 +53,46 @@ public function getFieldTypeIdentifier(): string
...
+
+            $this->logger->warning(
+                sprintf(
+                    'Thumbnail source image generated for %s field and %s variation could not be found (%s).',
+                    $field->fieldTypeIdentifier,
+                    $this->variationName,
+                    $e->getMessage()
+                )
+                . $contentDetails
```
alongosz: If you have already sprintf, concatenation is not necessary, just use sprintf:
```suggestion
                sprintf(
                    'Thumbnail source image generated for %s field and %s variation could not be found (%s).%s',
                    $field->fieldTypeIdentifier,
                    $this->variationName,
                    $e->getMessage(),
                    $contentDetails
                )
```

eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
```diff
@@ -43,12 +53,46 @@ public function getFieldTypeIdentifier(): string
...
+            );
+
+            return null;
+        } catch (Exception $e) {
+            $contentDetails = isset($versionInfo->contentInfo) && isset($versionInfo->versionNo)
+                ? sprintf(' Content: %d, Version No: %d', $versionInfo->contentInfo->id, $versionInfo->versionNo)
+                : '';
```
alongosz: This is a duplicate, so should be extracted to a private method and inline in the `sprintf` statement for both blocks, as described above.

alongosz (changes requested):

eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
barw4: Why though? We know `$logger` is set at this point.

eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
```diff
@@ -57,4 +93,11 @@ public function getThumbnail(Field $field, ?APIVersionInfo $versionInfo = null):
             'mimeType' => $variation->mimeType,
         ]);
     }
+
+    private function generateContentDetailsMessage(?APIVersionInfo $versionInfo)
```
alongosz:
```suggestion
    private function generateContentDetailsMessage(?APIVersionInfo $versionInfo): string
```

eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
Steveb-p: Technically `??` is more accurate. 

`?:` is a shorthand for ternary operator, which operates on `true`/`false` values (and `null` is implicitly cast to `false`). 

`??` checks against `null` without type casting - we have only `LoggerInterface|null`.

eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
```diff
@@ -43,12 +53,38 @@ public function getFieldTypeIdentifier(): string
...
+        } catch (SourceImageNotFoundException $e) {
+            $this->logger->warning(
+                sprintf(
+                    'Thumbnail source image generated for %s field and %s variation could not be found (%s). %s',
+                    $field->fieldTypeIdentifier,
+                    $this->variationName,
+                    $e->getMessage(),
+                    $this->generateContentDetailsMessage($versionInfo)
+                )
+            );
```
Steveb-p: Pass exception to logger context, so this can be picked up by logging tools.
```suggestion
            $this->logger->warning(
                sprintf(
                    'Thumbnail source image generated for %s field and %s variation could not be found (%s). %s',
                    $field->fieldTypeIdentifier,
                    $this->variationName,
                    $e->getMessage(),
                    $this->generateContentDetailsMessage($versionInfo)
                ),
                [
                    'exception' => $e,
                ],
            );
```

eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
```diff
@@ -43,12 +53,38 @@ public function getFieldTypeIdentifier(): string
...
+            return null;
+        } catch (Exception $e) {
+            $this->logger->warning(
+                sprintf(
+                    'Thumbnail could not be generated for %s field and %s variation due to %s. %s',
+                    $field->fieldTypeIdentifier,
+                    $this->variationName,
+                    $e->getMessage(),
+                    $this->generateContentDetailsMessage($versionInfo)
+                )
```
Steveb-p:
```suggestion
                sprintf(
                    'Thumbnail could not be generated for %s field and %s variation due to %s. %s',
                    $field->fieldTypeIdentifier,
                    $this->variationName,
                    $e->getMessage(),
                    $this->generateContentDetailsMessage($versionInfo)
                ),
                [
                    'exception' => $e,
                ],
```

Steveb-p (changes requested):

eZ/Publish/Core/FieldType/Image/ImageThumbnailStrategy.php
Steveb-p: > Why though? We know `$logger` is set at this point.

`isset` (and - by extension - `??` operator) does not only check that variable "is set", but also that it is not `null`.

alongosz (changes requested):

alongosz (approved):

Steveb-p (approved):

micszo (approved):
QA Approved on Ibexa Commerce 3.3.15 with diff.

Also tested on NFS.

PR #293, opened 2022-02-28
[CI] Moved Solr tests to Github Actions

mnocon:
JIRA: https://issues.ibexa.co/browse/IBX-1846

Related PR: https://github.com/ezsystems/ezplatform-solr-search-engine/pull/231/files

Moving the Solr tests to Github Actions, they will be using our Solr cointainer that has been created based on the product-catalog's solr image.


adamwojs (approved):

.github/workflows/ci.yaml
```diff
@@ -181,3 +181,52 @@ jobs:
...
+                    - '8.0'
+                    - '8.1'
+        steps:
+            -   name: Start redis container
+                run: >- 
+                    docker run 
+                    --name redis 
+                    -p 6379:6379 
+                    -d 
+                    --entrypoint redis-server redis --appendfsync no --maxmemory 60mb --save ""
```
Steveb-p: Why was redis container not declared as a GH action service?

See integration tests in product catalog declare postgres as a service:
```yaml
    integration-tests-postgres:
        needs: tests
        services:
            postgres:
                image: postgres:10
                ports:
                    - 5432
                env:
                    POSTGRES_PASSWORD: postgres
                    POSTGRES_DB: testdb
                options: >-
                    --health-cmd pg_isready
                    --health-interval 10s
                    --health-timeout 5s
                    --health-retries 5
                    --tmpfs /var/lib/postgresql/data
```

.github/workflows/ci.yaml
```diff
@@ -181,3 +181,52 @@ jobs:
...
+            
+            - name: Add solr dependency
+              run: composer require --no-update "ezsystems/ezplatform-solr-search-engine:dev-move-to-github-actions as 3.2.x-dev"
+
+            - uses: "ramsey/composer-install@v1"
+              with:
+                  dependency-versions: "highest"
+
+            - name: Init Solr
+              run: vendor/ezsystems/ezplatform-solr-search-engine/.github/init_solr.sh
```
Steveb-p: A pre-initialized solr container is already available:

```yaml
    solr-integration-tests:
        needs: tests
        runs-on: "ubuntu-20.04"
        timeout-minutes: 5
        permissions:
            packages: read
            contents: read

        services:
            solr:
                image: ghcr.io/ibexa/product-catalog/solr:latest
                ports:
                    - 8983
                options: >-
                    --health-cmd "solr status"
                    --health-interval 10s
                    --health-timeout 5s
                    --health-retries 10

```

Granted, build is not "proper" and the namespace of the container is incorrect, but it is usable / buildable.

.github/workflows/ci.yaml
```diff
@@ -181,3 +181,52 @@ jobs:
...
+                    --entrypoint redis-server redis --appendfsync no --maxmemory 60mb --save ""
+    
+            - uses: actions/checkout@v2
+              with:
+                  fetch-depth: 0
+
+            - name: Setup PHP Action
+              uses: shivammathur/setup-php@v2
+              with:
+                  php-version: 7.4
```
Steveb-p: Matrix defines this as tests against multiple PHP versions, yet here we are with a single one.

I guess it's an oversight?

Steveb-p (changes requested):

.github/workflows/ci.yaml
mnocon: Yep, thank you! ~~Will be fixed~~ Fixed in https://github.com/ezsystems/ezplatform-kernel/pull/293/commits/e88f79dd94aadfe82ee5b910da79fedd229f6d1d

.github/workflows/ci.yaml
mnocon: I've tried using it (see: https://github.com/ezsystems/ezplatform-kernel/runs/5373915671 , https://github.com/ezsystems/ezplatform-kernel/commit/39f4aa2448e6eeb137b23bfe94d193fb0b568e17#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR197-R209), but it didn't play nice:
```
Starting solr service container
  /usr/bin/docker --config /home/runner/work/_temp/.docker_858a99ea-2c09-4c40-956b-ec385f6ce7e7 login ghcr.io -u mnocon --password-stdin
  /usr/bin/docker --config /home/runner/work/_temp/.docker_858a99ea-2c09-4c40-956b-ec385f6ce7e7 pull ghcr.io/ibexa/product-catalog/solr:latest
  Error response from daemon: unauthorized
```

At that time I assumed it is not meant to be used in public repositories and dropped the idea, how to approach this?

.github/workflows/ci.yaml
mnocon: I've tried this, but I couldn't find the way to pass the redis config when redis is declared this way. By redis config I mean what's currently set by https://github.com/ezsystems/ezplatform-kernel/blob/1.3/bin%2F.travis%2Fprepare_unittest.sh#L9-L14

The issue is that arguments to the entrypoint should be passed after the image name (I think it's nicely described in https://oprearocks.medium.com/how-to-properly-override-the-entrypoint-using-docker-run-2e081e5feb9d), but no matter what I've tried the end result always ended up similar to this one:
```
  /usr/bin/docker create --name ef6b3612f2c3425d8247c4da73afcff5_redis_c73bf0 --label 7db2d7 --network github_network_e0093ae3985840ef81dc951d61f2b61b --network-alias redis -p 6379:6379 --entrypoint redis-server --appendfsync no --maxmemory 60mb --save "" -e GITHUB_ACTIONS=true -e CI=true redis
  unknown flag: --appendfsync
  See 'docker create --help'.
```
(it's taken from https://github.com/ezsystems/ezplatform-kernel/runs/5373915020)

But in order to work it should look like this:
```
  /usr/bin/docker create --name ef6b3612f2c3425d8247c4da73afcff5_redis_c73bf0 --label 7db2d7 --network github_network_e0093ae3985840ef81dc951d61f2b61b --network-alias redis -p 6379:6379 --entrypoint redis-server -e GITHUB_ACTIONS=true -e CI=true redis --appendfsync no --maxmemory 60mb --save "" 
```

and I haven't found a way how to make this work

.github/workflows/ci.yaml
Steveb-p: Reason this happens is that this image resides in a different repository. We would need to move to a public registry and/or into this repository.

.github/workflows/ci.yaml
mnocon: I guess I could try doing something similar to https://github.community/t/how-do-i-properly-override-a-service-entrypoint/17435/8 (create a redis service and configure it in a separate step), what do you think?

Steveb-p: The same options can be specified in a configuration file instead. We could simply provide an image with added layer, where this config is added.

Or, the same configuration file can be mounted as a service volume: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idservicesservice_idvolumes

Steveb-p (approved):

adamwojs:
Could you please merge up changes @mnocon?

PR #294, opened 2022-03-02
IBX-2382: Changed order command on command lists for Legacy Search

mateuszdebinski:
| **JIRA issue**                          | [IBX-2382](https://issues.ibexa.co/browse/IBX-2382)

#### Checklist:

### DIFF
--- a/eZ/Publish/API/Repository/Tests/SearchEngineIndexingTest.php
+++ b/eZ/Publish/API/Repository/Tests/SearchEngineIndexingTest.php
@@ -837,6 +837,23 @@ class SearchEngineIndexingTest extends BaseTest
     }
 
     /**
+     * Check if FullText indexing works for special cases of text.
+     *
+     * @param string $text Content Item field value text (to be indexed)
+     * @param string $searchForText text based on which Content Item should be found
+     * @param array $ignoreForSetupFactories list of SetupFactories to be ignored
+     * @dataProvider getLegacySpecialFullTextCases
+     */
+    public function testLegacyIndexingSpecialFullTextCases($text, $searchForText)
+    {
+        if (!$this->isLegacySearchEngineSetup()) {
+            return;
+        }
+
+        $this->testIndexingSpecialFullTextCases($text, $searchForText);
+    }
+
+    /**
      * Data Provider for {@see testIndexingSpecialFullTextCases()} method.
      *
      * @return array
@@ -854,6 +871,15 @@ class SearchEngineIndexingTest extends BaseTest
             ['Folder1.', 'Folder1.'],
             ['whitespaces', "     whitespaces  \n \t "],
             ["it's", "it's"],
+        ];
+    }
+
+    /**
+     * Data Provider for {@see testLegacyIndexingSpecialFullTextCases()} method.
+     */
+    public function getLegacySpecialFullTextCases(): array
+    {
+        return [
             ['MAKİNEİÇ', 'makIneIc'],
             ['DIŞ', 'DIŞ'],
             ['TİC', 'tIc'],
--- a/eZ/Publish/API/Repository/Tests/SearchEngineIndexingTest.php
+++ b/eZ/Publish/API/Repository/Tests/SearchEngineIndexingTest.php
@@ -854,7 +854,11 @@ class SearchEngineIndexingTest extends BaseTest
             ['Folder1.', 'Folder1.'],
             ['whitespaces', "     whitespaces  \n \t "],
             ["it's", "it's"],
-            ['with_underscore', 'with_underscore'],
+            ['MAKİNEİÇ', 'makIneIc'],
+            ['DIŞ', 'DIŞ'],
+            ['TİC', 'tIc'],
+            ['ŞTİ.', 'stI'],
+            ['ʻ', 'ʻ'],
         ];
     }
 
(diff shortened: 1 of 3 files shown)


alongosz (changes requested):
Missing integration test coverage.

eZ/Publish/API/Repository/Tests/SearchEngineIndexingTest.php
```diff
@@ -836,6 +836,23 @@ public function testIndexingSpecialFullTextCases($text, $searchForText)
...
+     * Check if FullText indexing works for special cases of text.
+     *
+     * @param string $text Content Item field value text (to be indexed)
+     * @param string $searchForText text based on which Content Item should be found
+     * @param array $ignoreForSetupFactories list of SetupFactories to be ignored
+     * @dataProvider getLegacySpecialFullTextCases
+     */
+    public function testLegacyIndexingSpecialFullTextCases($text, $searchForText)
+    {
+        if (!$this->isLegacySearchEngineSetup()) {
```
alongosz: why?

alongosz (changes requested):

eZ/Publish/API/Repository/Tests/SearchEngineIndexingTest.php
mateuszdebinski: Because only Legacy Search uses the transformation Processor for indexing, which causes a problem and for some characters, as in the test, the resulting word is different than for SOLR or ElasticSearch

alongosz: > Because only Legacy Search uses the transformation Processor for indexing, which causes a problem and for some characters, as in the test, the resulting word is different than for SOLR or ElasticSearch

Unfortunately then this solution is not acceptable. The test in the form given was there to ensure such difference does not exist. Every engine needs to work uniformly for this use case.

alongosz (dismissed):

eZ/Publish/API/Repository/Tests/SearchEngineIndexingTest.php
mateuszdebinski: After verification, I do not see any reason why the transformations were performed during the indexing for legacy search. From what I checked, removing this does not cause any problems, and now consistency is maintained for search engines between legacy, SOLR, and Elastic - I tested it manually and it is possible to search correctly

Steveb-p: Have you tried Postgresql / MySQL / MariaDB?

There are subtle differences coming from different database types / indexes. Specifically case sensivity, which might be the reason for this particular transformation to take place.

We did encounter this when working with Product Catalog, which made it necessary for us to introduce a separate field for a normalized name, because otherwise database indices would be unusable.

mateuszdebinski: @Steveb-p @alongosz 
As for the case sensivity, the transformationProcessor with ruleGroup: lowercase has been added. If we want to have consistent results for SOLR ELASTIC and Legacy search, unfortunately, this transformation needs to be removed. I have done tests on MySQL and also tests in PR for PostgreSQL, MySQL, and SQLite are green. Unfortunately, I do not see any other solution that would be able to have the same results for all supported search engines, the current solution that exists in the product already showed errors with a few special characters, which excludes them as correct

eZ/Publish/Core/Search/Legacy/Content/WordIndexer/Gateway/DoctrineDatabase.php
```diff
@@ -110,12 +110,7 @@ public function index(FullTextData $fullTextData)
             } else {
                 $integerValue = 0;
             }
-            $text = $this->transformationProcessor->transform(
-                $fullTextValue->value,
-                !empty($fullTextValue->transformationRules)
-                    ? $fullTextValue->transformationRules
-                    : $this->fullTextSearchConfiguration['commands']
-            );
+            $text = $this->transformationProcessor->transformByGroup($fullTextValue->value, 'lowercase');
```
alongosz: `$fullTextValue->transformationRules` is an extension point to add some custom logic for transforming values

alongosz (commented):
Hmm, this doesn't add up TBH. Me or other maintainer should probably organize some time to look into that.

eZ/Publish/Core/Search/Legacy/Content/WordIndexer/Gateway/DoctrineDatabase.php
Steveb-p: That is indeed true. I think what you want done here is to apply both transformations - the one we require (lowercase), followed by any custom transformations from `transformationRules`. Correct?

EDIT: Or the other way around.

Steveb-p:
```suggestion
            $text = $this->transformationProcessor->transform(
                $fullTextValue->value,
                !empty($fullTextValue->transformationRules)
                    ? $fullTextValue->transformationRules
                    : $this->fullTextSearchConfiguration['commands']
            );
            $text = $this->transformationProcessor->transformByGroup($text, 'lowercase');
```

eZ/Publish/API/Repository/Tests/SearchEngineIndexingTest.php
```diff
@@ -854,7 +854,11 @@ public function getSpecialFullTextCases()
             ['Folder1.', 'Folder1.'],
             ['whitespaces', "     whitespaces  \n \t "],
             ["it's", "it's"],
-            ['with_underscore', 'with_underscore'],
```
ViniTou, line 857: this should not be dropped, right?

ViniTou (approved):

mikadamczyk (approved):

konradoboza (approved):

Steveb-p (approved):

mateuszdebinski:
I'm closing due to a transfer to this PR: https://github.com/ezsystems/ezpublish-kernel/pull/3144

PR #305, opened 2022-05-04
IBX-2825: Implemented `Location` target

barw4:
| **JIRA issue**                          | [IBX-2825](https://issues.ibexa.co/browse/IBX-2825)

`Location` target in Language Limitation will now verify if all content's translations are contained in the current user's languages in his Language Limitation. If that is the case the Limitation will pass, if not it will deny the access.

#### Checklist:

### DIFF
--- a/eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
+++ b/eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
@@ -650,7 +650,7 @@ class LanguageLimitationTest extends BaseTest
      * @throws \eZ\Publish\API\Repository\Exceptions\ForbiddenException
      * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
      */
-    public function testHideAndUnhideLocationAndContentWithLanguageLimitationAndDifferentContentTranslations(
+    public function testHideAndUnhideLocationWithLanguageLimitationAndDifferentContentTranslations(
         array $limitationValues,
         bool $containsAllTranslations
     ): void {
@@ -681,7 +681,38 @@ class LanguageLimitationTest extends BaseTest
                 $content->getVersionInfo()->languageCodes,
                 $revealedLocation->getContent()->getVersionInfo()->languageCodes
             );
+        } else {
+            self::expectException(UnauthorizedException::class);
 
+            $locationService->hideLocation($location);
+
+            self::expectException(UnauthorizedException::class);
+
+            $locationService->unhideLocation($location);
+        }
+    }
+
+    /**
+     * @dataProvider providerForPrepareDataForTestsWithLanguageLimitationAndDifferentContentTranslations
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
+     * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
+     * @throws \eZ\Publish\API\Repository\Exceptions\ForbiddenException
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
+     */
+    public function testHideAndRevealContentWithLanguageLimitationAndDifferentContentTranslations(
+        array $limitationValues,
+        bool $containsAllTranslations
+    ): void {
+        $repository = $this->getRepository();
+        $contentService = $repository->getContentService();
+
+        $content = $this->testPrepareDataForTestsWithLanguageLimitationAndDifferentContentTranslations(
+            $limitationValues,
+            $containsAllTranslations
+        );
+
+        if ($containsAllTranslations) {
             $contentService->hideContent($content->contentInfo);
 
             $content = $contentService->loadContent($content->id);
@@ -694,14 +725,6 @@ class LanguageLimitationTest extends BaseTest
 
             self::assertFalse($content->contentInfo->isHidden);
         } else {
-            self::expectException(UnauthorizedException::class);
-
-            $locationService->hideLocation($location);
-
-            self::expectException(UnauthorizedException::class);
-
-            $locationService->unhideLocation($location);
-
             self::expectException(UnauthorizedException::class);
 
             $contentService->hideContent($content->contentInfo);
--- a/eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
+++ b/eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
@@ -12,6 +12,7 @@ use eZ\Publish\API\Repository\ContentService;
 use eZ\Publish\API\Repository\Exceptions\UnauthorizedException;
 use eZ\Publish\API\Repository\Tests\BaseTest;
 use eZ\Publish\API\Repository\Values\Content\Content;
+use eZ\Publish\API\Repository\Values\Content\LocationCreateStruct;
 use eZ\Publish\API\Repository\Values\User\Limitation\LanguageLimitation;
 use eZ\Publish\API\Repository\Values\User\User;
 use eZ\Publish\SPI\Limitation\Target\Builder\VersionBuilder;
@@ -66,6 +67,7 @@ class LanguageLimitationTest extends BaseTest
                 ['module' => 'content', 'function' => 'create', 'limitations' => $limitations],
                 ['module' => 'content', 'function' => 'edit', 'limitations' => $limitations],
                 ['module' => 'content', 'function' => 'publish', 'limitations' => $limitations],
+                ['module' => 'content', 'function' => 'manage_locations'],
             ]
         );
     }
@@ -488,6 +490,30 @@ class LanguageLimitationTest extends BaseTest
         $this->expectException(UnauthorizedException::class);
         $this->expectExceptionMessage("The User does not have the 'publish' 'content' permission");
         $contentService->publishVersion($draft->versionInfo, [self::GER_DE, self::ENG_GB]);
+    }
+
+    /**
+     * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
+     * @throws \eZ\Publish\API\Repository\Exceptions\ForbiddenException
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
+     */
+    public function testCopyContentWithLanguageLimitationDifferentThanInitialLanguageCode(): void
+    {
+        $repository = $this->getRepository();
+        $contentService = $repository->getContentService();
+        $permissionResolver = $repository->getPermissionResolver();
+
+        $draft = $this->createMultilingualFolderDraft($contentService);
+        $content = $contentService->publishVersion($draft->getVersionInfo());
+
+        $permissionResolver->setCurrentUserReference(
+            $this->createEditorUserWithLanguageLimitation([self::GER_DE])
+        );
+
+        $locationCreateStruct = new LocationCreateStruct(['parentLocationId' => 2]);
+        $content = $contentService->copyContent($content->contentInfo, $locationCreateStruct);
+
+        self::assertInstanceOf(Content::class, $content);
     }
 
     /**
--- a/eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
+++ b/eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
@@ -721,6 +721,9 @@ class LanguageLimitationTest extends BaseTest
         return $content;
     }
 
+    /**
+     * @return iterable<array{array<string>, bool}>
+     */
     public function providerForPrepareDataForTestsWithLanguageLimitationAndDifferentContentTranslations(): array
     {
         return [
--- a/eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
+++ b/eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
@@ -494,99 +494,215 @@ class LanguageLimitationTest extends BaseTest
     }
 
     /**
+     * @dataProvider providerForPrepareDataForTestsWithLanguageLimitationAndDifferentContentTranslations
+     *
      * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
      * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
      * @throws \eZ\Publish\API\Repository\Exceptions\ForbiddenException
      * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
      */
-    public function testCopyContentWithTranslationsDifferentThanLanguageLimitation(): void
-    {
+    public function testCopyContentWithLanguageLimitationAndDifferentContentTranslations(
+        array $limitationValues,
+        bool $containsAllTranslations
+    ): void {
         $repository = $this->getRepository();
         $contentService = $repository->getContentService();
 
-        $content = $this->testPrepareDataForTestsWithTranslationsDifferentThanLanguageLimitation();
+        $content = $this->testPrepareDataForTestsWithLanguageLimitationAndDifferentContentTranslations(
+            $limitationValues,
+            $containsAllTranslations
+        );
 
         $locationCreateStruct = new LocationCreateStruct(['parentLocationId' => 2]);
 
-        self::expectException(UnauthorizedException::class);
+        if ($containsAllTranslations) {
+            $clonedContent = $contentService->copyContent($content->contentInfo, $locationCreateStruct);
 
-        $contentService->copyContent($content->contentInfo, $locationCreateStruct);
+            self::assertSame($content->getVersionInfo()->languageCodes, $clonedContent->getVersionInfo()->languageCodes);
+        } else {
+            self::expectException(UnauthorizedException::class);
+
+            $contentService->copyContent($content->contentInfo, $locationCreateStruct);
+        }
     }
 
     /**
+     * @dataProvider providerForPrepareDataForTestsWithLanguageLimitationAndDifferentContentTranslations
+     *
      * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
      * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
      * @throws \eZ\Publish\API\Repository\Exceptions\ForbiddenException
      * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
      * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
      */
-    public function testCopySubtreeWithTranslationsDifferentThanLanguageLimitation(): void
-    {
+    public function testCopySubtreeWithLanguageLimitationAndDifferentContentTranslations(
+        array $limitationValues,
+        bool $containsAllTranslations
+    ): void {
         $repository = $this->getRepository();
         $locationService = $repository->getLocationService();
 
-        $content = $this->testPrepareDataForTestsWithTranslationsDifferentThanLanguageLimitation();
+        $content = $this->testPrepareDataForTestsWithLanguageLimitationAndDifferentContentTranslations(
+            $limitationValues,
+            $containsAllTranslations
+        );
 
         $contentLocation = $locationService->loadLocation($content->contentInfo->mainLocationId);
         $targetLocation = $locationService->loadLocation(2);
 
-        self::expectException(UnauthorizedException::class);
+        if ($containsAllTranslations) {
+            $location = $locationService->copySubtree($contentLocation, $targetLocation);
 
-        $locationService->copySubtree($contentLocation, $targetLocation);
+            self::assertSame(
+                $content->getVersionInfo()->languageCodes,
+                $location->getContent()->getVersionInfo()->languageCodes
+            );
+        } else {
+            self::expectException(UnauthorizedException::class);
+
+            $locationService->copySubtree($contentLocation, $targetLocation);
+        }
     }
 
     /**
+     * @dataProvider providerForPrepareDataForTestsWithLanguageLimitationAndDifferentContentTranslations
+     *
      * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
      * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
      * @throws \eZ\Publish\API\Repository\Exceptions\ForbiddenException
      * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
      * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
      */
-    public function testMoveSubtreeWithTranslationsDifferentThanLanguageLimitation(): void
-    {
+    public function testMoveSubtreeWithLanguageLimitationAndDifferentContentTranslations(
+        array $limitationValues,
+        bool $containsAllTranslations
+    ): void {
         $repository = $this->getRepository();
         $locationService = $repository->getLocationService();
         $contentService = $repository->getContentService();
 
         $targetContent = $this->createMultilingualFolderDraft($contentService);
-        $content = $this->testPrepareDataForTestsWithTranslationsDifferentThanLanguageLimitation();
+        $content = $this->testPrepareDataForTestsWithLanguageLimitationAndDifferentContentTranslations(
+            $limitationValues,
+            $containsAllTranslations
+        );
 
         $contentLocation = $locationService->loadLocation($content->contentInfo->mainLocationId);
         $targetLocation = $locationService->loadLocation($targetContent->contentInfo->mainLocationId);
 
-        self::expectException(UnauthorizedException::class);
+        if ($containsAllTranslations) {
+            $locationService->moveSubtree($contentLocation, $targetLocation);
+            $targetLocation = $locationService->loadLocation($targetLocation->id);
 
-        $locationService->moveSubtree($contentLocation, $targetLocation);
+            self::assertSame(
+                $content->getVersionInfo()->languageCodes,
+                $targetLocation->getContent()->getVersionInfo()->languageCodes
+            );
+        } else {
+            self::expectException(UnauthorizedException::class);
+
+            $locationService->moveSubtree($contentLocation, $targetLocation);
+        }
     }
 
     /**
+     * @dataProvider providerForPrepareDataForTestsWithLanguageLimitationAndDifferentContentTranslations
+     *
      * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
      * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
      * @throws \eZ\Publish\API\Repository\Exceptions\ForbiddenException
      * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
      */
-    public function testHideLocationWithTranslationsDifferentThanLanguageLimitation(): void
-    {
+    public function testHideLocationWithLanguageLimitationAndDifferentContentTranslations(
+        array $limitationValues,
+        bool $containsAllTranslations
+    ): void {
         $repository = $this->getRepository();
         $locationService = $repository->getLocationService();
 
-        $content = $this->testPrepareDataForTestsWithTranslationsDifferentThanLanguageLimitation();
+        $content = $this->testPrepareDataForTestsWithLanguageLimitationAndDifferentContentTranslations(
+            $limitationValues,
+            $containsAllTranslations
+        );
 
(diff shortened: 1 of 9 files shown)


adamwojs:
@alongosz What is your POV here?

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
```diff
@@ -490,6 +492,30 @@ public function testPublishVersionTranslationWhenUserHasNoAccessToAllLanguages()
...
+        $repository = $this->getRepository();
+        $contentService = $repository->getContentService();
+        $permissionResolver = $repository->getPermissionResolver();
+
+        $draft = $this->createMultilingualFolderDraft($contentService);
+        $content = $contentService->publishVersion($draft->getVersionInfo());
+
+        $permissionResolver->setCurrentUserReference(
+            $this->createEditorUserWithLanguageLimitation([self::GER_DE])
+        );
```
alongosz: in between we should have here some sanity check, that `initialLanguageCode` is different than `GER_DE`.

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
```diff
@@ -490,6 +492,30 @@ public function testPublishVersionTranslationWhenUserHasNoAccessToAllLanguages()
...
+
+        $locationCreateStruct = new LocationCreateStruct(['parentLocationId' => 2]);
+        $content = $contentService->copyContent($content->contentInfo, $locationCreateStruct);
+
+        self::assertInstanceOf(Content::class, $content);
```
alongosz: This assert will never throw because `copyContent` has strict `: Content` return type (it will result in a fatal error on a `copyContent` call). Maybe some Content load by ID with assertion of existing translations in a new subtree?

eZ/Publish/Core/Limitation/LanguageLimitationType.php
```diff
@@ -152,13 +153,19 @@ public function evaluate(
...
             }
 
-            $accessVote = $this->evaluateVersionTarget($target, $value);
+            if ($target instanceof Target\Version) {
+                $accessVote = $this->evaluateVersionTarget($target, $value);
+            } elseif ($target instanceof DestinationLocationTarget) {
+                $accessVote = $this->evaluateLocationTarget($target, $value);
+            } else {
+                continue;
+            }
```
alongosz, line 168: Note: At some point we would need to think about injecting strategy here, but given the time constraints and this being 3.3 LTS, ok with me for now.

src/contracts/Limitation/Target/DestinationLocation.php
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * @var int
+     */
+    protected $id;
+
+    /**
+     * @var \eZ\Publish\API\Repository\Values\Content\ContentInfo
+     */
+    protected $targetContentInfo;
+}
```
alongosz: 1. `final`
2. Quite fresh - we're getting performance hits due to `ValueObject::__get` magic. We should avoid it. Refactor this please to:

    - use strict constructor
    - use strict getters for the properties
    - it still can extend `ValueObject` for the consistency's sake, but should not utilize its magic.
    - CS: `/** @var <type> */` instead of multiline PHPDoc for the _properties_.

alongosz (changes requested):
Looks much cleaner than the previous version :muscle: 
Some remarks:

mikadamczyk (approved):

src/contracts/Limitation/Target/DestinationLocation.php
```diff
@@ -0,0 +1,43 @@
...
+use eZ\Publish\SPI\Limitation\Target;
+use eZ\Publish\SPI\Persistence\ValueObject;
+
+/**
+ * Location Limitation target.
+ */
+final class DestinationLocation extends ValueObject implements Target
+{
+    /** @var int */
+    private $id;
```
adamwojs: Please consider renaming `$id` property to one of the following suggestions 

```suggestion
    private $locationId;
```

```suggestion
    private $targetLocationId;
```

adamwojs (approved):

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
```diff
@@ -490,6 +493,129 @@ public function testPublishVersionTranslationWhenUserHasNoAccessToAllLanguages()
...
+     */
+    private function prepareDataForLocationRelatedTests(): Content
+    {
+        $repository = $this->getRepository();
+        $contentService = $repository->getContentService();
+        $permissionResolver = $repository->getPermissionResolver();
+
+        $draft = $this->createMultilingualFolderDraft($contentService);
+        $content = $contentService->publishVersion($draft->getVersionInfo());
+
```
alongosz, line 805: > in between we should have here some sanity check, that `initialLanguageCode` is different than `GER_DE`.

So... what I meant by this comment is that somewhere here, there should be a check that the actual `initialLanguageCode` is not the one from the Limitation values.

```suggestion
        self::assertNotSame(self::GER_DE, $content->getVersionInfo()->initialLanguageCode);

```

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
```diff
@@ -490,6 +493,129 @@ public function testPublishVersionTranslationWhenUserHasNoAccessToAllLanguages()
...
+        );
+    }
+
+    /**
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
+     * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
+     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
+     * @throws \eZ\Publish\API\Repository\Exceptions\ForbiddenException
+     */
+    private function prepareDataForLocationRelatedTests(): Content
```
alongosz: Following the previous comment, this should be less generic. E.g:
```suggestion
    private function prepareDataForTestsOfLimitationWithDifferentInitialLanguageCode(): Content
```
The name is a bit too long, but ATM have no better idea how to describe it briefly ;)

eZ/Publish/Core/Limitation/LanguageLimitationType.php
```diff
@@ -152,13 +153,19 @@ public function evaluate(
             $targets = [];
```
alongosz, line 153: Since you're fixing almost the entire limitation logic, could you also change the line above this one to use strict `===` comparison? It was a mistake.

eZ/Publish/Core/Limitation/LanguageLimitationType.php
```diff
@@ -187,12 +194,8 @@ private function evaluateObject(ValueObject $object, APILimitationValue $value):
...
-                $object = $this->persistenceContentHandler->loadVersionInfo(
-                    $object->id,
-                    $object->currentVersionNo
-                );
-            } catch (NotFoundException $e) {
+            $object = $this->loadVersionInfo($object);
+            if ($object === null) {
                 return self::ACCESS_DENIED;
             }
         }
```
alongosz, line 201: Minor issue: line 207 (can't add diff comment) - `in_array` should be called with 3rd argument. Requesting to fix for the same reasons as mentioned before - a low hanging fruit.

eZ/Publish/Core/Limitation/LanguageLimitationType.php
```diff
@@ -209,6 +212,18 @@ private function evaluateObject(ValueObject $object, APILimitationValue $value):
         return $accessVote;
     }
 
+    private function loadVersionInfo(ContentInfo $contentInfo): ?SPIVersionInfo
```
alongosz: CS/Naming: a long while ago we had an interesting naming convention. For the cases where you just tried to load something and return null if failed using exception-driven logic, you would name the method as following:
```suggestion
    private function tryLoadingVersionInfo(ContentInfo $contentInfo): ?SPIVersionInfo
```

eZ/Publish/Core/Limitation/LanguageLimitationType.php
```diff
@@ -255,6 +270,20 @@ private function evaluateVersionTarget(
...
+    private function evaluateLocationTarget(
+        DestinationLocationTarget $location,
+        APILimitationValue $value
+    ): ?bool {
+        $versionInfo = $this->loadVersionInfo($location->getTargetContentInfo());
+        if ($versionInfo === null) {
+            return self::ACCESS_DENIED;
+        }
+
+        return array_intersect($versionInfo->languageCodes, $value->limitationValues)
```
alongosz: This logic here makes me go back to the essence of the issue. While there's a bug there, is the expected result from [IBX-2825](https://issues.ibexa.co/browse/IBX-2825) a valid one?

To sum up, it states that if you have `[eng-GB, fre-FR]` multilingual article, but user has `fre-FR` only `LanguageLimitation`
> **Expected result:**
> You should be able to copy the article without any problems.

Let's consider the following scenario:
* As an Editor you are explicitly forbidden to create Content other than French in a "Subtree A".
* However you can create English and French Content in "Subtree B"
* After those changes, to work around your permissions for "Subtree A", it would be enough to create a content in "Subtree B" and copy/move it to "Subtree A".

I understand that this limits Editor actions just because some Content has other translations, but in such cases shouldn't any such action be delegated to a supervisor who has all the necessary permissions?

I would rather say that:
> **Expected result:**
> You should not be able to copy the article, if you don't have permission to manage all its translations

eZ/Publish/Core/Repository/LocationService.php
```diff
@@ -147,7 +148,13 @@ public function copySubtree(APILocation $subtree, APILocation $targetParentLocat
...
-        if (!$this->permissionResolver->canUser('content', 'create', $loadedSubtree->getContentInfo(), [$loadedTargetLocation])) {
+        $locationTarget = (new DestinationLocationTarget($targetParentLocation->id, $loadedSubtree->contentInfo));
+        if (!$this->permissionResolver->canUser(
+            'content',
+            'create',
+            $loadedSubtree->getContentInfo(),
+            [$loadedTargetLocation, $locationTarget]
+        )) {
             throw new UnauthorizedException('content', 'create', ['locationId' => $loadedTargetLocation->id]);
         }
```
alongosz: We're modifying here lines which contain duplicated code fragment, adding more duplication, which makes Sonar gate fail. Lines 151-183 and 729-761 could be extracted into a single common method:
```php
    private function checkCreatePermissionOnSubtreeTarget(
        APILocation $targetParentLocation,
        APILocation $loadedSubtree,
        APILocation $loadedTargetLocation
    ): void {
        $locationTarget = (new DestinationLocationTarget(
            $targetParentLocation->id,
            $loadedSubtree->contentInfo
        ));
        if (!$this->permissionResolver->canUser(
            'content',
            'create',
            $loadedSubtree->getContentInfo(),
            [$loadedTargetLocation, $locationTarget]
        )) {
            throw new UnauthorizedException(
                'content',
                'create',
                ['locationId' => $loadedTargetLocation->id]
            );
        }

        // Check read access to whole source subtree
        $contentReadCriterion = $this->permissionCriterionResolver->getPermissionsCriterion(
            'content',
            'read'
        );
        if ($contentReadCriterion === false) {
            throw new UnauthorizedException('content', 'read');
        } elseif ($contentReadCriterion !== true) {
            // Query if there are any content in subtree current user don't have access to
            $query = new Query(
                [
                    'limit' => 0,
                    'filter' => new CriterionLogicalAnd(
                        [
                            new CriterionSubtree($loadedSubtree->pathString),
                            new CriterionLogicalNot($contentReadCriterion),
                        ]
                    ),
                ]
            );
            $result = $this->repository->getSearchService()->findContent($query, [], false);
            if ($result->totalCount > 0) {
                throw new UnauthorizedException('content', 'read');
            }
        }
    }
```

src/contracts/Limitation/Target/DestinationLocation.php
alongosz: > Please consider renaming `$id` property to one of the following suggestions

:+1: :point_up:

alongosz (changes requested):
Besides some remarks, I have a serious concern about breaking permission logic here, as mentioned in the diff. It requires confirmation and/or further discussion.

eZ/Publish/Core/Limitation/LanguageLimitationType.php
barw4: Solution has been refactored according to JIRA and meetings.

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
```diff
@@ -490,6 +493,164 @@ public function testPublishVersionTranslationWhenUserHasNoAccessToAllLanguages()
...
+
+        // Hide
+        $hiddenLocation = $locationService->hideLocation($targetLocation);
+        $hiddenLocationContent = $hiddenLocation->getContent();
+        self::assertSame($content->getVersionInfo()->languageCodes, $hiddenLocationContent->getVersionInfo()->languageCodes);
+
+        // Unhide
+        $revealedLocation = $locationService->unhideLocation($hiddenLocation);
+        $revealedLocationContent = $revealedLocation->getContent();
+        self::assertSame($content->getVersionInfo()->languageCodes, $revealedLocationContent->getVersionInfo()->languageCodes);
```
alongosz: Please split this into separate tests. In this case it's justified to use `@depends`

eZ/Publish/Core/Limitation/LanguageLimitationType.php
```diff
@@ -255,6 +270,20 @@ private function evaluateVersionTarget(
...
+    private function evaluateLocationTarget(
+        DestinationLocationTarget $location,
+        APILimitationValue $value
+    ): ?bool {
+        $versionInfo = $this->tryLoadingVersionInfo($location->getTargetContentInfo());
+        if ($versionInfo === null) {
+            return self::ACCESS_DENIED;
+        }
+
+        return !array_diff($versionInfo->languageCodes, $value->limitationValues)
```
alongosz: Implicit casting from any type (here array) to boolean tends to be error prone, so to develop good habits, we should avoid them, even if in any specific case they would work.
```suggestion
        return empty(array_diff($versionInfo->languageCodes, $value->limitationValues))
```

alongosz (changes requested):

eZ/Publish/Core/Limitation/LanguageLimitationType.php
Steveb-p: @alongosz `in_array` with 3rd argument is part of the phpstan strict rules extension.

https://github.com/phpstan/phpstan-strict-rules

We could add these rules on the case-by-case basis, because some directly contradict our current style (for example disallowed `empty()`)

alongosz: > @alongosz `in_array` with 3rd argument is part of the phpstan strict rules extension.
> 
> https://github.com/phpstan/phpstan-strict-rules
> 
> We could add these rules on the case-by-case basis, because some directly contradict our current style (for example disallowed `empty()`)

Fully agree, but it seems to be out of scope for a Customer Request.

alongosz (approved):

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
```diff
@@ -490,6 +493,242 @@ public function testPublishVersionTranslationWhenUserHasNoAccessToAllLanguages()
...
+            : self::assertNotEmpty(array_diff($content->getVersionInfo()->languageCodes, $limitationValues));
+
+        $permissionResolver->setCurrentUserReference(
+            $this->createEditorUserWithLanguageLimitation($limitationValues)
+        );
+
+        return $content;
+    }
+
+    public function providerForPrepareDataForTestsWithLanguageLimitationAndDifferentContentTranslations(): array
```
Steveb-p, line 820: Describe data sets shape, so they can be statically checked once we get phpstan on sufficient level:

```suggestion
    /**
     * @return iterable<array{array<string>, bool}>
     */
    public function providerForPrepareDataForTestsWithLanguageLimitationAndDifferentContentTranslations(): iterable
```

This describes the returned type as:
1. An `iterable` (which we can use, `array` is an `iterable`)...
2. ...which will contain two-element arrays as elements...
3. ...which will consist of an array of strings as the first element, and boolean as second.

Optionally you can add descriptive keys to data sets.

Steveb-p (approved):

mikadamczyk (approved):

barw4:
Additional review is welcomed, so pinging @alongosz, @mikadamczyk, @Steveb-p as targets had to be added to Content hide/reveal methods.

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
```diff
@@ -650,12 +650,13 @@ public function testHideLocationWithLanguageLimitationAndDifferentContentTransla
      * @throws \eZ\Publish\API\Repository\Exceptions\ForbiddenException
      * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
      */
-    public function testHideAndUnhideLocationWithLanguageLimitationAndDifferentContentTranslations(
+    public function testHideAndUnhideLocationAndContentWithLanguageLimitationAndDifferentContentTranslations(
```
alongosz: Too much happens in this test now, it should be split. If you want to use previous test return value you can use `@depends`.

alongosz (changes requested):

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
```diff
@@ -490,6 +493,266 @@ public function testPublishVersionTranslationWhenUserHasNoAccessToAllLanguages()
...
+            $locationService->unhideLocation($location);
+            $revealedLocation = $locationService->loadLocation($location->id);
+
+            self::assertSame(
+                $content->getVersionInfo()->languageCodes,
+                $revealedLocation->getContent()->getVersionInfo()->languageCodes
+            );
+
+            $contentService->hideContent($content->contentInfo);
+
```
Steveb-p, line 720: Nitpick: overuse of whitespace. Once you start separating each line with an additional newline, the additional "readability" is lost. Use this separation for logical grouping of operations, or not at all.

Steveb-p (approved):

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
barw4: I cannot use `@depends` as a repository is cleaned after each use. Loading content/location won't work in the next test.

alongosz (approved):

barw4:
As @micszo has noticed (thanks) we are missing the Location target for `swapLocation`. Calling for re-review again :( @alongosz @Steveb-p

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
```diff
@@ -490,6 +493,327 @@ public function testPublishVersionTranslationWhenUserHasNoAccessToAllLanguages()
...
+                $revealedLocation->getContent()->getVersionInfo()->languageCodes
+            );
+        } else {
+            self::expectException(UnauthorizedException::class);
+
+            $locationService->hideLocation($location);
+
+            self::expectException(UnauthorizedException::class);
+
+            $locationService->unhideLocation($location);
```
alongosz: Just noticed, this block is never gonna be called, because an exception is thrown at line 725.

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
```diff
@@ -490,6 +493,327 @@ public function testPublishVersionTranslationWhenUserHasNoAccessToAllLanguages()
...
+
+            self::assertFalse($content->contentInfo->isHidden);
+        } else {
+            self::expectException(UnauthorizedException::class);
+
+            $contentService->hideContent($content->contentInfo);
+
+            self::expectException(UnauthorizedException::class);
+
+            $contentService->revealContent($content->contentInfo);
```
alongosz: same here

alongosz (changes requested):

micszo (approved):
QA Approved on Ibexa Experience v3.3.19.

Tested actions related to copying, moving, hiding and locations.

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
```diff
@@ -490,6 +493,338 @@ public function testPublishVersionTranslationWhenUserHasNoAccessToAllLanguages()
...
+        );
+
+        $locationCreateStruct = new LocationCreateStruct(['parentLocationId' => 2]);
+
+        if ($containsAllTranslations) {
+            $clonedContent = $contentService->copyContent($content->contentInfo, $locationCreateStruct);
+
+            self::assertSame($content->getVersionInfo()->languageCodes, $clonedContent->getVersionInfo()->languageCodes);
+        } else {
+            self::expectException(UnauthorizedException::class);
```
Steveb-p: This method is not static (despite not actually using any object properties).
```suggestion
            $this->expectException(UnauthorizedException::class);
```

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
```diff
@@ -490,6 +493,338 @@ public function testPublishVersionTranslationWhenUserHasNoAccessToAllLanguages()
...
+
+        if ($containsAllTranslations) {
+            $location = $locationService->copySubtree($contentLocation, $targetLocation);
+
+            self::assertSame(
+                $content->getVersionInfo()->languageCodes,
+                $location->getContent()->getVersionInfo()->languageCodes
+            );
+        } else {
+            self::expectException(UnauthorizedException::class);
```
Steveb-p:
```suggestion
            $this->expectException(UnauthorizedException::class);
```

alongosz (approved):

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
```diff
@@ -490,6 +493,338 @@ public function testPublishVersionTranslationWhenUserHasNoAccessToAllLanguages()
...
+        if ($containsAllTranslations) {
+            $locationService->unhideLocation($location);
+            $revealedLocation = $locationService->loadLocation($location->id);
+
+            self::assertSame(
+                $content->getVersionInfo()->languageCodes,
+                $revealedLocation->getContent()->getVersionInfo()->languageCodes
+            );
+        } else {
+            self::expectException(UnauthorizedException::class);
```
Steveb-p:
```suggestion
            $this->expectException(UnauthorizedException::class);
```

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
```diff
@@ -490,6 +493,338 @@ public function testPublishVersionTranslationWhenUserHasNoAccessToAllLanguages()
...
+            $locationService->swapLocation($location2, $location);
+
+            $location = $locationService->loadLocation($location2->id);
+
+            self::assertSame(
+                $content->getVersionInfo()->languageCodes,
+                $location->getContent()->getVersionInfo()->languageCodes
+            );
+        } else {
+            self::expectException(UnauthorizedException::class);
```
Steveb-p:
```suggestion
            $this->expectException(UnauthorizedException::class);
```

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
```diff
@@ -490,6 +493,338 @@ public function testPublishVersionTranslationWhenUserHasNoAccessToAllLanguages()
...
+        if ($containsAllTranslations) {
+            $locationService->moveSubtree($contentLocation, $targetLocation);
+            $targetLocation = $locationService->loadLocation($targetLocation->id);
+
+            self::assertSame(
+                $content->getVersionInfo()->languageCodes,
+                $targetLocation->getContent()->getVersionInfo()->languageCodes
+            );
+        } else {
+            self::expectException(UnauthorizedException::class);
```
Steveb-p:
```suggestion
            $this->expectException(UnauthorizedException::class);
```

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
```diff
@@ -490,6 +493,338 @@ public function testPublishVersionTranslationWhenUserHasNoAccessToAllLanguages()
...
+        if ($containsAllTranslations) {
+            $locationService->hideLocation($location);
+            $hiddenLocation = $locationService->loadLocation($location->id);
+
+            self::assertSame(
+                $content->getVersionInfo()->languageCodes,
+                $hiddenLocation->getContent()->getVersionInfo()->languageCodes
+            );
+        } else {
+            self::expectException(UnauthorizedException::class);
```
Steveb-p:
```suggestion
            $this->expectException(UnauthorizedException::class);
```

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
```diff
@@ -490,6 +493,338 @@ public function testPublishVersionTranslationWhenUserHasNoAccessToAllLanguages()
...
+        );
+
+        if ($containsAllTranslations) {
+            $contentService->hideContent($content->contentInfo);
+
+            $content = $contentService->loadContent($content->id);
+
+            self::assertTrue($content->contentInfo->isHidden);
+        } else {
+            self::expectException(UnauthorizedException::class);
```
Steveb-p:
```suggestion
            $this->expectException(UnauthorizedException::class);
```

eZ/Publish/API/Repository/Tests/Values/User/Limitation/LanguageLimitationTest.php
```diff
@@ -490,6 +493,338 @@ public function testPublishVersionTranslationWhenUserHasNoAccessToAllLanguages()
...
+        );
+
+        if ($containsAllTranslations) {
+            $contentService->revealContent($content->contentInfo);
+
+            $content = $contentService->loadContent($content->id);
+
+            self::assertFalse($content->contentInfo->isHidden);
+        } else {
+            self::expectException(UnauthorizedException::class);
```
Steveb-p:
```suggestion
            $this->expectException(UnauthorizedException::class);
```

eZ/Publish/Core/Repository/LocationService.php
```diff
@@ -967,4 +942,47 @@ public function find(Filter $filter, ?array $languages = null): LocationList
...
+                            new CriterionLogicalNot($contentReadCriterion),
+                        ]
+                    ),
+                ]
+            );
+            $result = $this->repository->getSearchService()->findContent($query, [], false);
+            if ($result->totalCount > 0) {
+                throw new UnauthorizedException('content', 'read');
+            }
+        }
```
Steveb-p: Since the original condition results in an exception, this can be split into subsequent checks (instead of `elseif`).

```suggestion
        } 
        
        if ($contentReadCriterion !== true) {
            // Query if there are any content in subtree current user don't have access to
            $query = new Query(
                [
                    'limit' => 0,
                    'filter' => new CriterionLogicalAnd(
                        [
                            new CriterionSubtree($loadedSubtree->pathString),
                            new CriterionLogicalNot($contentReadCriterion),
                        ]
                    ),
                ]
            );
            $result = $this->repository->getSearchService()->findContent($query, [], false);
            if ($result->totalCount > 0) {
                throw new UnauthorizedException('content', 'read');
            }
        }
```

I do think that we should check against `=== null` instead, but there is no difference at the end of the day, as only 3 values are allowed (`null`, `false`, `true`), and `false` cannot be one of them due to previous check.

Steveb-p (approved):

eZ/Publish/Core/Repository/LocationService.php
barw4: Done, however, this wasn't a part of PR.

PR #306, opened 2022-05-04
IBX-2618: Optimized search indexing performance

alongosz:
| **JIRA issue**                          | [IBX-2618](https://issues.ibexa.co/browse/IBX-2618)
| **Requires**                            | ezsystems/ezplatform-solr-search-engine#235

This PR optimizes several code paths performance related to the process of indexing. The PR is split into meaningful commits representing separate optimization changes:
* Search Field Value mappers were optimized to use strict `SPI\Search\Field` getters (#311) rather than magic `__get`
* Aggregate Field Value mapper was optimized to have O(1) rather than O(N) complexity when accessing concrete mappers
* `FieldNameGenerator::getTypedName` was simplified (irrelevant micro-optimization) and magic getter was replaced with the strict one.

## Road not taken

I've spent a lot of time trying to introduce partial search index updates for AdminUI publishing, which publishes only a single translation, copying 1:1 the others (via 2nd argument of `ContentService::publishVersion($versionInfo, [$translations]`). The idea was to load only specific translation(s) of a `Content` item and perform partial update for the remaining languages. Almost all went well, but in the end, when comparing Solr index contents before and after the changes, it turned out that nested `Location` documents were missing for the documents which were partially updated (for `Solr` the `Location`-related documents are nested inside `Content` documents). I haven't found a way to reindex them without reindexing the whole Content item and there seems to be no explanation for this in the Solr doc. The issue might be revisited after we process the first batch of optimization changes.

## TODO
- [x] See if this breaks anything (CI, regression tests)
- [x] Check on Elasticsearch
- [x] Remove TMP commit before merging
- [x] Hook the solution to blackfire to see performance changes

#### Checklist:
- [x] Checked that target branch is set correctly
- [x] Asked for a review

### DIFF
--- a/eZ/Publish/Core/Search/Common/FieldValueMapper/Aggregate.php
+++ b/eZ/Publish/Core/Search/Common/FieldValueMapper/Aggregate.php
@@ -23,6 +23,13 @@ class Aggregate extends FieldValueMapper
     protected $mappers = [];
 
     /**
+     * Array of simple mappers mapping specific Field (by its FQCN).
+     *
+     * @var array<string, \eZ\Publish\Core\Search\Common\FieldValueMapper>
+     */
+    protected $simpleMappers = [];
+
+    /**
      * Construct from optional mapper array.
      *
      * @param \eZ\Publish\Core\Search\Common\FieldValueMapper[] $mappers
@@ -34,9 +41,13 @@ class Aggregate extends FieldValueMapper
         }
     }
 
-    public function addMapper(FieldValueMapper $mapper): void
+    public function addMapper(FieldValueMapper $mapper, ?string $searchTypeFQCN = null): void
     {
-        $this->mappers[] = $mapper;
+        if (null !== $searchTypeFQCN) {
+            $this->simpleMappers[$searchTypeFQCN] = $mapper;
+        } else {
+            $this->mappers[] = $mapper;
+        }
     }
 
     public function canMap(Field $field): bool
@@ -55,9 +66,20 @@ class Aggregate extends FieldValueMapper
      */
     public function map(Field $field)
     {
+        $mapper = $this->simpleMappers[get_class($field->getType())]
+            ?? $this->findMapper($field);
+
+        return $mapper->map($field);
+    }
+
+    /**
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotImplementedException
+     */
+    private function findMapper(Field $field): FieldValueMapper
+    {
         foreach ($this->mappers as $mapper) {
             if ($mapper->canMap($field)) {
-                return $mapper->map($field);
+                return $mapper;
             }
         }
 
--- a/eZ/Publish/Core/Search/Common/FieldValueMapper/Aggregate.php
+++ b/eZ/Publish/Core/Search/Common/FieldValueMapper/Aggregate.php
@@ -41,6 +41,9 @@ class Aggregate extends FieldValueMapper
         }
     }
 
+    /**
+     * @param class-string<\eZ\Publish\SPI\Search\Field>|null $searchTypeFQCN
+     */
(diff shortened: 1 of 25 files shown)


eZ/Publish/Core/Search/Common/FieldValueMapper/Aggregate.php
```diff
@@ -34,24 +41,16 @@ public function __construct(array $mappers = [])
         }
     }
 
-    /**
-     * Adds mapper.
-     *
-     * @param \eZ\Publish\Core\Search\Common\FieldValueMapper $mapper
-     */
-    public function addMapper(FieldValueMapper $mapper)
+    public function addMapper(FieldValueMapper $mapper, ?string $searchTypeFQCN = null): void
```
Steveb-p, line 47: This can be described (check that I used the right class)

```suggestion
    /**
     * @param class-string<\eZ\Publish\SPI\Search\Field>|null $searchTypeFQCN
     */
    public function addMapper(FieldValueMapper $mapper, ?string $searchTypeFQCN = null): void
```

eZ/Publish/Core/settings/search_engines/field_value_mappers.yml
```diff
@@ -2,72 +2,74 @@ services:
     ezpublish.search.common.field_value_mapper.boolean:
         class: eZ\Publish\Core\Search\Common\FieldValueMapper\BooleanMapper
         tags:
-            - {name: ezpublish.search.common.field_value_mapper}
+            - { name: ezpublish.search.common.field_value_mapper, maps: eZ\Publish\SPI\Search\FieldType\BooleanField }
```
Steveb-p, line 5: Remember when merging this up, that this tag has to be repeated for aliased classes.

Steveb-p (approved):

konradoboza (approved):

webhdx (approved):

damianz5 (approved):

ViniTou (approved):

mikadamczyk (approved):

tests/bundle/Core/DependencyInjection/Compiler/AggregateFieldValueMapperPassTest.php
```diff
@@ -0,0 +1,49 @@
...
+namespace Ibexa\Tests\Bundle\Core\DependencyInjection\Compiler;
+
+use eZ\Publish\Core\Base\Container\Compiler\Search\AggregateFieldValueMapperPass;
+use eZ\Publish\SPI\Search\FieldType\BooleanField;
+use Matthias\SymfonyDependencyInjectionTest\PhpUnit\AbstractCompilerPassTestCase;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+
+class AggregateFieldValueMapperPassTest extends AbstractCompilerPassTestCase
```
adamwojs:
```suggestion
final class AggregateFieldValueMapperPassTest extends AbstractCompilerPassTestCase
```

and in all other introduced tests classes

adamwojs (approved):

tests/bundle/Core/DependencyInjection/Compiler/AggregateFieldValueMapperPassTest.php
alongosz: Fixed: b06aea0dcd and 05d2d55ac4 for the other

micszo (approved):
QA Approved on Ibexa Commerce 3.3.22-dev.
Tested locally and on Ibexa Cloud.
Tested with Solr and Elasticsearch, follow-up for ES linked.
Issues on Cloud with Solr were resolved after upsizing.

PR #309, opened 2022-05-17
IBX-2880: Added PasswordExpiredException

mateuszbieniek:
| **JIRA issue**                          | [IBX-2880](https://issues.ibexa.co/browse/IBX-2880)

[CredentialsExpiredListener](https://github.com/ezsystems/ezplatform-admin-ui/blob/2.3/src/lib/EventListener/CredentialsExpiredListener.php#L57) expects `CredentialsExpiredException` which will never happen, as the Exception is immediately replaced with `BadCredentialsException` in  https://github.com/symfony/security-core/blob/5.4/Authentication/Provider/UserAuthenticationProvider.php#L90

This PR adds a new `PasswordExpiredException` exception which extends `CustomUserMessageAccountStatusException` which allows for it to be propagated to the top so `CredentialsExpiredListener` can act properly when it happens.

AdminUI part of PR: https://github.com/ezsystems/ezplatform-admin-ui/pull/2045

#### Checklist:

### DIFF
--- a/src/lib/MVC/Symfony/Security/Exception/PasswordExpiredException.php
+++ b/src/lib/MVC/Symfony/Security/Exception/PasswordExpiredException.php
@@ -1,0 +1,17 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace Ibexa\Core\MVC\Symfony\Security\Exception;
+
+use Symfony\Component\Security\Core\Exception\CustomUserMessageAccountStatusException;
+
+class PasswordExpiredException extends CustomUserMessageAccountStatusException
+{
+    public function __construct(string $message = '')
+    {
+        parent::__construct($message);
+    }
+}
--- a/src/lib/MVC/Symfony/Security/Exception/PasswordExpiredException.php
+++ b/src/lib/MVC/Symfony/Security/Exception/PasswordExpiredException.php
@@ -8,7 +8,7 @@ namespace Ibexa\Core\MVC\Symfony\Security\Exception;
 
 use Symfony\Component\Security\Core\Exception\CustomUserMessageAccountStatusException;
 
-class PasswordExpiredException extends CustomUserMessageAccountStatusException
+final class PasswordExpiredException extends CustomUserMessageAccountStatusException
 {
     public function __construct(string $message = 'User account has expired.')
     {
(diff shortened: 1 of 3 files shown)


Steveb-p:
To be frank I'm not convinced that we should introduce a dedicated exception for something that already has it's own exception in the framework. Developers will need to respect both after this patch is added, which adds to the complexity and knowledge required to properly handle our library.

Instead I would propose to check `Exception::getPreviousException()` type as a "fallback" measure, in the admin-ui package itself. 

We also need to ensure that we are not exposing user account existence when credentials are expired. That's the purpose of the Symfony code. If we're always redirecting to the password refresh page, we might introduce a security issue.

mateuszbieniek:
Unfortunately, we cannot use previous on last authentication exception, as for this case it is always `null` - I guess when an exception is serialized and stored in session this data is lost.
Also, there is no security issue, as the exception is thrown at the "postAuth" stage - therefore, for expired password info to be presented, first, proper login and password have to be provided.

src/lib/MVC/Symfony/Security/Exception/PasswordExpiredException.php
```diff
@@ -0,0 +1,17 @@
...
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace Ibexa\Core\MVC\Symfony\Security\Exception;
+
+use Symfony\Component\Security\Core\Exception\CustomUserMessageAccountStatusException;
+
+class PasswordExpiredException extends CustomUserMessageAccountStatusException
+{
+    public function __construct(string $message = '')
```
Steveb-p:
```suggestion
    public function __construct(string $message = 'User account has expired.')
```

src/lib/MVC/Symfony/Security/Exception/PasswordExpiredException.php
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace Ibexa\Core\MVC\Symfony\Security\Exception;
+
+use Symfony\Component\Security\Core\Exception\CustomUserMessageAccountStatusException;
+
+class PasswordExpiredException extends CustomUserMessageAccountStatusException
```
Steveb-p:
```suggestion
final class PasswordExpiredException extends CustomUserMessageAccountStatusException
```

Steveb-p (approved):
Following private discussion we concluded that this is actually the proper solution.

konradoboza (approved):
+1 apart from @Steveb-p suggestions.

damianz5 (approved):

bogusez (approved):

PR #310, opened 2022-05-20
EZP-30746: Fixed respecting hidden Content when moving subtree

vidarl:
| **JIRA issue**                          | [EZP-30746](https://issues.ibexa.co/browse/EZP-30746)

`DoctrineDatabase::moveSubtreeNodes` only consider `ezcontentobject_tree.is_hidden` column when checking if nodes are hidden or not.
Since https://github.com/ezsystems/ezpublish-kernel/pull/254], we also needs to check if locations are hidden due to visibility set on the object level (`ezcontentobject.is_hidden` column)

This is a replacement for https://github.com/ezsystems/ezpublish-kernel/pull/3142 which was classed as 2.5 went out of maintenance. The PR for 3.3 is much smaller as most of the helpers needed are already available in 1.3

#### Checklist:

### DIFF
--- a/eZ/Publish/API/Repository/Tests/LocationServiceTest.php
+++ b/eZ/Publish/API/Repository/Tests/LocationServiceTest.php
@@ -3239,6 +3239,135 @@ class LocationServiceTest extends BaseTest
     }
 
     /**
+     * Test that is_visible is set correct for children when moving a location where a child is hidden by content (not by location).
+     *
+     * @covers \eZ\Publish\API\Repository\LocationService::moveSubtree
+     */
+    public function testMoveSubtreeKeepsContentHiddenOnChildrenAndParent(): void
+    {
+        $repository = $this->getRepository();
+
+        $mediaLocationId = $this->generateId('location', 43);
+
+        $locationService = $repository->getLocationService();
+        $contentService = $repository->getContentService();
+
+        $sourceFolderContent = $this->publishContentWithParentLocation('SourceFolder', $mediaLocationId); // media/SourceFolder
+        $subFolderContent1 = $this->publishContentWithParentLocation('subFolderContent1', $sourceFolderContent->contentInfo->mainLocationId);
+        $subFolderContent2 = $this->publishContentWithParentLocation('subFolderContent2', $sourceFolderContent->contentInfo->mainLocationId);
+        $targetFolderContent = $this->publishContentWithParentLocation('targetFolder', $mediaLocationId); // media/SourceFolder
+
+        $contentService->hideContent($subFolderContent1->contentInfo);
+
+        // Move media/SourceFolder to media/TargetFolder/
+        $locationService->moveSubtree(
+            $sourceFolderContent->contentInfo->getMainLocation(),
+            $targetFolderContent->contentInfo->getMainLocation()
+        );
+
+        $movedLocation = $locationService->loadLocation($sourceFolderContent->contentInfo->mainLocationId);
+        $newParentLocation = $locationService->loadLocation($targetFolderContent->contentInfo->mainLocationId);
+
+        // Assert Moved Location remains visible ( only child is hidden )
+        $this->assertPropertiesCorrect(
+            [
+                'hidden' => false,
+                'invisible' => false,
+                'depth' => $newParentLocation->depth + 1,
+                'parentLocationId' => $newParentLocation->id,
+                'pathString' => $newParentLocation->pathString . $this->parseId('location', $movedLocation->id) . '/',
+            ],
+            $movedLocation
+        );
+        self::assertFalse($movedLocation->getContentInfo()->isHidden);
+
+        // Assert children of Moved location
+        $childrenLocations = [
+            $subFolderContent1->contentInfo->getMainLocation(),
+            $subFolderContent2->contentInfo->getMainLocation(),
+        ];
+        foreach ($childrenLocations as $childLocation) {
+            $this->assertPropertiesCorrect(
+                [
+                    'hidden' => $childLocation === $subFolderContent1->contentInfo->getMainLocation(), // Only SubFolderContent1 should be hidden,
+                    'invisible' => $childLocation === $subFolderContent1->contentInfo->getMainLocation(), // Only SubFolderContent1 should be hidden
+                    'depth' => $movedLocation->depth + 1,
+                    'parentLocationId' => $movedLocation->id,
+                    'pathString' => $movedLocation->pathString . $this->parseId('location', $childLocation->id) . '/',
+                ],
+                $childLocation
+            );
+            self::assertEquals($childLocation === $subFolderContent1->contentInfo->getMainLocation(), $childLocation->getContentInfo()->isHidden);
+        }
+    }
+
+    /**
+     * Test that is_visible is set correct for children when moving a content which is hidden (location is not hidden).
+     *
+     * @covers \eZ\Publish\API\Repository\LocationService::moveSubtree
+     */
+    public function testMoveSubtreeKeepsContentHiddenOnChildren(): void
+    {
+        $repository = $this->getRepository();
+
+        $sourceLocationId = $this->createFolder(
+            [
+                'eng-GB' => 'SourceParentFolder',
+            ],
+            2
+        )->getVersionInfo()->getContentInfo()->mainLocationId;
+
+        $locationService = $repository->getLocationService();
+        $contentService = $repository->getContentService();
+
+        $sourceFolderContent = $this->publishContentWithParentLocation('SourceFolder', $sourceLocationId); // media/SourceFolder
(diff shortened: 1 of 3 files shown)


eZ/Publish/API/Repository/Tests/LocationServiceTest.php
```diff
@@ -3238,6 +3238,126 @@ public function testMoveSubtreeUpdatesPathIdentificationString(): void
         );
     }
 
+    /**
+     * Test that is_visible is set correct for children when moving a location where a child is hidden by content (not by location).
+     *
+     * @covers \eZ\Publish\API\Repository\LocationService::moveSubtree
+     * depends eZ\Publish\API\Repository\Tests\LocationServiceTest::testLoadLocation
```
alongosz: This line does not do anything in this form and can be removed.

eZ/Publish/API/Repository/Tests/LocationServiceTest.php
```diff
@@ -3238,6 +3238,126 @@ public function testMoveSubtreeUpdatesPathIdentificationString(): void
...
+            );
+            self::assertEquals($childLocation === $subFolderContent1->contentInfo->getMainLocation(), $childLocation->getContentInfo()->isHidden);
+        }
+    }
+
+    /**
+     * Test that is_visible is set correct for children when moving a content which is hidden (location is not hidden).
+     *
+     * @covers \eZ\Publish\API\Repository\LocationService::moveSubtree
+     * depends eZ\Publish\API\Repository\Tests\LocationServiceTest::testLoadLocation
```
alongosz: This one too

eZ/Publish/API/Repository/Tests/LocationServiceTest.php
```diff
@@ -3238,6 +3238,126 @@ public function testMoveSubtreeUpdatesPathIdentificationString(): void
...
+                'depth' => $newParentLocation->depth + 1,
+                'parentLocationId' => $newParentLocation->id,
+                'pathString' => $newParentLocation->pathString . $this->parseId('location', $movedLocation->id) . '/',
+            ],
+            $movedLocation
+        );
+        self::assertFalse($movedLocation->getContentInfo()->isHidden);
+
+        // Assert children of Moved location
+        foreach ([$subFolderContent1->contentInfo->getMainLocation(), $subFolderContent2->contentInfo->getMainLocation()] as $childLocation) {
```
alongosz: CS please extract the array to a separate variable for a readability

eZ/Publish/API/Repository/Tests/LocationServiceTest.php
```diff
@@ -3238,6 +3238,126 @@ public function testMoveSubtreeUpdatesPathIdentificationString(): void
...
+     * Test that is_visible is set correct for children when moving a content which is hidden (location is not hidden).
+     *
+     * @covers \eZ\Publish\API\Repository\LocationService::moveSubtree
+     * depends eZ\Publish\API\Repository\Tests\LocationServiceTest::testLoadLocation
+     */
+    public function testMoveSubtreeKeepsContentHiddenOnChildren(): void
+    {
+        $repository = $this->getRepository();
+
+        $mediaLocationId = $this->generateId('location', 43);
```
alongosz: Please create atomic test case instead of relying on existing legacy data with hardcoded database ids.

eZ/Publish/Core/Persistence/Legacy/Content/Location/Gateway/DoctrineDatabase.php
```diff
@@ -338,28 +339,33 @@ public function moveSubtreeNodes(array $sourceNodeData, array $destinationNodeDa
         }
     }
 
-    private function getHiddenNodeIds(array $rows): array
+    private function getHiddenNodeIds(int $contentObjectId)
```
alongosz:
```suggestion
    /*
     * @return int[]
     */
    private function getHiddenNodeIds(int $contentObjectId): array
```

alongosz (changes requested):

eZ/Publish/API/Repository/Tests/LocationServiceTest.php
vidarl: @alongosz 
All other tests here uses the same approach, but ok... Instead of relying on existing Media folder with location `43`, I can just create a new folder. However, in order to do that I'll need to hardcode either `1` or `2` as parent. Is that OK?
The only alternative I found would be to do `LocationService::loadAllLocations(0,1)`, but is that any better ?

eZ/Publish/API/Repository/Tests/LocationServiceTest.php
vidarl: Most of the other tests has dependencies to other tests. I assume it is in order to avoid a lot of false positives if some base test starts failing. In the case here I replaced `@depends` with `depends` in order to run the this test alone using filters while developing the test itself. I then forgot to re-add the `@` before commit.
I have anyway now removed the line in a99fb158a9447ae40c7638048cd981a209de12d1

eZ/Publish/API/Repository/Tests/LocationServiceTest.php
vidarl: Removed in a99fb158a9447ae40c7638048cd981a209de12d1

eZ/Publish/API/Repository/Tests/LocationServiceTest.php
vidarl: Fixed in I have anyway now removed the line in a99fb158a9447ae40c7638048cd981a209de12d1

eZ/Publish/API/Repository/Tests/LocationServiceTest.php
alongosz: > All other tests here uses the same approach, but ok... Instead of relying on existing Media folder with location `43`, I can just create a new folder. However, in order to do that I'll need to hardcode either `1` or `2` as parent. Is that OK? The only alternative I found would be to do `LocationService::loadAllLocations(0,1)`, but is that any better ?

It's a technical debt. We can hard-code for now Content root Location `2` because it's going to be easy to find & change once we switch to proper fixture data creation for tests. TBH that `2` probably should be a const for all tests, but that's out of scope here.

So indeed what we needed was
```php
       $sourceLocationId = $this->createFolder(
            [
                'eng-GB' => 'SourceParentFolder',
            ],
            2
        )->getVersionInfo()->getContentInfo()->mainLocationId;
```



**I'm a bit late with the response, so to speed up the process I've allowed myself to fix this and small minor things I found and rebase the PR. Let's see if CS passes.**

alongosz (approved):

eZ/Publish/API/Repository/Tests/LocationServiceTest.php
```diff
@@ -3238,6 +3238,135 @@ public function testMoveSubtreeUpdatesPathIdentificationString(): void
...
+
+        $locationService = $repository->getLocationService();
+        $contentService = $repository->getContentService();
+
+        $sourceFolderContent = $this->publishContentWithParentLocation('SourceFolder', $mediaLocationId); // media/SourceFolder
+        $subFolderContent1 = $this->publishContentWithParentLocation('subFolderContent1', $sourceFolderContent->contentInfo->mainLocationId);
+        $subFolderContent2 = $this->publishContentWithParentLocation('subFolderContent2', $sourceFolderContent->contentInfo->mainLocationId);
+        $targetFolderContent = $this->publishContentWithParentLocation('targetFolder', $mediaLocationId); // media/SourceFolder
```
vidarl:
```suggestion
        $targetFolderContent = $this->publishContentWithParentLocation('targetFolder', $mediaLocationId); // media/TargetFolder
```

eZ/Publish/API/Repository/Tests/LocationServiceTest.php
```diff
@@ -3238,6 +3238,135 @@ public function testMoveSubtreeUpdatesPathIdentificationString(): void
...
+            2
+        )->getVersionInfo()->getContentInfo()->mainLocationId;
+
+        $locationService = $repository->getLocationService();
+        $contentService = $repository->getContentService();
+
+        $sourceFolderContent = $this->publishContentWithParentLocation('SourceFolder', $sourceLocationId); // media/SourceFolder
+        $subFolderContent1 = $this->publishContentWithParentLocation('subFolderContent1', $sourceFolderContent->contentInfo->mainLocationId);
+        $subFolderContent2 = $this->publishContentWithParentLocation('subFolderContent2', $sourceFolderContent->contentInfo->mainLocationId);
+        $targetFolderContent = $this->publishContentWithParentLocation('targetFolder', $sourceLocationId); // media/SourceFolder
```
vidarl:
```suggestion
        $targetFolderContent = $this->publishContentWithParentLocation('targetFolder', $sourceLocationId); // media/TargetFolder
```

adamwojs (approved):

bogusez (approved):

PR #317, opened 2022-07-01
IBX-1119: Implemented Site Access listener for IO requests

barw4:
| **JIRA issue**                          | [IBX-1119](https://issues.ibexa.co/browse/IBX-1119)

Injecting a SiteAccess allows for proper resolution of a repository, and hence, a `var_dir`.

#### Checklist:
- [x]  Asked for a review (ping `@ezsystems/engineering-team`).

### DIFF
--- a/eZ/Bundle/EzPublishIOBundle/Resources/config/io.yml
+++ b/eZ/Bundle/EzPublishIOBundle/Resources/config/io.yml
@@ -157,3 +157,12 @@ services:
             - '@eZ\Publish\Core\IO\IOConfigProvider'
         tags:
             - { name: kernel.event_subscriber }
+
+    Ibexa\Bundle\IO\EventListener\StreamFileSiteAccessListener:
+        arguments:
+            - '@ezpublish.config.resolver'
+            - '@ezpublish.siteaccess_router'
+            - '@ezpublish.siteaccess_service'
+            - '%ezpublish.siteaccess.list%'
+        tags:
+            - { name: kernel.event_subscriber }
--- a/src/bundle/IO/EventListener/StreamFileSiteAccessListener.php
+++ b/src/bundle/IO/EventListener/StreamFileSiteAccessListener.php
@@ -1,0 +1,68 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace Ibexa\Bundle\IO\EventListener;
+
+use eZ\Publish\Core\MVC\ConfigResolverInterface;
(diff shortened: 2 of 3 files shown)


eZ/Bundle/EzPublishIOBundle/Resources/config/io.yml
```diff
@@ -157,3 +157,12 @@ services:
             - '@eZ\Publish\Core\IO\IOConfigProvider'
         tags:
             - { name: kernel.event_subscriber }
+
+    Ibexa\Bundle\IO\EventListener\StreamFileSiteAccessListener:
+        arguments:
+            - '@ezpublish.config.resolver'
+            - '@ezpublish.siteaccess_router'
+            - '@ezpublish.siteaccess_service'
+            - '%ezpublish.siteaccess.list%'
```
alongosz, line 166: It would be better here to use named parameters. More flexible.

src/bundle/IO/EventListener/StreamFileSiteAccessListener.php
```diff
@@ -0,0 +1,68 @@
...
+
+    public function onKernelRequest(RequestEvent $event): void
+    {
+        if ($event->getRequestType() !== HttpKernelInterface::MAIN_REQUEST) {
+            return;
+        }
+
+        $pathInfo = $event->getRequest()->getPathInfo();
+
+        foreach ($this->siteAccessList as $siteAccess) {
```
alongosz, line 60: Full SiteAccess list is provided by the SiteAccessService (`SiteAccessServiceInterface::getAll`)

alongosz (changes requested):

adamwojs (changes requested):
Sorry, but this is wrong direction. 

Executing proposed logic for all available site accesses in every request (there might be hundreds of them since we have site factory) will kill application performance.

barw4:
> Sorry, but this is wrong direction.
> 
> Executing proposed logic for all available site accesses in every request (there might be hundreds of them since we have site factory) will kill application performance.

@adamwojs Sure, I agree, however, this seems like an edge case so would that really kill the performance that much? Please let me know how else can we approach this with having only the `localhost/var/repository2/storage/images/image.png` URL.

eZ/Bundle/EzPublishIOBundle/Resources/config/io.yml
Steveb-p: For our own services this is true. Do keep in mind that external services can have their argument names changed, so it's valid to use argument order there (just a side note).

alongosz (changes requested):
@barw4 given the JIRA description
> When adding manualy the current siteaccess to the request (or a siteaccess using this second repo), the image is correctly served: http://example.com/admin2/var/repo2/storage/images/…

We would rather like you to investigate why generated URI does not include `admin2` component and what can be done to change that. Additionally what would be the BC impact of such change and how can we ensure BC (we were thinking about SA Matcher).

PR #320, opened 2022-07-06
IBX-3036: Provided fallback for Content-Disposition header in `DownloadController`

barw4:
| **JIRA issue**                          | [IBX-3036](https://issues.ibexa.co/browse/IBX-3036)

This change would still allow downloading file with the original filename.

#### Checklist:

### DIFF
--- a/eZ/Publish/Core/MVC/Symfony/Controller/Content/DownloadController.php
+++ b/eZ/Publish/Core/MVC/Symfony/Controller/Content/DownloadController.php
@@ -75,7 +75,11 @@ class DownloadController extends Controller
         }
 
         $response = new BinaryStreamResponse($this->ioService->loadBinaryFile($field->value->id), $this->ioService);
-        $response->setContentDisposition(ResponseHeaderBag::DISPOSITION_ATTACHMENT, $filename, md5($filename));
+        $response->setContentDisposition(
+            ResponseHeaderBag::DISPOSITION_ATTACHMENT,
+            $filename,
+            bin2hex(random_bytes(8))
+        );
 
         return $response;
     }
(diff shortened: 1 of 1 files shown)


eZ/Publish/Core/MVC/Symfony/Controller/Content/DownloadController.php
```diff
@@ -75,7 +75,7 @@ public function downloadBinaryFileAction($contentId, $fieldIdentifier, $filename
         }
 
         $response = new BinaryStreamResponse($this->ioService->loadBinaryFile($field->value->id), $this->ioService);
-        $response->setContentDisposition(ResponseHeaderBag::DISPOSITION_ATTACHMENT, $filename);
+        $response->setContentDisposition(ResponseHeaderBag::DISPOSITION_ATTACHMENT, $filename, md5($filename));
```
alongosz: SonarLint reports here a security issue. It's false negative, but to avoid any discussion / doubts is it technically possible here to use some better algorithm?

alongosz (changes requested):

eZ/Publish/Core/MVC/Symfony/Controller/Content/DownloadController.php
barw4: We can use anything here really, but technically this doesn't have to be secure. It's rather a matter of encoding.

alongosz: > We can use anything here really, but technically this doesn't have to be secure. It's rather a matter of encoding.

Ok, then please avoid `md5`. It always takes a second to figure out if it is security or not, so better use something that does not raise that question :)

alongosz (approved):

konradoboza (approved):

Steveb-p (approved):

glye (approved):

mnocon (approved):
I've reproduced the issue and verified that it no longer occurs when the fix from this PR is applied.

Approved!

PR #321, opened 2022-07-07
IBX-2634: Added argument to functions to ignore ownership

mateuszdebinski:
| **JIRA issue**                          | [IBX-2634](https://issues.ibexa.co/browse/IBX-2634)
| **Type**                                   | bug/improvement

In order not to change the interface and not to cause BC, only the functions have been changed so that you can pass the argument to ignore the owner - the mechanism itself already existed in our product

#### Checklist:

### DIFF
--- a/eZ/Publish/API/Repository/Tests/BaseContentTypeServiceTest.php
+++ b/eZ/Publish/API/Repository/Tests/BaseContentTypeServiceTest.php
@@ -7,6 +7,7 @@
 namespace eZ\Publish\API\Repository\Tests;
 
 use eZ\Publish\API\Repository\Values\Content\Location;
+use eZ\Publish\API\Repository\Values\ContentType\ContentTypeDraft;
 
 /**
  * Base class for content type specific tests.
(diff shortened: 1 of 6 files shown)


alongosz (changes requested):
AFAIR any change needs to be reflected on an API interface, because we use ProxyManager which won't see them otherwise. But... @adamwojs could you validate that statement? ;)

Question: what is the use case?

mateuszdebinski:
@alongosz 

If I remember correctly, the API interface change won't go through because of BC.

In migration, we allow setting creatorId or modifierId, if a user does not have the appropriate permissions then we throw an exception. However, I probably went a bit too far, because as I read the ticket now, the client would like that if the user does not exist or does not have permission to create/edit, should be used, default user in migration (admin) or  from --user-context option

alongosz (changes requested):
It feels to me that this should be fixed in ibexa/migrations.
POV ping @Steveb-p

mateuszdebinski:
I close this PR as we can do it in the migration bundle itself

PR #324, opened 2022-07-26
IBX-3451: [Behat] Removed depending on APP_ENV when loading Context services

mnocon:
JIRA: https://issues.ibexa.co/browse/IBX-3451 (Customer request)

Backport of a change introduced in https://github.com/ibexa/core/pull/92 (https://github.com/ibexa/core/pull/92/commits/073a5cce403fc060b803d7f98b4df253013ffc25) to v3.3 because of the ticket mentioned above.
Relying on a parameter (prepended by ibexa/behat package) is better than relying on hardcoded Symfony env (and does not require adding optional dependencies to bypass it).

I've also added small tests, because it's a Customer Request.

Related PR: https://github.com/ezsystems/ezplatform-content-forms/pull/62

### DIFF
--- a/eZ/Bundle/EzPublishCoreBundle/DependencyInjection/EzPublishCoreExtension.php
+++ b/eZ/Bundle/EzPublishCoreBundle/DependencyInjection/EzPublishCoreExtension.php
@@ -91,8 +91,7 @@ class EzPublishCoreExtension extends Extension implements PrependExtensionInterf
 
         $configuration = $this->getConfiguration($configs, $container);
 
-        $environment = $container->getParameter('kernel.environment');
-        if (in_array($environment, ['behat', 'test'])) {
+        if ($this->shouldLoadTestServices($container)) {
             $loader->load('feature_contexts.yml');
         }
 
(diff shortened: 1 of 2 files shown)


konradoboza (approved):

adamwojs (approved):

micszo (approved):

eZ/Bundle/EzPublishCoreBundle/DependencyInjection/EzPublishCoreExtension.php
```diff
@@ -91,8 +91,7 @@ public function load(array $configs, ContainerBuilder $container)
 
         $configuration = $this->getConfiguration($configs, $container);
 
-        $environment = $container->getParameter('kernel.environment');
-        if (in_array($environment, ['behat', 'test'])) {
+        if ($this->shouldLoadTestServices($container)) {
```
alongosz, line 94: Now that I see it, I think it should be a separate bundle, not a part of Core ;)

alongosz (approved):

bogusez (approved):

bogusez (approved):

PR #329, opened 2022-08-11
IBX-3588: Enhanced `ContentFieldValidation` exception

barw4:
| **JIRA issue**                          | [IBX-3588](https://issues.ibexa.co/browse/IBX-3588)

`ContentFieldValidation` exception message to include validation messages for each field.

Related PR: https://github.com/ezsystems/ezplatform-admin-ui/pull/2060

Example:
image

SonarCloud duplication analysis is not related to this PR.

#### Checklist:
- [x]  Checked that target branch is set correctly (master for features, the oldest supported for bugs).

### DIFF
--- a/eZ/Publish/Core/Base/Exceptions/ContentFieldValidationException.php
+++ b/eZ/Publish/Core/Base/Exceptions/ContentFieldValidationException.php
@@ -32,7 +32,7 @@ class ContentFieldValidationException extends APIContentFieldValidationException
     protected $errors;
 
     /** @var string|null */
-    protected $target;
+    protected $contentName;
 
     /**
      * Generates: Content fields did not validate.
@@ -41,15 +41,15 @@ class ContentFieldValidationException extends APIContentFieldValidationException
      *
      * @param array<array-key, array<string, \eZ\Publish\Core\FieldType\ValidationError>> $errors
      */
-    public function __construct(array $errors, ?string $target = null)
+    public function __construct(array $errors, ?string $contentName = null)
     {
         $this->errors = $errors;
-        $this->target = $target;
+        $this->contentName = $contentName;
 
-        $this->setMessageTemplate('Content Fields %contentInfo%did not validate: %errors%');
+        $this->setMessageTemplate('Content Fields %contentName%did not validate: %errors%');
         $this->setParameters([
             '%errors%' => $this->generateValidationErrorsMessages(),
-            '%contentInfo%' => $this->target !== null ? sprintf('of Content "%s" ', $this->target) : '',
+            '%contentName%' => $this->contentName !== null ? sprintf('of Content "%s" ', $this->contentName) : '',
         ]);
 
         parent::__construct($this->getBaseTranslation());
@@ -67,24 +67,29 @@ class ContentFieldValidationException extends APIContentFieldValidationException
 
     private function generateValidationErrorsMessages(): string
     {
-        $message = '';
+        $validationErrors = $this->collectValidationErrors();
+
+        return "\n- " . implode("\n- ", $validationErrors);
+    }
+
+    /**
+     * @return array<\eZ\Publish\Core\FieldType\ValidationError>
+     */
+    private function collectValidationErrors(): array
+    {
+        $messages = [];
         foreach ($this->getFieldErrors() as $validationErrors) {
             foreach ($validationErrors as $validationError) {
                 if (is_array($validationError)) {
                     foreach ($validationError as $item) {
-                        $message .= $this->generateValidationErrorMessage($item);
+                        $messages[] = $item->getTranslatableMessage();
                     }
                 } else {
-                    $message .= $this->generateValidationErrorMessage($validationError);
+                    $messages[] = $validationError->getTranslatableMessage();
                 }
             }
         }
 
-        return $message;
-    }
-
-    private function generateValidationErrorMessage(ValidationError $validationError): string
-    {
-        return sprintf("\n- %s", $validationError->getTranslatableMessage());
+        return $messages;
     }
 }
--- a/eZ/Publish/Core/Base/Exceptions/ContentFieldValidationException.php
+++ b/eZ/Publish/Core/Base/Exceptions/ContentFieldValidationException.php
@@ -9,7 +9,6 @@ namespace eZ\Publish\Core\Base\Exceptions;
 use eZ\Publish\API\Repository\Exceptions\ContentFieldValidationException as APIContentFieldValidationException;
 use eZ\Publish\Core\Base\Translatable;
 use eZ\Publish\Core\Base\TranslatableBase;
-use eZ\Publish\Core\FieldType\ValidationError;
 
 /**
  * This Exception is thrown on create or update content one or more given fields are not valid.
@@ -17,6 +16,8 @@ use eZ\Publish\Core\FieldType\ValidationError;
 class ContentFieldValidationException extends APIContentFieldValidationException implements Translatable
 {
     use TranslatableBase;
+
+    private const MAX_MESSAGES_NUMBER = 32;
 
     /**
      * Contains an array of field ValidationError objects indexed with FieldDefinition id and language code.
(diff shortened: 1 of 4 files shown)


eZ/Publish/Core/Base/Exceptions/ContentFieldValidationException.php
```diff
@@ -30,17 +31,27 @@ class ContentFieldValidationException extends APIContentFieldValidationException
...
+    public function __construct(array $errors, ?string $target = null)
     {
         $this->errors = $errors;
-        $this->setMessageTemplate('Content Fields did not validate');
+        $this->target = $target;
+
+        $this->setMessageTemplate('Content Fields %contentInfo%did not validate: %errors%');
+        $this->setParameters([
+            '%errors%' => $this->generateValidationErrorsMessages(),
+            '%contentInfo%' => $this->target !== null ? sprintf('of Content "%s" ', $this->target) : '',
```
alongosz: Naming:
I don't follow `contentInfo` naming. What is this target really, since it's `?string` by strict definition? If I see `contentInfo` I expect there an instance of API ContentInfo.

Update: after reading further you rather want here `?string $contentName` and '%contentName%' respectively.

eZ/Publish/Core/Base/Exceptions/ContentFieldValidationException.php
```diff
@@ -53,4 +64,27 @@ public function getFieldErrors()
...
+            }
+        }
+
+        return $message;
+    }
+
+    private function generateValidationErrorMessage(ValidationError $validationError): string
+    {
+        return sprintf("\n- %s", $validationError->getTranslatableMessage());
+    }
```
alongosz: This should rather be an array which gets to be `implode` after collecting. Seems performance wise better.

eZ/Publish/Core/Base/Exceptions/ContentFieldValidationException.php
```diff
@@ -53,4 +64,27 @@ public function getFieldErrors()
...
+        $message = '';
+        foreach ($this->getFieldErrors() as $validationErrors) {
+            foreach ($validationErrors as $validationError) {
+                if (is_array($validationError)) {
+                    foreach ($validationError as $item) {
+                        $message .= $this->generateValidationErrorMessage($item);
+                    }
+                } else {
+                    $message .= $this->generateValidationErrorMessage($validationError);
+                }
```
alongosz: What is the use case for this structure to be object instead of sub-array? AFAIK this is always map of Field Definition Identifier to Language Code (or the other way around) to the error message.

alongosz (changes requested):
This is much needed improvement! Thanks! Let's rectify some issues before we ship it.
Is there a way to trigger that error instead of pre-checks done by AdminUI so QA can test it?

Sonar duplicates report is relevant to the changes in `eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php`. Please use PHPStorm to identify refactor duplication in that class.

Other remarks:

eZ/Publish/Core/Base/Exceptions/ContentFieldValidationException.php
```diff
@@ -30,17 +31,27 @@ class ContentFieldValidationException extends APIContentFieldValidationException
...
+
     /**
      * Generates: Content fields did not validate.
      *
      * Also sets the given $fieldErrors to the internal property, retrievable by getFieldErrors()
      *
      * @param array<array-key, array<string, \eZ\Publish\Core\FieldType\ValidationError>> $errors
      */
-    public function __construct(array $errors)
+    public function __construct(array $errors, ?string $target = null)
```
alongosz: Thinking about this further, maybe here should be rather `?ContentInfo` or `?Content` (depending on what fits better in the context)? It's a value and this exception is related to Content, so I don't see mixed layers issue here. The reason here is that I'd rather expect `[<content_id>] <content_name>` pattern in error message, as the content name by itself might be to ambiguous.

barw4: Tbh `content_id` gives not much of a value. For example, when we are copying a subtree or content it will just be a meaningless number as it would be just an ID of a newly created Content ID. What is more, this could be misleading if someone wants to find the guilty content.

But if you still believe that we should place it in there, I'm fine with it.

barw4:
> Is there a way to trigger that error instead of pre-checks done by AdminUI so QA can test it?

@alongosz probably the easiest way would be to use Public PHP API.

> Sonar duplicates report is relevant to the changes in eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php. Please use PHPStorm to identify refactor duplication in that class.

Yes, but I haven't touched lines that are duplicated. SonarCloud validated the whole file.

eZ/Publish/Core/Base/Exceptions/ContentFieldValidationException.php
barw4: This can be either an object or an array, depending on the number of validation errors for a single field identifier.

alongosz:
>> Sonar duplicates report is relevant to the changes in eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php. Please use PHPStorm to identify refactor duplication in that class.
>
> Yes, but I haven't touched lines that are duplicated. SonarCloud validated the whole file.

@barw4 the thing is I cannot have false negative CI status when merging, because that leads to maintenance issues. You have 2 choices here. Either be proactive and fix it yourself (doesn't matter if in this PR or separate, since it's relevant for the updated test case) or wait for one of the Maintainers to fix it.

barw4:
@alongosz duplications were removed.

alongosz (approved):

eZ/Publish/Core/Base/Exceptions/ContentFieldValidationException.php
```diff
@@ -53,4 +64,32 @@ public function getFieldErrors()
     {
         return $this->errors;
     }
+
+    private function generateValidationErrorsMessages(): string
+    {
+        $validationErrors = $this->collectValidationErrorsMessages();
```
adamwojs: IMHO we should limit errors included into message to some reasonable number e.g. 32 and add text ellipsis if this limit is exceeded.

eZ/Publish/Core/Base/Exceptions/ContentFieldValidationException.php
```diff
@@ -30,17 +31,27 @@ class ContentFieldValidationException extends APIContentFieldValidationException
...
+
     /**
      * Generates: Content fields did not validate.
      *
      * Also sets the given $fieldErrors to the internal property, retrievable by getFieldErrors()
      *
      * @param array<array-key, array<string, \eZ\Publish\Core\FieldType\ValidationError>> $errors
      */
-    public function __construct(array $errors)
+    public function __construct(array $errors, ?string $contentName = null)
```
adamwojs: Probably static factory method might be better choice to express your intention: creating ContentFieldValidationException for content update scenario

adamwojs (approved):

eZ/Publish/Core/Base/Exceptions/ContentFieldValidationException.php
webhdx: I second that. A new static method to create a new instance with multiline message would be a cleaner approach.

eZ/Publish/Core/Base/Exceptions/ContentFieldValidationException.php
```diff
@@ -30,17 +31,27 @@ class ContentFieldValidationException extends APIContentFieldValidationException
...
     {
         $this->errors = $errors;
-        $this->setMessageTemplate('Content Fields did not validate');
+        $this->contentName = $contentName;
+
+        $this->setMessageTemplate('Content Fields %contentName%did not validate: %errors%');
```
webhdx:
```suggestion
        $this->setMessageTemplate('Content "%contentName%" fields did not validate: %errors%');
```

webhdx (changes requested):

eZ/Publish/Core/Base/Exceptions/ContentFieldValidationException.php
barw4: Actually, a single space has to be there in order to display message properly in case `%contentName%` is empty.

eZ/Publish/Core/Base/Exceptions/ContentFieldValidationException.php
```diff
@@ -30,20 +32,35 @@ class ContentFieldValidationException extends APIContentFieldValidationException
...
+            '%errors%' => $this->generateValidationErrorsMessages(),
+            '%contentName%' => $this->contentName !== null ? sprintf('"%s" ', $this->contentName) : '',
+        ]);
+
         parent::__construct($this->getBaseTranslation());
     }
 
+    public static function createNewWithMultiline(array $errors, ?string $contentName = null): self
+    {
+        return new self($errors, $contentName);
```
webhdx: It doesn't provide any value as it is. You need to move whole logic from the constructor to `createNewWithMultiline` and leave the constructor as it was prior to this PR.

webhdx (changes requested):

webhdx (approved):

micszo (approved):
QA Approved on Ibexa Commerce 3.3.24.

mikadamczyk (approved):

adamwojs:
@barw4 Could you please merge up changes?

PR #330, opened 2022-08-11
[WiP] IBX-3636: Implemented multilingual archiving of URL aliases

alongosz:
| **JIRA issue**                          | [IBX-3636](https://issues.ibexa.co/browse/IBX-3636)

### Steps to reproduce 

1. Given I have multiple languages on my Admin UI (lets say english and french, where english is the default language).
2. I create a new content type (lets name it "Test type") where :
   1. I have only one translatable text line, that's used as the content name.
   2. (important) I uncheck the default availability checkbox, so that all my future content items won't be always available by default.
3. In the content tree, I create and publish one content of type "Test type" in english and name it "foo".
4. I translate this content with the same name in french.
   1. At this point, "/foo" on the default english siteaccess gives me the new content, as expected.
5. I then edit the english version of my content and change the name to "foobar".

### Expected behaviour

I expect "/foo" to redirect me to "/foobar", or at least display the full view of my "foobar" content item.

### Actual behaviour

The content is not found, and a 404 error is thrown. "/foobar" is working though, and displays the newly updated content item.

### Extended primary key 

To solve the issue we can extend PK on `ezurlalias_ml` table from `(parent, text_md5)` to `(parent, text_md5, lang_mask)`. However the consequence will be:

1. Content items without translation in specific language and content type without always available flag are no longer available in Site access until specific language is not included into fallback languages.   

2. All url alias segments needs to have translation in given language or always available flag to successfully match a route.

3. Increased number of `ezurlalias_ml` table/persistence cache entries. Additional `n - 1` records, where `n` is number of languages in content item version 

4. Increased number of url alias lookup calls. Additional `n - 1` url alias lookups, where `n` is number of prioritized language    
 

#### Checklist:
- [ ] Adapt persistence cache layer 
- [ ] Checked that target branch is set correctly (master for features, the oldest supported for bugs).


alongosz:
The `ezurlalias_ml` relation will be most likely refactored in Ibexa 5.0 to rely on `language_id` due to the upcoming 63+ languages support  feature, so this solution is obsolete ATM.

PR #333, opened 2022-09-01
IBX-3445: Softened `TargetContentValidator` content check

barw4:
| **JIRA issue**                          | [IBX-3445](https://issues.ibexa.co/browse/IBX-3445)

Mentioned PR will allow to update content that has relation not available to a current user. 

#### Checklist:
- [x]  Ran PHP CS Fixer for new PHP code (use `$ composer fix-cs`).

### DIFF
--- a/src/lib/Repository/Validator/TargetContentValidator.php
+++ b/src/lib/Repository/Validator/TargetContentValidator.php
@@ -8,10 +8,8 @@ declare(strict_types=1);
 
 namespace Ibexa\Core\Repository\Validator;
 
-use eZ\Publish\API\Repository\ContentService;
-use eZ\Publish\API\Repository\ContentTypeService;
 use eZ\Publish\API\Repository\Exceptions\NotFoundException;
-use eZ\Publish\API\Repository\Exceptions\UnauthorizedException;
+use eZ\Publish\API\Repository\Repository;
 use eZ\Publish\Core\FieldType\ValidationError;
 
 /**
@@ -21,25 +19,21 @@ use eZ\Publish\Core\FieldType\ValidationError;
  */
 final class TargetContentValidator implements TargetContentValidatorInterface
 {
-    /** @var \eZ\Publish\API\Repository\ContentService */
-    private $contentService;
+    /** @var \eZ\Publish\API\Repository\Repository */
+    private $repository;
 
-    /** @var \eZ\Publish\API\Repository\ContentTypeService */
-    private $contentTypeService;
-
-    public function __construct(
-        ContentService $contentService,
-        ContentTypeService $contentTypeService
-    ) {
-        $this->contentService = $contentService;
-        $this->contentTypeService = $contentTypeService;
+    public function __construct(Repository $repository)
+    {
+        $this->repository = $repository;
     }
 
     public function validate(int $value, array $allowedContentTypes = []): ?ValidationError
     {
         try {
-            $contentInfo = $this->contentService->loadContentInfo($value);
-            $contentType = $this->contentTypeService->loadContentType($contentInfo->contentTypeId);
+            $contentInfo = $this->repository->sudo(static function (Repository $repository) use ($value) {
+                return $repository->getContentService()->loadContentInfo($value);
+            });
+            $contentType = $this->repository->getContentTypeService()->loadContentType($contentInfo->contentTypeId);
 
             if (!empty($allowedContentTypes) && !in_array($contentType->identifier, $allowedContentTypes, true)) {
                 return new ValidationError(
@@ -51,13 +45,11 @@ final class TargetContentValidator implements TargetContentValidatorInterface
                     'targetContentId'
                 );
             }
-        } catch (NotFoundException | UnauthorizedException $e) {
+        } catch (NotFoundException $e) {
             return new ValidationError(
-                'Content with identifier %contentId% is not a valid relation target',
+                'Content is not a valid relation target',
                 null,
-                [
-                    '%contentId%' => $value,
-                ],
+                [],
                 'targetContentId'
             );
         }
--- a/src/lib/Repository/Validator/TargetContentValidator.php
+++ b/src/lib/Repository/Validator/TargetContentValidator.php
@@ -9,8 +9,8 @@ declare(strict_types=1);
 namespace Ibexa\Core\Repository\Validator;
 
 use eZ\Publish\API\Repository\Exceptions\NotFoundException;
-use eZ\Publish\API\Repository\Repository;
 use eZ\Publish\Core\FieldType\ValidationError;
+use eZ\Publish\SPI\Persistence\Content;
 
 /**
  * Validator for checking existence of content and its content type.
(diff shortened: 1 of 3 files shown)


src/lib/Repository/Validator/TargetContentValidator.php
```diff
@@ -51,13 +45,11 @@ public function validate(int $value, array $allowedContentTypes = []): ?Validati
                     'targetContentId'
                 );
             }
-        } catch (NotFoundException | UnauthorizedException $e) {
+        } catch (NotFoundException $e) {
             return new ValidationError(
-                'Content with identifier %contentId% is not a valid relation target',
```
adamwojs, line 56: What is the reason behind this change ?

src/lib/Repository/Validator/TargetContentValidator.php
```diff
@@ -21,25 +19,21 @@
...
 
     public function validate(int $value, array $allowedContentTypes = []): ?ValidationError
     {
         try {
-            $contentInfo = $this->contentService->loadContentInfo($value);
-            $contentType = $this->contentTypeService->loadContentType($contentInfo->contentTypeId);
+            $contentInfo = $this->repository->sudo(static function (Repository $repository) use ($value) {
+                return $repository->getContentService()->loadContentInfo($value);
+            });
+            $contentType = $this->repository->getContentTypeService()->loadContentType($contentInfo->contentTypeId);
```
adamwojs: 1) return type is missing in callback 
2) Please consider using \eZ\Publish\SPI\Persistence\Content\Handler::loadContentInfo and \eZ\Publish\SPI\Persistence\Content\Type\Handler::load instead of `sudo`

adamwojs (commented):
Related discussion: https://github.com/ezsystems/ezpublish-kernel/pull/3131#discussion_r858570959

src/lib/Repository/Validator/TargetContentValidator.php
barw4: That was @glye idea, please see the internal comment on the JIRA ticket.

src/lib/Repository/Validator/TargetContentValidator.php
```diff
@@ -21,25 +19,21 @@
...
     }
 
     public function validate(int $value, array $allowedContentTypes = []): ?ValidationError
     {
         try {
-            $contentInfo = $this->contentService->loadContentInfo($value);
-            $contentType = $this->contentTypeService->loadContentType($contentInfo->contentTypeId);
+            $contentInfo = $this->repository->sudo(static function (Repository $repository) use ($value) {
+                return $repository->getContentService()->loadContentInfo($value);
+            });
```
alongosz: Please avoid injecting Repository to fetch services at runtime. This is allowed for integration tests only. For production code it's something that I plan to get rid of at some point. Repository was an attempt to wrap parts of eZ Publish architecture, in effect creating a container within a Symfony container. The same can be achieved on a SF Framework level without intermediate classes. There's also underlying performance issue there. For some reason we have at runtime 2 instances of `Repository` which occupy memory. Moreover fetching services via getter methods is slightly slower than when injected into constructor and you might end-up with a service obtained without proper SA context.

If you want `sudo`, there are 2 things to inject ATM (needs to be changed as well): `PermissionResolver` and `Repository`, while the latter one can be used only as a formal argument of `PermissionResolver::sudo(callback, Repository)` method call.
So it should be rather something like this:
```php
            $contentInfo = $this->permissionResolver->sudo(
                function (Repository $repository) use ($value): ContentInfo {
                    return $this->contentService->loadContentInfo($value);
                },
                $this->repository
            );
```
:exclamation:  BUT before applying please read other comments, in this case we don't need to inject API and use `sudo`

src/lib/Repository/Validator/TargetContentValidator.php
```diff
@@ -21,25 +19,21 @@
...
-    public function __construct(
-        ContentService $contentService,
-        ContentTypeService $contentTypeService
-    ) {
-        $this->contentService = $contentService;
-        $this->contentTypeService = $contentTypeService;
+    public function __construct(Repository $repository)
+    {
+        $this->repository = $repository;
     }
```
alongosz: This is not necessary, let's load lighter Persistence Layer.
```php
use eZ\Publish\SPI\Persistence\Content;
// ...
    public function __construct(
        Content\Handler $contentHandler,
        Content\Type\Handler $contentTypeHandler
    ) {
        $this->contentHandler = $contentHandler;
        $this->contentTypeHandler = $contentTypeHandler;
    }
```

src/lib/Repository/Validator/TargetContentValidator.php
```diff
@@ -21,25 +19,21 @@
...
     {
         try {
-            $contentInfo = $this->contentService->loadContentInfo($value);
-            $contentType = $this->contentTypeService->loadContentType($contentInfo->contentTypeId);
+            $contentInfo = $this->repository->sudo(static function (Repository $repository) use ($value) {
+                return $repository->getContentService()->loadContentInfo($value);
+            });
+            $contentType = $this->repository->getContentTypeService()->loadContentType($contentInfo->contentTypeId);
 
             if (!empty($allowedContentTypes) && !in_array($contentType->identifier, $allowedContentTypes, true)) {
```
alongosz: So following the previous comment:
```php
            $contentInfo = $this->contentHandler->loadContentInfo($value);
            $contentType = $this->contentTypeHandler->load($contentInfo->contentTypeId);

            if (
                !empty($allowedContentTypes) &&
                !in_array($contentType->identifier, $allowedContentTypes, true)
            ) {
```

alongosz (changes requested):

src/lib/Repository/Validator/TargetContentValidator.php
barw4: Thanks for the explanation

barw4:
Not sure about the CS check, both files were cs-fixed

Steveb-p:
> Not sure about the CS check, both files were cs-fixed

It's highly likely a php-cs-fixer update "uncovered" some new issues. Re-run `composer update` & `composer fix-cs`.

src/lib/Repository/Validator/TargetContentValidator.php
Steveb-p: To be frank saying that some Content ID does not exist (because that's what will actually be happening) is hardly a security issue. It's not like we're saying what usernames there are. Even if an ID is leaked, user is still not capable of reaching it. And we're depriving ourselves of a rather important debugging information :disappointed: 

Like in many other places, we're losing critical information, and when dealing with any bugs related to this code we will need to manually expose the ID.

So for me it's either this shouldn't be changed, or we need a follow up sometime to include relevant debugging information. (Adding internal information to an exception is not uncommon)

Steveb-p (approved):

alongosz (commented):
The CS issue is related to the changes, due to outdated fixer. Proceed as Steve suggested above please.

barw4:
@alongosz @Steveb-p CS was fixed and I also reverted the change with removing content identifier.

alongosz (approved):

Steveb-p (approved):

konradoboza (approved):

micszo (approved):
QA Approved on Ibexa Commerce 3.3.27-dev.

PR #334, opened 2022-09-02
IBX-3791: Fixed DFS queries for Postgres

Steveb-p:
| **JIRA issue**                          | [IBX-3791](https://issues.ibexa.co/browse/IBX-3791)

This PR allows DFS to work with Postgres database on a best effort basis.

It replaces two things:
 * ~Custom SQL code for upserts is now platform-dependent, with Postgres being tailored to work in the same fashion as MySQL one~ Code was replaced to comply with SQL standard in each case.
 * SQL select conditions no longer use `expired != 1` comparison, which is invalid (boolean vs. int) comparison in Postgres. They are replaced with `expired != true`, which is handled by both databases (MySQL treats `true` as `1` and `false` as `0` - see https://dev.mysql.com/doc/refman/8.0/en/boolean-literals.html).

#### Checklist:
- [x] ~Provided automated test coverage~.

### DIFF
--- a/eZ/Publish/Core/IO/IOMetadataHandler/LegacyDFSCluster.php
+++ b/eZ/Publish/Core/IO/IOMetadataHandler/LegacyDFSCluster.php
@@ -8,14 +8,13 @@ namespace eZ\Publish\Core\IO\IOMetadataHandler;
 
 use DateTime;
 use Doctrine\DBAL\Connection;
-use Doctrine\DBAL\DBALException;
+use Doctrine\DBAL\Exception;
 use eZ\Publish\Core\Base\Exceptions\InvalidArgumentException;
 use eZ\Publish\Core\IO\Exception\BinaryFileNotFoundException;
 use eZ\Publish\Core\IO\IOMetadataHandler;
 use eZ\Publish\Core\IO\UrlDecorator;
 use eZ\Publish\SPI\IO\BinaryFile as SPIBinaryFile;
 use eZ\Publish\SPI\IO\BinaryFileCreateStruct as SPIBinaryFileCreateStruct;
-use RuntimeException;
 
 /**
  * Manages IO metadata in a mysql table, ezdfsfile.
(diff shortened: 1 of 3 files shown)


eZ/Publish/Core/IO/IOMetadataHandler/LegacyDFSCluster.php
```diff
@@ -59,30 +59,39 @@ public function create(SPIBinaryFileCreateStruct $binaryFileCreateStruct)
...
-            $stmt->bindValue('size', $binaryFileCreateStruct->size);
-            $stmt->bindValue('scope', $this->getScope($binaryFileCreateStruct));
-            $stmt->bindValue('datatype', $binaryFileCreateStruct->mimeType);
-            $stmt->execute();
+SQL;
+        }
+
+
+        try {
+            $this->db->executeStatement($sql, $params);
```
alongosz: This is not a clean way to do it. It either has to be uniform ANSI SQL (can be more than one statement) or a strategy depending on DB. ATM this violates Open-Closed principle of SOLID.

alongosz (changes requested):

eZ/Publish/Core/IO/IOMetadataHandler/LegacyDFSCluster.php
Steveb-p: Agreed. As discussed with @webhdx on slack we will move to two separate statements. The end effect will be practically the same.

webhdx (approved):

adamwojs (approved):

eZ/Publish/Core/IO/IOMetadataHandler/LegacyDFSCluster.php
```diff
@@ -153,11 +164,28 @@ public function exists($spiBinaryFileId)
     {
         $path = (string)$this->addPrefix($spiBinaryFileId);
 
-        $stmt = $this->db->prepare('SELECT name FROM ezdfsfile WHERE name_hash LIKE ? and mtime > 0 and expired != 1');
-        $stmt->bindValue(1, md5($path));
-        $stmt->execute();
-
-        return $stmt->rowCount() == 1;
+        $qb = $this->db->createQueryBuilder();
```
alongosz: I would rather prefer explicit naming, like `queryBuilder`, but that one has also its advantages.

alongosz (approved):

eZ/Publish/Core/IO/IOMetadataHandler/LegacyDFSCluster.php
Steveb-p: Renamed as per request :)

micszo (approved):
Tested with DFS/NFS on Postgres and MySQL.

QA Approved on Ibexa Commerce 3.3.26-dev.

PR #336, opened 2022-09-22
IBX-3856: Changed Checkbox's value to never be considered empty

mnocon:
| **JIRA issue**                          | [IBX-3856](https://issues.ibexa.co/browse/IBX-3856)
| **Type**                                   |bugfix
| **BC breaks**                          |  yes, not significant

Replacement for https://github.com/ezsystems/ezplatform-version-comparison/pull/81 

I've:
1) implemented a failing test showing the desired behaviour (https://github.com/ezsystems/ezplatform-kernel/pull/336/commits/ab459ce3aa17f8ee78148f2f6c4cb64d8ff39ea1)
2) implemented the actual change (the new test now passes) (https://github.com/ezsystems/ezplatform-kernel/pull/336/commits/e01db7a34233f697c30e982cd924b516af3d188a)
3) smuggled a small improvement to the codebase (https://github.com/ezsystems/ezplatform-kernel/pull/336/commits/9f21465e5a7bcb3b6717f1648fcc514c331dc84f) - the `isEmptyValue(SPIValue $value)` signature does not allow nulls, this removal is even suggested by PHPStorm.

In the original discussion there was this suggestion:
```
From my point of view, isEmptyValue for CheckboxFieldtype should check only for null value as default value have that value set.
```

But I'm not sure about it - the default value seems to be `false` and `null` seems is not even specified as an allowed value:
https://github.com/ezsystems/ezplatform-kernel/blob/1.3/eZ/Publish/Core/Persistence/Legacy/Content/FieldValue/Converter/CheckboxConverter.php#L74
https://github.com/ezsystems/ezplatform-kernel/blob/1.3/eZ/Publish/Core/Persistence/Legacy/Content/FieldValue/Converter/CheckboxConverter.php#L39
https://github.com/ezsystems/ezplatform-kernel/blob/1.3/eZ/Publish/Core/FieldType/Checkbox/Value.php#L16-L21

Changing the default value to `null` (and allowing the value to be `bool|null`) seems like asking for trouble, because 3rd party code might not be ready to deal with nulls in the value - that's why I went the "this field is never empty" way.

If accepted, I will prepare a PR removing this condition:
https://github.com/ezsystems/ezplatform-admin-ui/blob/e9f2b07f26a47f0158417ea0e2ddf4f0ea5436b7/src/bundle/Resources/views/themes/admin/content/content_view_fields.html.twig#L25

as it will be no longer needed.

I will gladly accept any feedback, as my kernel-fu is very, very basic 😄 

#### Checklist:

### DIFF
--- a/eZ/Publish/API/Repository/Tests/FieldType/CheckboxIntegrationTest.php
+++ b/eZ/Publish/API/Repository/Tests/FieldType/CheckboxIntegrationTest.php
@@ -387,18 +387,16 @@ class CheckboxIntegrationTest extends SearchBaseIntegrationTest
         self::assertSame($isActive, $value->bool);
     }
 
-    public function testAddFieldDefinition()
+    public function testAddFieldDefinition(): void
     {
+        $fieldTypeService = $this->getRepository()->getFieldTypeService();
         $content = $this->addFieldDefinition();
 
-        $this->assertCount(2, $content->getFields());
-
-        $this->assertFalse(
-            $this->getRepository()->getFieldTypeService()->getFieldType(
-                $this->getTypeName()
-            )->isEmptyValue(
-                $content->getFieldValue('data')
-            )
+        self::assertCount(2, $content->getFields());
+        self::assertFalse(
+            $fieldTypeService
+                ->getFieldType($this->getTypeName())
+                ->isEmptyValue($content->getFieldValue('data'))
         );
     }
 
--- a/eZ/Publish/API/Repository/Tests/FieldType/CheckboxIntegrationTest.php
+++ b/eZ/Publish/API/Repository/Tests/FieldType/CheckboxIntegrationTest.php
@@ -306,22 +306,17 @@ class CheckboxIntegrationTest extends SearchBaseIntegrationTest
 
     public function providerForTestIsEmptyValue()
     {
-        return [
-            [new CheckboxValue()],
-            [new CheckboxValue(null)],
-            [new CheckboxValue(false)],
-        ];
+        return [];
     }
 
     public function providerForTestIsNotEmptyValue()
     {
         return [
-            [
-                $this->getValidCreationFieldData(),
-            ],
-            [
-                new CheckboxValue(true),
-            ],
+            [$this->getValidCreationFieldData()],
+            [new CheckboxValue(true)],
+            [new CheckboxValue()],
+            [new CheckboxValue(null)],
+            [new CheckboxValue(false)],
         ];
     }
 
@@ -390,6 +385,21 @@ class CheckboxIntegrationTest extends SearchBaseIntegrationTest
         $value = $contentItem->getField('is_active')->value;
         /** @var \eZ\Publish\Core\FieldType\Checkbox\Value $value */
         self::assertSame($isActive, $value->bool);
+    }
+
+    public function testAddFieldDefinition()
+    {
+        $content = $this->addFieldDefinition();
+
+        $this->assertCount(2, $content->getFields());
+
+        $this->assertFalse(
+            $this->getRepository()->getFieldTypeService()->getFieldType(
+                $this->getTypeName()
+            )->isEmptyValue(
+                $content->getFieldValue('data')
+            )
+        );
     }
 
     protected function createCheckboxContentItems(Repository $repository): void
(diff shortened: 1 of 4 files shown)


ViniTou (approved):

eZ/Publish/API/Repository/Tests/FieldType/CheckboxIntegrationTest.php
```diff
@@ -392,6 +387,21 @@ public function testFindContentFieldCriterion(bool $isActive): void
         self::assertSame($isActive, $value->bool);
     }
 
+    public function testAddFieldDefinition()
```
alongosz:
```suggestion
    public function testAddFieldDefinition(): void
```

eZ/Publish/API/Repository/Tests/FieldType/CheckboxIntegrationTest.php
```diff
@@ -392,6 +387,21 @@ public function testFindContentFieldCriterion(bool $isActive): void
...
+    {
+        $content = $this->addFieldDefinition();
+
+        $this->assertCount(2, $content->getFields());
```
alongosz: Would like to see sanity check prior `addFieldDefinition` - assert count = 1

alongosz: Static methods should be called statically.
```suggestion
        self::assertCount(2, $content->getFields());
```

eZ/Publish/API/Repository/Tests/FieldType/CheckboxIntegrationTest.php
```diff
@@ -392,6 +387,21 @@ public function testFindContentFieldCriterion(bool $isActive): void
...
+
+        $this->assertFalse(
```
alongosz:
```suggestion
        self::assertFalse(
```

eZ/Publish/API/Repository/Tests/FieldType/CheckboxIntegrationTest.php
```diff
@@ -392,6 +387,21 @@ public function testFindContentFieldCriterion(bool $isActive): void
...
+            $this->getRepository()->getFieldTypeService()->getFieldType(
```
alongosz: Readability, please extract
```php
$fieldTypeService = $this->getRepository()->getFieldTypeService()
```
as the first call of the method. If you need to call it after `addFieldDefinition` then it should be called as `getRepository(false)` because the method has side effects.

eZ/Publish/Core/FieldType/Tests/CheckboxTest.php
```diff
@@ -297,4 +297,30 @@ public function provideDataForGetName(): array
             [new CheckboxValue(false), '0', [], 'en_GB'],
         ];
     }
+
+    /**
+     * @dataProvider provideForValueIsNeverEmpty
+     */
+    public function testValueIsNeverEmpty(CheckboxValue $value)
```
alongosz:
```suggestion
    public function testValueIsNeverEmpty(CheckboxValue $value): void
```

alongosz (changes requested):

alongosz:
@ezsystems/documentation-team this is behavior change - CheckBox Field Type after the change is not going to be considered as empty in any case - it can be either true or false. This should be reflected on the Doc page for that Field Type.

eZ/Publish/API/Repository/Tests/FieldType/CheckboxIntegrationTest.php
mnocon: Not sure how to do this - this test is a variation of the base one:
https://github.com/ezsystems/ezplatform-kernel/blob/59ca1434ce8fb5169d39db258dfdc2ebc249120b/eZ/Publish/API/Repository/Tests/FieldType/BaseIntegrationTest.php#L1009-L1022

But I had to change the assertion.

The method 'addFieldDefinition` first removes the fieldDefinition:
https://github.com/ezsystems/ezplatform-kernel/blob/59ca1434ce8fb5169d39db258dfdc2ebc249120b/eZ/Publish/API/Repository/Tests/FieldType/BaseIntegrationTest.php#L985
and then adds it again

The flow inside `addFieldDefinition` looks like this then:
- The Content item is created (with 2 fields)
- the tested field definition is removed
- the tested field definition is added again

Current approach makes the `$content` variable available only after the call to `addFieldDefinition` finishes, I can't really add that sanity check without reworking this test.

I wanted to reuse the existing setup methods, but if you think it's good to add some custom setup for this test then I will do it

alongosz (approved):

webhdx (approved):

tomaszszopinski (approved):
QA approved IbexaDXP 3.3 commerce.

Steveb-p (approved):

PR #338, opened 2022-10-03
Added data type for sudo $callback argument for static analysis

Steveb-p:

This PR adds type information for static analysis tools regarding `$callback` callable argument for `Repository::sudo`.

This ensures that code using the callback will have matching return type as needed.

#### Checklist:
- [ ] ~Provided automated test coverage.~

### DIFF
--- a/eZ/Publish/Core/Repository/Permission/PermissionResolver.php
+++ b/eZ/Publish/Core/Repository/Permission/PermissionResolver.php
@@ -405,7 +405,7 @@ class PermissionResolver implements PermissionResolverInterface
      *         }
      *     );
      *
-     * @param \Closure $callback
+     * @param \callable(\eZ\Publish\API\Repository\Repository): mixed $callback
      * @param \eZ\Publish\API\Repository\Repository $outerRepository
      *
      * @throws \RuntimeException Thrown on recursive sudo() use.
--- a/eZ/Publish/API/Repository/Repository.php
+++ b/eZ/Publish/API/Repository/Repository.php
@@ -25,14 +25,15 @@ interface Repository
      *         }
      *     );
      *
+     * @template T
      *
-     * @param callable $callback
+     * @param callable(\eZ\Publish\API\Repository\Repository): T $callback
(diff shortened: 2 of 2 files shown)


konradoboza (approved):

eZ/Publish/Core/Repository/Permission/PermissionResolver.php
```diff
@@ -405,7 +405,7 @@ private function isDeniedByRoleLimitation(
      *         }
      *     );
      *
-     * @param \Closure $callback
+     * @param \callable(\eZ\Publish\API\Repository\Repository): mixed $callback
```
alongosz, line 408: raising internal question here: why does the definition of this parameter differ from `Repository::sudo`?

alongosz (commented):
Could be ok for now. My plan was rather to deprecate `sudo` on Repository and introduce contract for `PermissionResolver::sudo`, but without 2nd argument of `Repository` type. There is cyclic dependency here which negatively impacts building the Service Container. Repository depends on PermissionResolver which depends on Repository.

eZ/Publish/Core/Repository/Permission/PermissionResolver.php
Steveb-p: There seems to be some sort of a difference between declaring types like this for an interface and for an actual implementation (or at least, when it's interface-less).

See https://github.com/phpstan/phpstan/issues/4913 / https://phpstan.org/r/370e514d-13da-45c2-b735-478143d91976.

Steveb-p:
> Could be ok for now. My plan was rather to deprecate `sudo` on Repository and introduce contract for `PermissionResolver::sudo`, but without 2nd argument of `Repository` type. There is cyclic dependency here which negatively impacts building the Service Container. Repository depends on PermissionResolver which depends on Repository.

I'm simply properly documenting current state.

ViniTou (approved):

adamwojs (approved):

webhdx (approved):

alongosz (approved):

PR #344, opened 2022-10-18
IBX-4018: Implement `supportsDenormalization` method for `CompoundMatcherNormalizer`

barw4:
| **JIRA issue**                          | [IBX-4018](https://issues.ibexa.co/browse/IBX-4018)

Following PR prevents all objects to be denormalized using `CompoundMatcherNormalizer` when using Symfony's serializer.

#### Checklist:

### DIFF
--- a/eZ/Publish/Core/MVC/Symfony/Component/Serializer/CompoundMatcherNormalizer.php
+++ b/eZ/Publish/Core/MVC/Symfony/Component/Serializer/CompoundMatcherNormalizer.php
@@ -31,4 +31,9 @@ class CompoundMatcherNormalizer extends AbstractPropertyWhitelistNormalizer
     {
         return $data instanceof Matcher\Compound;
     }
+
+    public function supportsDenormalization($data, string $type, string $format = null)
+    {
+        return $type === Matcher\Compound::class;
+    }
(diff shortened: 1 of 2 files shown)


Steveb-p (approved):
I can agree to the fix purely because the issue stems from the original bad pick for an abstract.

`PropertyNormalizer` is usually a bad pick for a parent class, because if you know how a class looks like (which is the case, we're looking for `Matcher\Compound`), then you also know what properties the class has and there is no need to use "magic" `PropertyNormalizer`, which includes reading class reflection data (sometimes cached) to perform the normalization.

What should have happened instead is, we should have had a class that very accurately picks the right properties and normalizes/denormalizes the object.

adamwojs (approved):

konradoboza (approved):

tomaszszopinski (approved):
QA approved on Ibexa DXP 3.3 commerce.

Steveb-p:
Merged up into `4.2` in [1fcb7c4e](https://github.com/ibexa/core/commit/1fcb7c4e).
Merged up into `main` in [1cf713a0](https://github.com/ibexa/core/commit/1cf713a0)

PR #347, opened 2022-11-09
IBX-4276: Created command to revoke user passwords

webhdx:
| **JIRA issue**                          | [IBX-4276](https://issues.ibexa.co/browse/IBX-4276)

`php bin/console ibexa:user:expire-password`

This command sets password as expired for selected users. This can be helpful in case of sensitive data leakage when you want to force users to change their passwords.

During the run, the command will check if user content type is configured for password expiration, if not, it'll update its configuration.

There are a few options to change the bahavior:
- `--user-id|-u` specify users using IDs, you can pass multiple `--user-id` options
- `--user-group-id|-ug` expire password of all users assigned to specific group, accepts multiple groups
- `--user-content-type-identifier|-ct` expire passwords of all users using specific content, accepts multiple content types
- `--force|-f` commit the change
- `--iteration-count|-c` how many users will be fetched at once, lowering this can be helpful to mitigate memory issues
- `--password-ttl|-t` when CT is updated it'll set this value as after how many days passwords will expire

## Example usage:
`php bin/console ibexa:user:expire-password --user-content-type-identifier=user --force`
> Expire passwords of all users using `user` content type, commit operation.

`php bin/console ibexa:user:expire-password --user-group-id=13`
> Expire passwords of all users using assigned to user group with ID 13, perform a dry-run (no change will be saved in the database)

#### Checklist:

### DIFF
--- a/eZ/Bundle/EzPublishCoreBundle/Resources/config/commands.yml
+++ b/eZ/Bundle/EzPublishCoreBundle/Resources/config/commands.yml
@@ -59,6 +59,6 @@ services:
         autowire: true
         autoconfigure: true
         arguments:
-            $userHandler: '@ezpublish.spi.persistence.legacy.user.handler'
+            $userHandler: '@ezpublish.spi.persistence.user_handler'
         tags:
             - { name: console.command }
--- a/eZ/Bundle/EzPublishCoreBundle/Resources/config/commands.yml
+++ b/eZ/Bundle/EzPublishCoreBundle/Resources/config/commands.yml
@@ -54,3 +54,11 @@ services:
             $ioService: '@ezpublish.fieldType.ezimage.io_service'
         tags:
             - { name: console.command }
+
+    Ibexa\Bundle\Core\Command\ExpireUserPasswordsCommand:
+        autowire: true
+        autoconfigure: true
+        arguments:
+            $userHandler: '@ezpublish.spi.persistence.legacy.user.handler'
+        tags:
+            - { name: console.command }
--- a/src/bundle/Core/Command/ExpireUserPasswordsCommand.php
+++ b/src/bundle/Core/Command/ExpireUserPasswordsCommand.php
@@ -1,0 +1,374 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace Ibexa\Bundle\Core\Command;
+
+use Exception;
+use eZ\Publish\API\Repository\Repository;
+use eZ\Publish\API\Repository\Values\Content\Content;
+use eZ\Publish\API\Repository\Values\Content\ContentList;
+use eZ\Publish\API\Repository\Values\Content\Field;
+use eZ\Publish\API\Repository\Values\Content\Query;
+use eZ\Publish\API\Repository\Values\Content\Query\Criterion\LogicalAnd;
+use eZ\Publish\API\Repository\Values\ContentType\ContentType;
+use eZ\Publish\API\Repository\Values\ContentType\FieldDefinition;
+use eZ\Publish\API\Repository\Values\Filter\Filter;
+use eZ\Publish\SPI\Persistence\User\Handler;
+use InvalidArgumentException;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class ExpireUserPasswordsCommand extends Command
+{
+    protected static $defaultName = 'ibexa:user:expire-password';
+
+    public const DEFAULT_BATCH_SIZE = 50;
+
+    public const BEFORE_RUNNING_HINTS = <<<EOT
+<error>Before you continue:</error>
+- Make sure to back up your database.
+- Take installation offline, during the script execution the database should not be modified.
+- Run this command without memory limit.
+- Run this command in production environment using <info>--env=prod</info>
+EOT;
+
+    /** @var \eZ\Publish\API\Repository\Repository */
+    private $repository;
+
+    /** @var \eZ\Publish\SPI\Persistence\User\Handler */
+    private $userHandler;
+
+    public function __construct(
+        Repository $repository,
+        Handler $userHandler
+    ) {
+        $this->repository = $repository;
+        $this->userHandler = $userHandler;
+
+        parent::__construct();
+    }
+
+    protected function configure()
+    {
+        $beforeRunningHints = self::BEFORE_RUNNING_HINTS;
+        $this
+            ->setDescription('Expire passwords for selected users.')
+            ->addOption(
+                'user-id',
+                null,
+                InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
+                'Expire password for specific User identified by ID'
+            )
+            ->addOption(
+                'user-group-id',
+                null,
+                InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
+                'Expire passwords of all users assigned to specific User Group'
+            )
+            ->addOption(
+                'user-content-type-identifier',
+                null,
+                InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
+                'Expire passwords of all users based on specific Content Type'
+            )
+            ->addOption(
+                'force',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'Perform setting user passwords as expired',
+                false
+            )
+            ->addOption(
+                'batch-size',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'Number of users to process at once',
+                self::DEFAULT_BATCH_SIZE
+            )
+            ->setHelp(
+                <<<EOT
+The command <info>%command.name%</info> expires passwords of specific users. 
+Use this tool wisely as next time affect users tries to log in, they will be forced to set a new password.
+Note: This script can potentially run for a very long time, and in Symfony dev environment it will consume memory exponentially with size of dataset.
+
+{$beforeRunningHints}
+EOT
+            );
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $batchSize = $input->getOption('batch-size');
+        $userIds = $input->getOption('user-id');
+        $userGroupIds = $input->getOption('user-group-id');
+        $force = $input->getOption('force');
+
+        if (!empty($userIds) && !empty($userGroupIds)) {
+            throw new InvalidArgumentException('You cannot use --user-id and --user-group-id options at once.');
+        }
+
+        $userContentTypeIdentifiers = $input->getOption('user-content-type-identifier');
+
+        $this->summarizeSearchCriteria($output, $userIds, $userGroupIds, $userContentTypeIdentifiers);
+
+        $criteria = $this->getCriteria($userIds, $userGroupIds, $userContentTypeIdentifiers);
+
+        $filter = new Filter(new LogicalAnd($criteria));
+        $filter->withLimit(0);
+
+        $contentList = $this->repository->sudo(static function (Repository $repository) use ($filter): ContentList {
+            return $repository->getContentService()->find($filter);
+        });
+        $count = $contentList->getTotalCount();
+
+        if ($count === 0) {
+            $output->writeln('<info>There are no users matching given criteria</info>');
+
+            return Command::SUCCESS;
+        }
+
+        $output->writeln(sprintf(
+            '<info>Found %d User(s) matching given criteria</info>',
+            $count
+        ));
+
+        $displayProgressBar = !($output->isVerbose() || $output->isVeryVerbose() || $output->isDebug());
+
+        if ($displayProgressBar) {
+            $progressBar = new ProgressBar($output, $count);
+            $progressBar->setFormat(
+                '%current%/%max% [%bar%] %percent:3s%% %elapsed:6s%/%estimated:-6s% %memory:6s%' . PHP_EOL
+            );
+            $progressBar->start();
+        }
+
+        $processedUsersCount = 0;
+        $processedContentTypes = [];
+
+        $this->repository->beginTransaction();
+        try {
+            do {
+                $filter
+                    ->withLimit((int) $batchSize)
+                    ->withOffset($processedUsersCount);
+
+                $contentList = $this->repository->sudo(static function (Repository $repository) use ($filter): ContentList {
+                    return $repository->getContentService()->find($filter);
+                });
+
+                /** @var \eZ\Publish\API\Repository\Values\Content\Content $content */
+                foreach ($contentList as $content) {
+                    $fields = $content->getFields();
+                    $userField = $this->findUserField($fields);
+
+                    if (null === $userField) {
+                        // content item isn't of a user content type
+                        continue;
+                    }
+
+                    $contentType = $content->getContentType();
+                    $fieldDefinition = $contentType->getFieldDefinition($userField->fieldDefIdentifier);
+
+                    $validatorConfiguration = $fieldDefinition->getValidatorConfiguration();
+                    $fieldSettings = $fieldDefinition->getFieldSettings();
+
+                    if (
+                        (
+                            !$validatorConfiguration['PasswordValueValidator']['requireNewPassword']
+                            || 0 === $fieldSettings['PasswordTTL']
+                        )
+                        && !in_array($contentType->identifier, $processedContentTypes, true)
+                    ) {
+                        $output->writeln(sprintf(
+                            '<info>Content Type "%s" needs to be updated</info>',
+                            $contentType->identifier
+                        ));
+
+                        //  enforce CT to use password expiration feature
+                        $validatorConfiguration['PasswordValueValidator']['requireNewPassword'] = true;
+                        $fieldSettings['PasswordTTL'] = 90;
+
+                        $this->updateContentType(
+                            $contentType,
+                            $fieldSettings,
+                            $validatorConfiguration,
+                            $fieldDefinition
+                        );
+
+                        $processedContentTypes[] = $contentType->identifier;
+                    }
+
+                    $spiUser = $this->userHandler->load($content->id);
+                    $updatedUser = clone $spiUser;
+                    $updatedUser->passwordUpdatedAt = 1;
+                    $this->userHandler->updatePassword($updatedUser);
+
+                    ++$processedUsersCount;
+
+                    if ($displayProgressBar) {
+                        $progressBar->advance(1);
+                    }
+                }
+            } while ($processedUsersCount < $count);
+        } catch (Exception $e) {
+            $this->repository->rollback();
+            $output->writeln(
+                '<error>Something went wrong. See the exception below, '
+                . 'fix the issue and rerun this command</error>'
+            );
+
+            throw $e;
+        }
+
+        if ($force) {
+            $this->repository->commit();
+        } else {
+            $this->repository->rollback();
+
+            $output->writeln(
+                '<warning>No changes made. If you want to proceed rerun '
+                . 'this command with --force flag.</warning>'
+            );
+        }
+
+        $output->writeln(sprintf(
+            '<info>Expired passwords of %d user(s)</info>',
+            $processedUsersCount
+        ));
+
+        return Command::SUCCESS;
+    }
+
+    /**
+     * @param array<int> $userIds
+     * @param array<int> $userGroupIds
+     * @param array<string> $userContentTypeIdentifiers
+     *
+     * @return array<\eZ\Publish\API\Repository\Values\Content\Query\Criterion>
+     */
+    private function getCriteria(
+        array $userIds,
+        array $userGroupIds,
+        array $userContentTypeIdentifiers
+    ): array {
+        $criteria = [];
+
+        if (!empty($userIds)) {
+            $criteria[] = new Query\Criterion\ContentId($userIds);
+        }
+
+        if (!empty($userGroupIds)) {
+            $criteria[] = new Query\Criterion\ParentLocationId($userGroupIds);
+        }
+
+        if (!empty($userContentTypeIdentifiers)) {
+            $criteria[] = new Query\Criterion\ContentTypeIdentifier($userContentTypeIdentifiers);
+        }
+
+        return $criteria;
+    }
+
+    /**
+     * @param array<int> $userIds
+     * @param array<int> $userGroupIds
+     * @param array<string> $userContentTypeIdentifiers
+     */
+    private function summarizeSearchCriteria(
+        OutputInterface $output,
+        $userIds,
+        $userGroupIds,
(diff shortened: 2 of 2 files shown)


src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+use eZ\Publish\API\Repository\Values\Filter\Filter;
+use eZ\Publish\SPI\Persistence\User\Handler;
+use InvalidArgumentException;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class ExpireUserPasswordsCommand extends Command
```
adamwojs:
```suggestion
final class ExpireUserPasswordsCommand extends Command
```

konradoboza:
```suggestion
final class ExpireUserPasswordsCommand extends Command
```

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+            )
+            ->addOption(
+                'force',
+                null,
+                InputOption::VALUE_NONE,
+                'Perform setting user passwords as expired'
+            )
+            ->addOption(
+                'batch-size',
+                null,
```
alongosz: Convention: multiple commands in this bundle for such thing use `iteration-count` and `c` shortcut.

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+            );
+            $progressBar->start();
+        }
+
+        $processedUsersCount = 0;
+        $processedContentTypes = [];
+
+        $this->repository->beginTransaction();
+        try {
+            do {
```
konradoboza, line 171: It feels like the content of `do ... while` loop can be extracted to at least one method, it's hardly readable to me in the current form tbh.

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+
+    /**
+     * @param array<\eZ\Publish\API\Repository\Values\Content\Field> $fields
+     */
+    private function findUserField(array $fields): ?Field
+    {
+        $userField = null;
+
+        foreach ($fields as $field) {
+            if ($field->fieldTypeIdentifier !== 'ezuser') {
```
konradoboza: `ezuser` could be a const.

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+
+    protected function configure()
+    {
+        $beforeRunningHints = self::BEFORE_RUNNING_HINTS;
+        $this
+            ->setDescription('Expire passwords for selected users.')
+            ->addOption(
+                'user-id',
+                null,
+                InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
```
adamwojs:
```suggestion
                InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY,
```

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+        Repository $repository,
+        Handler $userHandler
+    ) {
+        $this->repository = $repository;
+        $this->userHandler = $userHandler;
+
+        parent::__construct();
+    }
+
+    protected function configure()
```
alongosz: You can safely add here
```suggestion
    protected function configure(): void
```

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+        $count = $contentList->getTotalCount();
+
+        if ($count === 0) {
+            $output->writeln('<info>There are no users matching given criteria</info>');
+
+            return Command::SUCCESS;
+        }
+
+        $output->writeln(sprintf(
+            '<info>Found %d User(s) matching given criteria</info>',
```
konradoboza: For consistency:
```suggestion
            '<info>Found %d users matching given criteria</info>',
```

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+
+        $displayProgressBar = !($output->isVerbose() || $output->isVeryVerbose() || $output->isDebug());
+
+        if ($displayProgressBar) {
+            $progressBar = new ProgressBar($output, $count);
+            $progressBar->setFormat(
+                '%current%/%max% [%bar%] %percent:3s%% %elapsed:6s%/%estimated:-6s% %memory:6s%' . PHP_EOL
+            );
+            $progressBar->start();
+        }
```
Steveb-p: IMO it's better to always display the ProgressBar and let it determine the format on it's own (as specified in Symfony docs). Developers can always silence it using `--quiet` command option.

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+        $force = $input->getOption('force');
+        $passwordTTL = $input->getOption('password-ttl');
+
+        if (!empty($userIds) && !empty($userGroupIds)) {
+            throw new InvalidArgumentException('You cannot use --user-id and --user-group-id options at once.');
+        }
+
+        $userContentTypeIdentifiers = $input->getOption('user-content-type-identifier');
+
+        $this->summarizeSearchCriteria($output, $userIds, $userGroupIds, $userContentTypeIdentifiers);
```
konradoboza: Maybe:
```suggestion
        $this->supplySearchCriteria($output, $userIds, $userGroupIds, $userContentTypeIdentifiers);
```
would sound better?

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class ExpireUserPasswordsCommand extends Command
+{
+    protected static $defaultName = 'ibexa:user:expire-password';
+
+    public const DEFAULT_BATCH_SIZE = 50;
+
+    public const DEFAULT_PASSWORD_TTL = 00;
```
adamwojs:
```suggestion
    public const DEFAULT_PASSWORD_TTL = 0;
```

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+        $this->repository->beginTransaction();
+        try {
+            do {
+                $filter
+                    ->withLimit((int) $batchSize)
+                    ->withOffset($processedUsersCount);
+
+                $contentList = $this->repository->sudo(static function (Repository $repository) use ($filter): ContentList {
+                    return $repository->getContentService()->find($filter);
+                });
```
alongosz: This is the anti-pattern we're trying to move away from. You need to inject ContentService to the outer class and use it in the callback instead of relying on Repository passed as callback argument.

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+    /**
+     * @param array<int> $userIds
+     * @param array<int> $userGroupIds
+     * @param array<string> $userContentTypeIdentifiers
+     */
+    private function summarizeSearchCriteria(
+        OutputInterface $output,
+        $userIds,
+        $userGroupIds,
+        $userContentTypeIdentifiers
```
alongosz: missing array type declarations

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+            $output->writeln(
+                '<error>Something went wrong. See the exception below, '
+                . 'fix the issue and rerun this command</error>'
+            );
+
+            throw $e;
+        }
+
+        $output->writeln(sprintf(
+            '<info>Expired passwords of %d user(s)</info>',
```
konradoboza:
```suggestion
            '<info>Expired passwords of %d users</info>',
```

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+        }
+
+        $userContentTypeIdentifiers = $input->getOption('user-content-type-identifier');
+
+        $this->summarizeSearchCriteria($output, $userIds, $userGroupIds, $userContentTypeIdentifiers);
+
+        $criteria = $this->getCriteria($userIds, $userGroupIds, $userContentTypeIdentifiers);
+
+        $filter = new Filter(new LogicalAnd($criteria));
+        $filter->withLimit(0);
```
alongosz: Limit by default is 0

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+
+        foreach ($fields as $field) {
+            if ($field->fieldTypeIdentifier !== 'ezuser') {
+                continue;
+            }
+            $userField = $field;
+        }
+
+        return $userField;
+    }
```
adamwojs: `ezuser` FT is is singular so you can simplify this method to:

```suggestion
    /**
     * @param array<\eZ\Publish\API\Repository\Values\Content\Field> $fields
     */
    private function findUserField(array $fields): ?Field
    {
        foreach ($fields as $field) {
            if ($field->fieldTypeIdentifier === 'ezuser') {
                return $field;
            }
        }

        return null;
    }
```

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+
+            return Command::SUCCESS;
+        }
+
+        $output->writeln(sprintf(
+            '<info>Found %d User(s) matching given criteria</info>',
+            $count
+        ));
+
+        $displayProgressBar = !($output->isVerbose() || $output->isVeryVerbose() || $output->isDebug());
```
alongosz: By default on normal output progress bar is displayed. If you want it not to be displayed, you typically execute your command with `-q` switch. This is controlled by:

`\Symfony\Component\Console\Output\Output::write:`
```php
        if ($verbosity > $this->getVerbosity()) {
            return;
        }
```

and TBH I don't see a reason we should "mingle" with Symfony's default behavior

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+            } while ($processedUsersCount < $count);
+        } catch (Exception $e) {
+            $this->repository->rollback();
+            $output->writeln(
+                '<error>Something went wrong. See the exception below, '
+                . 'fix the issue and rerun this command</error>'
+            );
+
+            throw $e;
+        }
```
alongosz: Could you please refactor this big block into smaller parts so it's more readable? Maybe at least contents of try-catch could be refactored into a method with maybe some inner methods call. Cognitive complexity now is very high.

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+            throw new InvalidArgumentException('You cannot use --user-id and --user-group-id options at once.');
+        }
+
+        $userContentTypeIdentifiers = $input->getOption('user-content-type-identifier');
+
+        $this->summarizeSearchCriteria($output, $userIds, $userGroupIds, $userContentTypeIdentifiers);
+
+        $criteria = $this->getCriteria($userIds, $userGroupIds, $userContentTypeIdentifiers);
+
+        $filter = new Filter(new LogicalAnd($criteria));
```
alongosz: Filter is fluent so you don't have to wrap it artificially.
This and `getCriteria` above could be:
```php
        $filter = new Filter();
        if (!empty($userIds)) {
            $filter->andWithCriterion(new Query\Criterion\ContentId($userIds));
        }

        if (!empty($userGroupIds)) {
            $filter->andWithCriterion(new Query\Criterion\ParentLocationId($userGroupIds));
        }

        if (!empty($userContentTypeIdentifiers)) {
            $filter->andWithCriterion(new Query\Criterion\ContentTypeIdentifier($userContentTypeIdentifiers));
        }
```

It's fine to group this into a method, like you did (due to the mentioned cognitive complexity), but then how about
```php
        $filter = new Filter();
        $this->applyCriteria($filter, $userIds, $userGroupIds, $userContentTypeIdentifiers);
```
?

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+     */
+    private function findUserField(array $fields): ?Field
+    {
+        $userField = null;
+
+        foreach ($fields as $field) {
+            if ($field->fieldTypeIdentifier !== 'ezuser') {
+                continue;
+            }
+            $userField = $field;
```
alongosz: how about:
```suggestion
            if ($field->fieldTypeIdentifier === 'ezuser') {
                return $field;
            }
```

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+     */
+    private function updateContentType(
+        ContentType $contentType,
+        array $fieldSettings,
+        array $validatorConfiguration,
+        FieldDefinition $fieldDefinition
+    ): void {
+        $this->repository->sudo(
+            static function (
+                Repository $repository
```
alongosz: the same remark for `sudo` with callback here as well. We want to avoid using Repository getters as much as possible.

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+
+                    $validatorConfiguration = $fieldDefinition->getValidatorConfiguration();
+                    $fieldSettings = $fieldDefinition->getFieldSettings();
+
+                    if (
+                        (
+                            !$validatorConfiguration['PasswordValueValidator']['requireNewPassword']
+                            || 0 === $fieldSettings['PasswordTTL']
+                        )
+                        && !in_array($contentType->identifier, $processedContentTypes, true)
```
alongosz: This statement/expression could be wrapped by self-commenting method call

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+            $this->repository->rollback();
+            $output->writeln(
+                '<error>Something went wrong. See the exception below, '
+                . 'fix the issue and rerun this command</error>'
+            );
+
+            throw $e;
+        }
+
+        $output->writeln(sprintf(
```
alongosz, line 203: Somewhere here you should have
```php
$progressBar->finish();
```

eZ/Bundle/EzPublishCoreBundle/Resources/config/commands.yml
```diff
@@ -54,3 +54,11 @@ services:
             $ioService: '@ezpublish.fieldType.ezimage.io_service'
         tags:
             - { name: console.command }
+
+    Ibexa\Bundle\Core\Command\ExpireUserPasswordsCommand:
+        autowire: true
+        autoconfigure: true
+        arguments:
+            $userHandler: '@ezpublish.spi.persistence.legacy.user.handler'
```
alongosz: This is Legacy storage handler implementation. What you want here is top layer handler rather:
```suggestion
            $userHandler: '@ezpublish.spi.persistence.user_handler'
```
unless... was there a specific reason for injecting that one?

alongosz (changes requested):

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+
+        $this->summarizeSearchCriteria($output, $userIds, $userGroupIds, $userContentTypeIdentifiers);
+
+        $criteria = $this->getCriteria($userIds, $userGroupIds, $userContentTypeIdentifiers);
+
+        $filter = new Filter(new LogicalAnd($criteria));
+        $filter->withLimit(0);
+
+        $contentList = $this->repository->sudo(static function (Repository $repository) use ($filter): ContentList {
+            return $repository->getContentService()->find($filter);
```
mikadamczyk: `ContentService` should be used directly here

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,383 @@
...
+
+                    $spiUser = $this->userHandler->load($content->id);
+                    $updatedUser = clone $spiUser;
+                    $updatedUser->passwordUpdatedAt = 1;
+                    $this->userHandler->updatePassword($updatedUser);
+
+                    ++$processedUsersCount;
+
+                    if ($displayProgressBar) {
+                        $progressBar->advance(1);
```
mikadamczyk:
```suggestion
                        $progressBar->advance();
```

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
webhdx: WDYM? It needs to be used together with `sudo` as we expect to find all users despite permissions.

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
webhdx: It was inspired by one of the other commands. But I agree with you and fixed it.

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
mikadamczyk: sorry I was thinking about not using a repository here

webhdx: Ok. Fixed: [6687acf](https://github.com/ezsystems/ezplatform-kernel/pull/347/commits/6687acfcdd4e978296fffbf6f1c8240b8a6c5d04)

webhdx:
Thanks for all the comments. I've fixed everything in [6687acf](https://github.com/ezsystems/ezplatform-kernel/pull/347/commits/6687acfcdd4e978296fffbf6f1c8240b8a6c5d04)

mikadamczyk (approved):

glye (approved):

alongosz (approved):

mnocon (approved):

mnocon (approved):

konradoboza (approved):

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,446 @@
...
+                't',
+                InputOption::VALUE_REQUIRED,
+                'After how many days passwords expire. Set when Content Type needs to be updated.',
+                self::DEFAULT_PASSWORD_TTL
+            )
+            ->setHelp(
+                <<<EOT
+The command <info>%command.name%</info> expires passwords of specific users. 
+Use this tool wisely as next time affect users tries to log in, they will be forced to set a new password.
+Note: This script can potentially run for a very long time, and in Symfony dev environment it will consume memory exponentially with size of dataset.
```
Steveb-p, line 121: This is actually primarily due to incorrect logging settings (well, correct for dev environment). If a developer wants to execute this in dev env, one should ensure that debugging messages are not buffered (`monolog.yaml`).

That is what usually happens. Just a note for us, I don't expect a note change.

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
```diff
@@ -0,0 +1,446 @@
...
+        }
+
+        $fieldDefinition = $fields->first();
+
+        $validatorConfiguration = $fieldDefinition->getValidatorConfiguration();
+        $fieldSettings = $fieldDefinition->getFieldSettings();
+
+        $isUpdateNeeded = !$validatorConfiguration['PasswordValueValidator']['requireNewPassword']
+            || 0 === $fieldSettings['PasswordTTL'];
+        $isContentTypeAlreadyProcessed = in_array($contentType->identifier, $processedContentTypes, true);
```
Steveb-p: This can be established way earlier (at the beginning of the method) and used as an early return.

Steveb-p (approved):

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
webhdx: Alright, fixed in [50cf4fa](https://github.com/ezsystems/ezplatform-kernel/pull/347/commits/50cf4fad6a44ff4f50320b7b3551aab0ba2d6670)

src/bundle/Core/Command/ExpireUserPasswordsCommand.php
alongosz: > This is actually primarily due to incorrect logging settings (well, correct for dev environment). If a developer wants to execute this in dev env, one should ensure that debugging messages are not buffered (`monolog.yaml`).
> 
> That is what usually happens. Just a note for us, I don't expect a note change.

@Steveb-p this is true, but that's not the only one issue.
1. Symfony profiling afaik is enabled by default in debug mode, which on a long running operation iterating multiple times over the same code can cause big overhead
2. PDO due to Doctrine v2 extending PDOStatement has a memory leak for which memory consumption increases in a linear way while the whole execution time decreases exponentially (this is my empirical conclusion, not actual stats, but the issue was confirmed long time ago by Doctrine _somewhere_). The issue is not impacting the process that much in prod env. Why? I'm not exactly sure, but this is what we usually recommend for all these commands (the text is common).

Steveb-p: > 2\. The issue is not impacting the process that much in prod env. Why? I'm not exactly sure

@alongosz 
v2 has a dedicated logging that is usually enabled when `%kernel.debug%` is `true` (which it is on dev mode). It is a separate logger from the one included in Symfony.

`Doctrine\DBAL\Logging\SQLLogger`

See https://symfony.com/doc/current/reference/configuration/doctrine.html#doctrine-dbal-configuration, look for `logging`.

alongosz: > > 2. The issue is not impacting the process that much in prod env. Why? I'm not exactly sure
> 
> @alongosz v2 has a dedicated logging that is usually enabled when `%kernel.debug%` is `true` (which it is on dev mode). It is a separate logger from the one included in Symfony.
> 
> `Doctrine\DBAL\Logging\SQLLogger`
> 
> See https://symfony.com/doc/current/reference/configuration/doctrine.html#doctrine-dbal-configuration, look for `logging`.

I've been there Steve, disabling it helps, but there's still the PDO mem leak.

PR #348, opened 2022-11-15
Introduced `IbexaKernelTestTrait` to reduce code duplication

Steveb-p:

This PR moves code from `Ibexa\Contracts\Core\Test\IbexaKernelTestCase` to a new trait, `Ibexa\Contracts\Core\Test\IbexaKernelTestTrait`.

This is done to remove code duplication in dependent packages, which happens when one want to use `Symfony\Bundle\FrameworkBundle\Test\WebTestCase` in combination with our Kernel testing features (which requires code from both abstract classes, which is obviously not possible in PHP).

#### Checklist:

### DIFF
--- a/src/contracts/Test/IbexaKernelTestTrait.php
+++ b/src/contracts/Test/IbexaKernelTestTrait.php
@@ -32,7 +32,7 @@ trait IbexaKernelTestTrait
 {
     private static $anonymousUserId = 10;
 
-    private static $adminUserId = 10;
+    private static $adminUserId = 14;
 
     final protected static function loadSchema(): void
     {
--- a/src/contracts/Test/IbexaKernelTestTrait.php
+++ b/src/contracts/Test/IbexaKernelTestTrait.php
@@ -31,6 +31,7 @@ use RuntimeException;
 trait IbexaKernelTestTrait
 {
     private static $anonymousUserId = 10;
+
     private static $adminUserId = 10;
 
     final protected static function loadSchema(): void
--- a/src/contracts/Test/IbexaKernelTestTrait.php
+++ b/src/contracts/Test/IbexaKernelTestTrait.php
@@ -30,10 +30,6 @@ use RuntimeException;
  */
 trait IbexaKernelTestTrait
 {
-    private static $anonymousUserId = 10;
-
-    private static $adminUserId = 14;
-
     final protected static function loadSchema(): void
     {
         $schemaImporter = self::getContainer()->get(LegacySchemaImporter::class);
@@ -169,11 +165,13 @@ trait IbexaKernelTestTrait
 
     protected static function setAnonymousUser(): void
     {
-        self::getPermissionResolver()->setCurrentUserReference(new UserReference(self::$anonymousUserId));
+        $anonymousUserId = 10;
+        self::getPermissionResolver()->setCurrentUserReference(new UserReference($anonymousUserId));
     }
 
     protected static function setAdministratorUser(): void
     {
-        self::getPermissionResolver()->setCurrentUserReference(new UserReference(self::$adminUserId));
+        $adminUserId = 14;
+        self::getPermissionResolver()->setCurrentUserReference(new UserReference($adminUserId));
     }
 }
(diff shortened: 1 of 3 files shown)


alongosz (commented):
Is it possible to have a "service"[1] which performs these operations instead of the trait? The main reason I'm not a huge fan of traits is the risk of overlapping method names (as I call it "loosely coupled strict contract").

[1] These are tests without DI, so I was thinking about sth like:
```php
class FooTest extends WebTestCase
{
    public function setUp(): void
    {
        $repository = new IbexaTestRepository(self::getKernel());
        $repository->loadSchema();
    }
}
```
(yeah, keeping "Repository" naming, for tests it makes sense to me - though not 100% sure about it)

I'm not saying we should do it like that, but I'm just interested what do you think about the alternative. The main goal is to reduce boilerplate code to minium, so if trait gives that, we can go with trait.

Steveb-p:
> The main reason I'm not a huge fan of traits is the risk of overlapping method names (as I call it "loosely coupled strict contract").

In case of a method collision PHP already has tools to resolve them. See [Traits - Conflict Resolution](https://www.php.net/manual/en/language.oop5.traits.php#language.oop5.traits.conflict).

And you are forced to resolve them when they occur, as they are treated as a syntax error, so they will not go unnoticed.

> Is it possible to have a "service"[1] which performs these operations instead of the trait?

Technically we're already using two of our services, so a test-only service is definitely possible and probably would alleviate the issue of two additional methods that currently reside in the TestKernel (`getSchemaFiles()` & `getFixtures()`). These still need to be available in two contexts somehow: during bootstrap process when tests are executed using DAMADoctrineBundle (in transaction) and in `TestCase::setUp()` for those who don't want to bootstrap like that.

Further research is needed for this to happen, but IMO a trait will be necessary nonetheless, purely because it provides common methods for accessing a lot of IbexaTestKernel features that we're using throughout the codebase. And again, the main issue here is that we need these in context of `WebTestCase` at times (not just `KernelTestCase`) which makes abstract unusable, hence trait.

IMO this is a perfect fit, because test cases usually are a result of composition of different features that are needed, and they need to be a single class (so no object composition).

To sum up, the feature requested is currently out of scope and can be considered later on, after solving the immediate issue.

src/contracts/Test/IbexaKernelTestTrait.php
```diff
@@ -0,0 +1,179 @@
...
+use RuntimeException;
+
+/**
+ *  @experimental
+ */
+trait IbexaKernelTestTrait
+{
+    private static $anonymousUserId = 10;
+
+    private static $adminUserId = 10;
```
alongosz:
```suggestion
    private static $adminUserId = 14;
```

alongosz: maybe it would be better to make them public consts in IbexaTestKernel?

alongosz (changes requested):

src/contracts/Test/IbexaKernelTestTrait.php
Steveb-p: Strange. I was sure I fixed that because it broke my tests (obviously).

Steveb-p: I'd rather prefer them to stay here, because if we're gonna go with your suggestion about a bootstrapping service (which it would be effectively) then we would want IbexaTestKernel to contain as little custom code as possible.

We could maybe extract these from Kernel parameters, as I think we've discussed once that those are stored somewhere in ParametersBag?

alongosz (approved):

mikadamczyk (approved):

alongosz (approved):

Steveb-p:
Merged up into `ibexa/core/4.2` in [a1f5916c](https://github.com/ibexa/core/commit/a1f5916c599df77d9fe23e61a355cd88b41d5238).
Merged up into `ibexa/core/main` in [e5905804](https://github.com/ibexa/core/commit/e5905804a874d55c9ae05cc825b6834a65e4bf94).

PR #349, opened 2022-11-15
Added `IbexaTestKernel::addSyntheticService` to reduce duplication

Steveb-p:

`IbexaTestKernel::createSyntheticService()` is available inside multiple of our packages, and seems to be a common thing when a service replacement is necessary due to bundle dependency - and we don't actually need a particular package.

While best solution would be of course to not have those dependencies (or make them optional, but that's personal opinion and goes against our code practices) that's obviously not happening.

To prevent conflict with existing methods I'm adding it with a slightly changed naming, which accidentally also describes a little bit more what is happening, i.e. that a synthetic service definition is actually added do `ContainerBuilder`.

#### Checklist:

### DIFF
--- a/src/contracts/Test/IbexaTestKernel.php
+++ b/src/contracts/Test/IbexaTestKernel.php
@@ -229,4 +229,27 @@ class IbexaTestKernel extends Kernel
     {
         $container->setDefinition('logger', new Definition(NullLogger::class));
     }
+
+    /**
+     * Creates synthetic services in container, allowing compilation of container when some services are missing.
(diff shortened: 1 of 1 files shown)


alongosz (approved):

mikadamczyk (approved):

Steveb-p:
Merged up into `ibexa/core/4.2` in [a1f5916c](https://github.com/ibexa/core/commit/a1f5916c599df77d9fe23e61a355cd88b41d5238).
Merged up into `ibexa/core/main` in [e5905804](https://github.com/ibexa/core/commit/e5905804a874d55c9ae05cc825b6834a65e4bf94).

PR #352, opened 2022-11-24
IBX-146: Fixed `languageCode` resolving for embedded content

konradoboza:
| **JIRA issue**                          | [IBX-146](https://issues.ibexa.co/browse/IBX-146)
| **Target Ibexa version** | `v3.3`, `v4.2`, `v4.3`

This issue presents itself only in AdminUI (frontend rendering is not affected) for embedded images and ImageAsset. It seems we don't operate in the context of content for embeds, so `languageCode` is resolved to `null`. It can be easily fetched from `mainRequest` attributes though.

#### Checklist:

### DIFF
--- a/eZ/Publish/Core/MVC/Symfony/View/Builder/ContentViewBuilder.php
+++ b/eZ/Publish/Core/MVC/Symfony/View/Builder/ContentViewBuilder.php
@@ -124,9 +124,7 @@ class ContentViewBuilder implements ViewBuilder
                 throw new InvalidArgumentException('Content', 'Could not load any content from the parameters');
             }
 
-            $mainRequest = $this->requestStack->getMainRequest();
-            $requestLanguageCode = $mainRequest === null ? null : $mainRequest->attributes->get('languageCode');
-            $languageCode = $parameters['languageCode'] ?? $requestLanguageCode;
+            $languageCode = $parameters['languageCode'] ?? $this->resolveMainRequestLanguageCode();
 
             $content = $view->isEmbed() ? $this->loadEmbeddedContent($contentId, $location, $languageCode) : $this->loadContent($contentId, $languageCode);
         }
@@ -168,6 +166,13 @@ class ContentViewBuilder implements ViewBuilder
         $this->viewConfigurator->configure($view);
 
         return $view;
+    }
+
+    private function resolveMainRequestLanguageCode(): ?string
+    {
+        $mainRequest = $this->requestStack->getMainRequest();
+
+        return $mainRequest === null ? null : $mainRequest->attributes->get('languageCode');
     }
 
     /**
--- a/eZ/Publish/Core/MVC/Symfony/View/Builder/ContentViewBuilder.php
+++ b/eZ/Publish/Core/MVC/Symfony/View/Builder/ContentViewBuilder.php
@@ -125,7 +125,7 @@ class ContentViewBuilder implements ViewBuilder
             }
 
             $mainRequest = $this->requestStack->getMainRequest();
-            $requestLanguageCode = $mainRequest === null ? null : $mainRequest->get('languageCode');
+            $requestLanguageCode = $mainRequest === null ? null : $mainRequest->attributes->get('languageCode');
             $languageCode = $parameters['languageCode'] ?? $requestLanguageCode;
 
             $content = $view->isEmbed() ? $this->loadEmbeddedContent($contentId, $location, $languageCode) : $this->loadContent($contentId, $languageCode);
--- a/eZ/Publish/Core/MVC/Symfony/View/Builder/ContentViewBuilder.php
+++ b/eZ/Publish/Core/MVC/Symfony/View/Builder/ContentViewBuilder.php
@@ -124,7 +124,9 @@ class ContentViewBuilder implements ViewBuilder
                 throw new InvalidArgumentException('Content', 'Could not load any content from the parameters');
             }
 
-            $languageCode = $parameters['languageCode'] ?? null;
+            $mainRequest = $this->requestStack->getMainRequest();
+            $requestLanguageCode = $mainRequest === null ? null : $mainRequest->get('languageCode');
+            $languageCode = $parameters['languageCode'] ?? $requestLanguageCode;
 
             $content = $view->isEmbed() ? $this->loadEmbeddedContent($contentId, $location, $languageCode) : $this->loadContent($contentId, $languageCode);
         }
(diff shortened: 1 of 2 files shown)


eZ/Publish/Core/MVC/Symfony/View/Tests/Builder/ContentViewBuilderTest.php
```diff
@@ -296,8 +381,6 @@ public function testBuildViewWithTranslatedContentWithoutLocation(): void
             ->method('getContentService')
             ->willReturn($contentServiceMock);
 
-        $this->contentViewBuilder->buildView($parameters);
```
ViniTou, line 299: Removing that line seems kinda random?

konradoboza: Actually, it doesn't need to be called twice, few lines below we have:
`$this->assertEquals($expectedView, $this->contentViewBuilder->buildView($parameters));`.

eZ/Publish/Core/MVC/Symfony/View/Tests/Builder/ContentViewBuilderTest.php
```diff
@@ -221,7 +224,89 @@ public function testBuildEmbedViewWithoutContentViewEmbedPermission(): void
...
+        $attributes
+            ->expects(self::once())
+            ->method('get')
+            ->with('languageCode')
+            ->willReturn($languageCode);
+
+        $request = new Request();
+        $reflectionClass = new ReflectionClass($request);
+        $reflectionProperty = $reflectionClass->getProperty('attributes');
+        $reflectionProperty->setValue($request, $attributes);
```
Steveb-p, line 289: Why not simply:
```suggestion
        $request = new Request([], [], $attributes);
```
instead?

_EDIT: and for PHP 8.x:_
```suggestion
        $request = new Request(attributes: $attributes);
```

eZ/Publish/Core/MVC/Symfony/View/Tests/Builder/ContentViewBuilderTest.php
```diff
@@ -221,7 +224,89 @@ public function testBuildEmbedViewWithoutContentViewEmbedPermission(): void
...
+            ->method('getMainRequest')
+            ->willReturn(null);
+
+        $this->repository
+            ->expects(self::once())
+            ->method('sudo')
+            ->willReturn($content);
+
+        $this->permissionResolver
+            ->expects(self::any())
```
Steveb-p: Either declare a specific amount of calls, or do not specify `expects` at all.
```suggestion
```

eZ/Publish/Core/MVC/Symfony/View/Tests/Builder/ContentViewBuilderTest.php
```diff
@@ -221,7 +224,89 @@ public function testBuildEmbedViewWithoutContentViewEmbedPermission(): void
...
+            ->method('getMainRequest')
+            ->willReturn($request);
+
+        $this->repository
+            ->expects(self::once())
+            ->method('sudo')
+            ->willReturn($content);
+
+        $this->permissionResolver
+            ->expects(self::any())
```
Steveb-p:
```suggestion
```

eZ/Publish/Core/MVC/Symfony/View/Builder/ContentViewBuilder.php
```diff
@@ -124,7 +124,9 @@ public function buildView(array $parameters)
                 throw new InvalidArgumentException('Content', 'Could not load any content from the parameters');
             }
 
-            $languageCode = $parameters['languageCode'] ?? null;
+            $mainRequest = $this->requestStack->getMainRequest();
+            $requestLanguageCode = $mainRequest === null ? null : $mainRequest->attributes->get('languageCode');
+            $languageCode = $parameters['languageCode'] ?? $requestLanguageCode;
```
Steveb-p: If parameters has `languageCode`, then there is no reason to get it from the `Request::attributes`. Move it into a method and change the order of operations.

eZ/Publish/Core/MVC/Symfony/View/Tests/Builder/ContentViewBuilderTest.php
konradoboza: Public property `$attributes` is actually a `parameterBag` which I mocked above. The idea was to test `$request->attributes->get('language')` invocation.

Steveb-p: You cannot test the output of the `buildView` method instead?

It seems awfully specific to test all those calls inside the service, and expect that fact alone to have an impact on the end result. Specifically, there is no guarantee based on these assertions that the language used for the content block rendered is the one that you will be passing.

You're only testing that services are called, not that whatever they return is used.

alongosz: > You cannot test the output of the `buildView` method instead?
> 
> It seems awfully specific to test all those calls inside the service, and expect that fact alone to have an impact on the end result. Specifically, there is no guarantee based on these assertions that the language used for the content block rendered is the one that you will be passing.
> 
> You're only testing that services are called, not that whatever they return is used.

Indeed we need to think here about better way of testing this, without a need for Reflection.

konradoboza: Due to the fact that most objects are mocks and retrieving content takes place in a private method this might become challenging. Maybe some integration test should be composed in order to verify whether all the needed parameters are there?

webhdx (approved):

ViniTou (approved):

ciastektk (approved):

Steveb-p (approved):

alongosz (approved):

eZ/Publish/Core/MVC/Symfony/View/Tests/Builder/ContentViewBuilderTest.php
konradoboza: We decided within a team that proposed test coverage will do at the moment. We are lacking integration test coverage for that part of the system in general and implemented test cases are in-tact with how `ContentViewBuilder` is unit-tested for the time being.

katarzynazawada (approved):

PR #354, opened 2022-12-14
IBX-4581: Added missing groups setup when siteaccess is matched

ViniTou:
| **JIRA issue**                          | [IBX-4581](https://issues.ibexa.co/browse/IBX-4581)

I also fixed/added groups serialization, it was no related to problem with `getCurrent()` itself, but it seemed related and could save as a problem in the future.

#### Checklist:

### DIFF
--- a/eZ/Publish/Core/MVC/Symfony/EventListener/SiteAccessMatchListener.php
+++ b/eZ/Publish/Core/MVC/Symfony/EventListener/SiteAccessMatchListener.php
@@ -15,6 +15,7 @@ use eZ\Publish\Core\MVC\Symfony\MVCEvents;
 use eZ\Publish\Core\MVC\Symfony\Routing\SimplifiedRequest;
 use eZ\Publish\Core\MVC\Symfony\SiteAccess;
 use eZ\Publish\Core\MVC\Symfony\SiteAccess\Router as SiteAccessRouter;
+use eZ\Publish\Core\MVC\Symfony\SiteAccessGroup;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\Request;
@@ -80,7 +81,9 @@ class SiteAccessMatchListener implements EventSubscriberInterface
                     $request->attributes->get('serialized_siteaccess_sub_matchers')
                 );
             }
-
+            $siteAccess->groups = $this->buildGroups(
+                $siteAccess->groups
+            );
             $request->attributes->set(
                 'siteaccess',
                 $siteAccess
@@ -164,6 +167,20 @@ class SiteAccessMatchListener implements EventSubscriberInterface
         }
 
         return $matcher;
+    }
+
+    /**
+     * @return \eZ\Publish\Core\MVC\Symfony\SiteAccessGroup[]
+     */
+    private function buildGroups(
+        array $serializedGroups
+    ): array {
+        return array_map(
+            static function (array $serializedGroup): SiteAccessGroup {
+                return new SiteAccessGroup($serializedGroup['name']);
+            },
+            $serializedGroups
+        );
     }
 
     /**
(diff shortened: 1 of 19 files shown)


ViniTou:
Close & reopen to force github actions

eZ/Publish/Core/MVC/Symfony/SiteAccessGroup.php
```diff
@@ -8,7 +8,9 @@
 
 namespace eZ\Publish\Core\MVC\Symfony;
 
-final class SiteAccessGroup
+use JsonSerializable;
+
+final class SiteAccessGroup implements JsonSerializable
```
Steveb-p, line 13: Both locations that perform `SiteAccess` serialization for the purposes of storing them in `Request` attributes object have access to the serializer service. I think that using it is preferable over maintaining encoding (in this file) and decoding (in the listener).

ViniTou: @Steveb-p 
sorry for late replay, 
true, but:
If someone is relying on `JsonSerializable` and current state of things, he would get wrongly serialized object. So that's need to be added/fixed as well, and as a followup/additional step it could be redone to use Serializer, right?

adamwojs: AFAIK n this case we always  rely on native PHP serialization nad later on symfony/serializer :wink:

adamwojs (approved):

kisztof (approved):

eZ/Publish/Core/MVC/Symfony/EventListener/SiteAccessMatchListener.php
```diff
@@ -166,6 +169,20 @@ private function deserializeMatcher(
         return $matcher;
     }
 
+    /**
+     * @return \eZ\Publish\Core\MVC\Symfony\SiteAccessGroup[]
```
konradoboza, line 175: Missing docblock for `$serializedGroups` param.

eZ/Publish/Core/MVC/Symfony/SiteAccessGroup.php
```diff
@@ -27,4 +29,11 @@ public function __toString()
     {
         return $this->name;
     }
+
+    public function jsonSerialize(): array
```
konradoboza, line 36:
```suggestion
    /**
     * @return array<string, string>
     */
    public function jsonSerialize(): array
```

konradoboza (approved):

eZ/Publish/Core/MVC/Symfony/SiteAccessGroup.php
Steveb-p: > If someone is relying on JsonSerializable and current state of things

Currently calling `json_encode` on this object is an error (if not straight up PHP error, it is a logical error - this object was not supposed to be json encoded this way).

Adding this interface is only acceptable if we are treating this change as internal only, and only to make it slightly faster to implement. However, I would say that a proper normalizer should be added instead, as the object should not be aware how it is presented in REST (I assume that's the target).

alongosz (approved):

alongosz:
Seems it's worth to send this to QA due the bug manifesting itself in a couple of places. It indeed requires creating some PHP classes, as does not occur in the backoffice. @ezsystems/qa-team please see JIRA issue for the detailed information on how this can be tested.

micszo (approved):
Retested successfully on Ibexa Commerce 3.3.30-dev.

Using `getCurrent` now returns groups in the application in the browser.
CLI still doesn't, that will have to be resolved in another iteration.

> the bug manifesting itself in a couple of places

So @alongosz you were on to something 😉 
Also thank you for the extensive details and the code! 🖤

alongosz:
Merging.

_SonarCloud failure seems like false negative and will show up again if someone touches these specific files. We might consider in the future making some shared data provider for all these test cases if possible (TL;DR)._

PR #362, opened 2023-02-09
[Composer] Disallowed every not specified Composer plugin

alongosz:
| **JIRA issue**                          | n/a

Seems indirect dependency `php-http/discovery` introduces Composer plugin for discovering proper HTTPlug client
```
$ composer why php-http/discovery
friendsofsymfony/http-cache       2.15.0 requires php-http/discovery (^1.12) 
knplabs/github-api                v3.9.0 requires php-http/discovery (^1.12) 
php-http/multipart-stream-builder 1.2.0  requires php-http/discovery (^1.7)
```

This needs to be handled in composer.json so CI test jobs won't fail unlike [this one](https://github.com/ezsystems/ezplatform-kernel/actions/runs/4135434711/jobs/7147881743#step:4:242).

Given risky approach by 3rd party vendors to Composer plugins, causing BC breaks, we've discovered that we can use either
```json
"config": {
        "allow-plugins": {
            "*": false
        }
    }
```
to disallow everything except what is already allowed, or simpler:
```json
"config": {
        "allow-plugins": false
    }
```
if there's no such key yet in `composer.json`

to avoid such issues in the future. If a plugin is needed then it can be explicitly enabled.

#### Checklist:
- [x] Asked for a review

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -115,7 +115,7 @@
         "process-timeout": 3000,
         "allow-plugins": {
             "composer/package-versions-deprecated": true,
-            "php-http/discovery": true
+            "*/*": false
         }
     },
     "scripts": {
--- a/composer.json
+++ b/composer.json
@@ -115,7 +115,7 @@
         "process-timeout": 3000,
         "allow-plugins": {
             "composer/package-versions-deprecated": true,
-            "*/*": false
+            "*": false
         }
     },
     "scripts": {
--- a/composer.json
+++ b/composer.json
@@ -114,7 +114,8 @@
     "config": {
         "process-timeout": 3000,
         "allow-plugins": {
-            "composer/package-versions-deprecated": true
+            "composer/package-versions-deprecated": true,
+            "php-http/discovery": true
         }
     },
     "scripts": {


konradoboza (approved):

composer.json
```diff
@@ -114,7 +114,8 @@
     "config": {
         "process-timeout": 3000,
         "allow-plugins": {
-            "composer/package-versions-deprecated": true
+            "composer/package-versions-deprecated": true,
+            "php-http/discovery": true
```
mnocon:
```suggestion
            "*/*": false
```

Disclaimer: I haven't tested if it works, but I like this direction more.

Hopefully it would both enable installation and save us from other "helpful" 3rd party plugins in the future

(I don't feel like we actually need this new plugin to be enabled)

alongosz: Looks like it works, thanks!

konradoboza (approved):

mnocon (approved):
Thanks!

kisztof (approved):

composer.json
```diff
@@ -114,7 +114,8 @@
     "config": {
         "process-timeout": 3000,
         "allow-plugins": {
-            "composer/package-versions-deprecated": true
+            "composer/package-versions-deprecated": true,
+            "*/*": false
```
Steveb-p: Works too :)
```suggestion
            "*": false
```

Steveb-p (approved):

Steveb-p:
Additionally, if/when `composer` introduces "optional plugins" (which is discussed in the issue report that this PR originated from) we should switch back hard fails IMO, because some plugins might indeed be mandatory.

Nattfarinn (approved):

PR #363, opened 2023-02-20
IBX-3972: Fixed password validation for UserService::updateUserPassword

alongosz:
| **JIRA issue**                          | [IBX-3972](https://issues.ibexa.co/browse/IBX-3972)
| **Related PRs**                         | ezsystems/ezplatform-user#113, ezsystems/ezplatform-admin-ui#2094

This PR refactors `PasswordValidator` service to use all validators, including the one which determines if old password was re-used. The responsibility of password validation has been moved from `UserService` to `PasswordValidator` itself. `UserService` now entirely delegates password validation to `PasswordValidator`.

`UserService::getPasswordInfo` method has also been delegated to `PasswordValidator`. The method relies on `FieldDefinition::$fieldSettings`, but it contains mostly validation logic.

It needs to be pointed out that `UserService::updateUserPassword` currently throws `ContentFieldValidationException` while in the context of the user account password itself some more specific wrapper containing `ValidationError` list would be more appropriate. `ContentFieldValidationException` expects `ValidationError` list in the format
`[<fieldId> => [<laguageCode> => array<ValidationError>]]` which doesn't make much sense in the context. Due to BC this however cannot be simply changed. What was fixed was the format of `$errors` array so it follows that API specification. It might be worth to mention in the release notes.

The PR also introduces alignment of tests and fix for processing `ContentFieldValidationException` by test setup.

## Documentation

`UserService::updateUserPassword` PHP API throws now `ContentFieldValidationException` with proper Field Errors array format accessible via `ContentFieldValidationException::getFieldErrors` method. The format of the returned array should always be
```
array<<int fieldId>, array<<string language code>, array<ValidationError>>
```
This is BC break on behavior, but it fixes promise on API specification. It doesn't undergo BC policy, but we definitely should mention this in Release Notes and/or upgrade instruction. Keep in mind this is a bugfix so it's gonna be a part of next patch release (3.3.x), not minor release.

## QA

For the part related to kernel and PHP API `UserService::updateUserPassword` should now correctly throw `ContentFieldValidationException` for the case when reusing old password was forbidden but old password was reused.

For the CR fix part see ezsystems/ezplatform-user#113.

#### Checklist:

### DIFF
--- a/eZ/Publish/Core/Repository/User/PasswordValidator.php
+++ b/eZ/Publish/Core/Repository/User/PasswordValidator.php
@@ -8,7 +8,15 @@ declare(strict_types=1);
 
 namespace eZ\Publish\Core\Repository\User;
 
+use DateInterval;
+use DateTime;
+use DateTimeImmutable;
 use eZ\Publish\API\Repository\Values\ContentType\FieldDefinition;
+use eZ\Publish\API\Repository\Values\User\PasswordInfo;
+use eZ\Publish\API\Repository\Values\User\User;
+use eZ\Publish\API\Repository\Values\User\User as APIUser;
+use eZ\Publish\Core\FieldType\User\Type as UserType;
+use eZ\Publish\Core\FieldType\ValidationError;
 use eZ\Publish\Core\Repository\Validator\UserPasswordValidator;
 
 /**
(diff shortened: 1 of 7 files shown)


kisztof (approved):

eZ/Publish/Core/Repository/User/PasswordValidator.php
```diff
@@ -8,24 +8,109 @@
...
-    public function validatePassword(string $password, FieldDefinition $userFieldDefinition): array
-    {
+    public function validatePassword(
+        string $password,
+        FieldDefinition $userFieldDefinition,
+        ?User $user = null
+    ): array {
         $configuration = $userFieldDefinition->getValidatorConfiguration();
         if (!isset($configuration['PasswordValueValidator'])) {
             return [];
```
glye, line 45: This is existing code and off topic, but still: If no validator configuration is set, it returns empty array, i.e. no errors, the password is considered valid for storage. That's logical, but is there any actual use case (outside of dev) where it makes sense to have zero password requirements? Sure, our password validation UI may allow disabling all validation, but should it? Or should we set some absolute minimum requirement, like length? 8 chars is the current NIST minimum: https://en.wikipedia.org/wiki/Password_policy

The industry has moved away from mandating detailed requirements, but some minimum requirements are still considered sensible. My guess is some clients may not like that's possible to configure it into an unsafe state. We could allow it, but show a warning message if you do: "This is unsafe, are you sure?" Of course it will always be possible to configure _some_ aspect of the system stack into an unsafe state, that's unavoidable. Anyway, it doesn't belong in this PR, but it would be nice to hear what others think.

glye (approved):
Seems a sensible solution to me.

eZ/Publish/Core/Repository/User/PasswordValidator.php
alongosz: We provide good zero-config - there are password rules required which can be disabled. We shouldn't cripple flexibility of the system. If you decide to use no requirements - you should be allowed at your own risk.

glye: Yes, there's a default config, and it might even be too rigid. We could argue it shouldn't demand upper/lower case and numbers, length is enough. But I think it shouldn't be possible to disable everything without at least a warning, if length is set below 8. I'll see if I can do a PR on that at some point.

ViniTou (approved):

konradoboza (approved):

mateuszdebinski (approved):

micszo (approved):

PR #367, opened 2023-03-13
IBX-5270: Implemented cache indices validator and fixed broken `loadSubtreeIds` pattern

barw4:
| **JIRA issue**                          | [IBX-5270](https://issues.ibexa.co/browse/IBX-5270)

A newly created `CacheIndicesValidator` will validate whether some of an object's cache indices are not corresponding to a provided prefix and log this later on.

The mentioned validator caught a problem with the `LocationHandler` -> `loadSubtreeIds()` method so this method has been fixed as well. 

Exemplary logged message: 
`[2023-03-13T15:34:39.999000+01:00] ibexa.core.ERROR: There is no corresponding cache index for key prefix page-. Cache indexes are as follows: ez-page-218-2-eng-GB. [] []`

Related PR: https://github.com/ezsystems/ezplatform-page-fieldtype/pull/245

#### Checklist:


eZ/Publish/Core/Persistence/Cache/PersistenceLogger.php
```diff
@@ -249,4 +257,18 @@ public function getLoadedUnCachedHandlers(): array
     {
         return $this->unCachedHandlers;
     }
+
+    /**
+     * @param array<string> $cacheIndices
+     */
+    public function logNoIndexAndPrefixMatch(array $cacheIndices, string $keyPrefix): void
```
Steveb-p: This method seems unused.

src/lib/Persistence/Cache/CacheIndicesValidator.php
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * @internal
+ */
+final class CacheIndicesValidator implements CacheIndicesValidatorInterface, LoggerAwareInterface
+{
+    use LoggerAwareTrait;
+
+    public function __construct(?LoggerInterface $logger = null)
+    {
+        $this->logger = $logger ?? new NullLogger();
```
Steveb-p: Since the essence of this class is to actually interact with logger, I'd personally skip initialization to `NullLogger`, and instead return early in `validate` method if the logger is `null`.

eZ/Publish/Core/Persistence/Cache/Tests/AbstractBaseHandlerTest.php
```diff
@@ -67,6 +68,9 @@ abstract class AbstractBaseHandlerTest extends TestCase
     /** @var \Ibexa\Core\Persistence\Cache\LocationPathConverter */
     protected $locationPathConverter;
 
+    /** @var \Ibexa\Core\Persistence\Cache\CacheIndicesValidatorInterface */
+    protected $cacheIndicesValidator;
```
Steveb-p, line 72: Note: Tests are unlikely to interact with this property (since mock never returns a value, and does not throw exceptions).

eZ/Publish/Core/Persistence/Cache/AbstractInMemoryHandler.php
```diff
@@ -99,6 +98,9 @@ final protected function getCacheValue(
         $this->logger->logCacheMiss($arguments ?: [$id], 3);
 
         $object = $backendLoader($id);
+
+        $this->cacheIndicesValidator->validate($keyPrefix, $object, $cacheIndexes);
```
adamwojs: Please ensure this logic is executed only if kernel is in debug mode

src/lib/Persistence/Cache/CacheIndicesValidator.php
```diff
@@ -0,0 +1,50 @@
...
+
+        foreach ($cacheIndicesUnpacked as $cacheIndex) {
+            if (strpos($cacheIndex, $keyPrefix) === 0) {
+                return;
+            }
+        }
+
+        $this->logger->error(
+            sprintf(
+                'There is no corresponding cache index for key prefix %s. Cache indexes are as follows: %s.',
```
konradoboza: To keep the adopted naming consistency.
```suggestion
                'There is no corresponding cache index for key prefix %s. Cache indices are as follows: %s.',
```

eZ/Publish/Core/Persistence/Cache/AbstractInMemoryHandler.php
Steveb-p: Then may I suggest we only create this `CacheIndicesValidator` in debug mode, and treat it as an optional dependency?

src/lib/Persistence/Cache/CacheIndicesValidator.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
```
konradoboza, line 6:
```suggestion
 */
 declare(strict_types=1);
 
```

src/lib/Persistence/Cache/CacheIndicesValidatorInterface.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace Ibexa\Core\Persistence\Cache;
```
konradoboza, line 9:
```suggestion
declare(strict_types=1);

namespace Ibexa\Core\Persistence\Cache;
```

eZ/Publish/Core/Persistence/Cache/AbstractInMemoryHandler.php
barw4: @adamwojs @Steveb-p `CacheIndicesValidator` created only in debug mode as of now: https://github.com/ezsystems/ezplatform-kernel/pull/367/commits/18dcebca9725baa4452b1eac7a25b9d8d51b2773

barw4: Also, updated https://github.com/ezsystems/ezplatform-page-fieldtype/pull/245 to match the CompilerPass solution

konradoboza (approved):

src/bundle/Core/DependencyInjection/Compiler/CacheIndicesValidatorPass.php
```diff
@@ -0,0 +1,61 @@
...
+        $abstractInMemoryHandlerDefinition->setArgument(
+            '$cacheIndicesValidator',
+            new Reference(CacheIndicesValidatorInterface::class)
+        );
+        $abstractInMemoryPersistenceHandlerDefinition = $container
+            ->getDefinition('ezpublish.spi.persistence.cache.abstractInMemoryPersistenceHandler');
+        $abstractInMemoryPersistenceHandlerDefinition->setArgument(
+            '$cacheIndicesValidator',
+            new Reference(CacheIndicesValidatorInterface::class)
+        );
```
adamwojs: You should be able to inject cache validator using https://symfony.com/doc/current/service_container/optional_dependencies.html

src/bundle/Core/DependencyInjection/Compiler/CacheIndicesValidatorPass.php
```diff
@@ -0,0 +1,61 @@
...
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+
+final class CacheIndicesValidatorPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
```
adamwojs:
```suggestion
    public function process(ContainerBuilder $container): void
```

src/bundle/Core/DependencyInjection/Compiler/CacheIndicesValidatorPass.php
```diff
@@ -0,0 +1,61 @@
...
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+
+final class CacheIndicesValidatorPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
```
adamwojs:
```suggestion
```

src/bundle/Core/DependencyInjection/Compiler/CacheIndicesValidatorPass.php
barw4: As a side note, shouldn't we keep with the interface's definition?

src/bundle/Core/DependencyInjection/Compiler/CacheIndicesValidatorPass.php
Steveb-p, line 1: The expected approach for this is to declare those services in a separate file and load them conditionally in the extension. 

Like so:
```php
# IbexaCoreExtension.php
if ($container->getParameter('kernel.debug') === true) {
    $loader->load('cache_validator.yaml');
}
```

It should be way easier to maintain when those services are declared in yaml format.

src/bundle/Core/DependencyInjection/Compiler/CacheIndicesValidatorPass.php
Steveb-p: In this case we are well within the expected interface shape. In PHP you can (and often should) narrow down the return type (it's called covariance: https://www.php.net/manual/en/language.oop5.variance.php).

While full covariance was introduced in PHP 7.4, methods without types can have their return type added if there wasn't any before (in interface/parent). Should probably be checked though.

barw4: Thank you

src/bundle/Core/DependencyInjection/Compiler/CacheIndicesValidatorPass.php
alongosz: > The expected approach for this is to declare those services in a separate file and load them conditionally in the extension.
> 
> Like so:
> 
> ```
> # IbexaCoreExtension.php
> if ($container->getParameter('kernel.debug')) {
>     $loader->load('cache_validator.yaml');
> }
> ```
> 
> It should be way easier to maintain when those services are declared in yaml format.

@Steveb-p Shouldn't we then place `cache_validator.yaml` in some `debug` directory so it's also clear for someone who looks over the configuration?

eZ/Bundle/EzPublishCoreBundle/DependencyInjection/EzPublishCoreExtension.php
```diff
@@ -399,6 +399,11 @@ private function handleCache(array $config, ContainerBuilder $container, FileLoa
 
             $container->setParameter('ezpublish.http_cache.purge_type', $purgeType);
         }
+
+        $debugParam = 'kernel.debug';
```
alongosz: Good idea to extract it, but this is rather const, not a variable.

alongosz (approved):

adamwojs (approved):

webhdx (approved):

konradoboza (approved):

ciastektk (approved):

Steveb-p (approved):

eZ/Bundle/EzPublishCoreBundle/Resources/config/debug/cache_validator.yaml
Steveb-p, line 1: Small note, we tend to stick to 4 spaces as indentation in .yaml files. Not important enough to block merging, and there are arguments to use 2 spaces (I would even prefer that myself), but that is the current standard everywhere.

micszo (approved):
QA Approved on Ibexa Experience 3.3.32-dev.

PR #368, opened 2023-04-06
IBX-5068: Fixed mapping fields for update to account for language

barw4:
| **JIRA issue**                          | [IBX-5068](https://issues.ibexa.co/browse/IBX-5068)

Before this PR fields were fetched in a main language - that was the cause for adding an additional field to update even if values in concurrent languages were the same.

#### Checklist:

### DIFF
--- a/eZ/Publish/API/Repository/Values/Content/Field.php
+++ b/eZ/Publish/API/Repository/Values/Content/Field.php
@@ -47,7 +47,7 @@ class Field extends ValueObject
     /**
      * the language code.
      *
-     * @var string
+     * @var string|null
      */
     protected $languageCode;
 
@@ -76,7 +76,7 @@ class Field extends ValueObject
         return $this->value;
     }
 
-    public function getLanguageCode(): string
+    public function getLanguageCode(): ?string
     {
         return $this->languageCode;
     }
--- a/eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php
+++ b/eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php
@@ -1030,7 +1030,7 @@ class ContentTest extends BaseServiceMockTest
             ->with(42)
             ->will($this->returnValue(['version']));
 
-        /* @var APIVersionInfo $versionInfo */
+        /* @var \eZ\Publish\API\Repository\Values\Content\VersionInfo $versionInfo */
         $contentService->deleteVersion($versionInfo);
     }
 
@@ -5849,7 +5849,7 @@ class ContentTest extends BaseServiceMockTest
             ->with($spiVersionInfo)
             ->will($this->returnValue($versionInfoMock));
 
-        /* @var APIVersionInfo $versionInfoMock */
+        /* @var \eZ\Publish\API\Repository\Values\Content\VersionInfo $versionInfoMock */
         $content = $this->mockPublishVersion(123456, 126666, true);
         $locationServiceMock->expects($this->once())
             ->method('createLocation')
--- a/eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php
+++ b/eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php
@@ -27,6 +27,8 @@ use eZ\Publish\API\Repository\Values\ContentType\FieldDefinition as APIFieldDefi
 use eZ\Publish\Core\Base\Exceptions\ContentFieldValidationException;
 use eZ\Publish\Core\Base\Exceptions\ContentValidationException;
 use eZ\Publish\Core\Base\Exceptions\NotFoundException;
+use eZ\Publish\Core\FieldType\TextLine\Type;
+use eZ\Publish\Core\FieldType\TextLine\Value as TextLineValue;
 use eZ\Publish\Core\FieldType\ValidationError;
 use eZ\Publish\Core\FieldType\Value;
 use eZ\Publish\Core\Repository\ContentService;
@@ -45,6 +47,7 @@ use eZ\Publish\Core\Repository\Values\User\UserReference;
 use eZ\Publish\SPI\FieldType\FieldType;
 use eZ\Publish\SPI\FieldType\FieldType as SPIFieldType;
 use eZ\Publish\SPI\FieldType\Value as SPIValue;
+use eZ\Publish\SPI\Limitation\Target\Builder\VersionBuilder;
 use eZ\Publish\SPI\Persistence\Content as SPIContent;
 use eZ\Publish\SPI\Persistence\Content\ContentInfo as SPIContentInfo;
 use eZ\Publish\SPI\Persistence\Content\CreateStruct as SPIContentCreateStruct;
@@ -1027,7 +1030,7 @@ class ContentTest extends BaseServiceMockTest
             ->with(42)
             ->will($this->returnValue(['version']));
 
-        /* @var \eZ\Publish\API\Repository\Values\Content\VersionInfo $versionInfo */
+        /* @var APIVersionInfo $versionInfo */
         $contentService->deleteVersion($versionInfo);
     }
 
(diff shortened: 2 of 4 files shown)


eZ/Publish/API/Repository/Values/Content/Field.php
```diff
@@ -76,7 +76,7 @@ public function getValue()
         return $this->value;
     }
 
-    public function getLanguageCode(): string
+    public function getLanguageCode(): ?string
```
alongosz, line 79: This is a BC break. You cannot change promise of strict language code into a "maybe a code, maybe a null". What is the actual issue here? Is the language code not set?

eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php
```diff
@@ -1027,7 +1030,7 @@ public function testDeleteVersionThrowsBadStateExceptionLastVersion()
             ->with(42)
             ->will($this->returnValue(['version']));
 
-        /* @var \eZ\Publish\API\Repository\Values\Content\VersionInfo $versionInfo */
+        /* @var APIVersionInfo $versionInfo */
```
alongosz: CS: we use FQCN in var PHPDoc

eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php
```diff
@@ -3170,6 +3173,109 @@ public function testUpdateContentThrowsUnauthorizedException()
         $mockedService->updateContent($versionInfo, $contentUpdateStruct);
     }
 
+    /**
+     * @covers \eZ\Publish\Core\Repository\ContentService::updateContent
+     */
+    public function testUpdateContentGetsProperFieldsToUpdate(): void
```
alongosz: This test doesn't bring enough value. What we need here is an integration test. Looking at the issue steps to reproduce writing such test should be straightforward.

eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php
```diff
@@ -3170,6 +3173,109 @@ public function testUpdateContentThrowsUnauthorizedException()
...
+                        ->updateFieldsTo(
+                            $contentUpdateStruct->initialLanguageCode,
+                            $contentUpdateStruct->fields
+                        )
+                        ->build(),
+                ]
+            )
+            ->willReturn(false);
+
+        $this->expectException(UnauthorizedException::class);
```
alongosz: wasn't the issue that you're actually getting an exception and you shouldn't?

eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php
```diff
@@ -5743,7 +5849,7 @@ public function testCopyContent()
             ->with($spiVersionInfo)
             ->will($this->returnValue($versionInfoMock));
 
-        /* @var \eZ\Publish\API\Repository\Values\Content\VersionInfo $versionInfoMock */
+        /* @var APIVersionInfo $versionInfoMock */
```
alongosz: Why do you feel you need to change those? They were ok.

alongosz (changes requested):

eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php
barw4: this is a direct result of CS-fixer

eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php
barw4: I'm really not sure why would we want to test an edge case. The real problem is that `getFieldsForUpdate` is invalid. The fact that we have an exception is just an edge case resulting from the original problem which is why fields to update are unnecessarily added. Isn't this rather a problem with the `ContentMapper` which is what I'm testing here?

eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php
barw4: Please, see above. This test rather tests `ContentMapper` as it's apparently broken. It validates fields destined to be updated, not whether we should have an exception from the different package (`ezplatform-permissions`). The exception thrown is irrelevant here.

eZ/Publish/API/Repository/Values/Content/Field.php
barw4: Well, apparently it can be `null`: 
https://github.com/ezsystems/ezplatform-kernel/blob/6c5d6e44399c073e2954baf8e687cd70721c8d63/eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php#L3863
Are data providers invalid then?

There is really no issue besides failing tests, looking at data providers it sure can be `null`.

eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php
barw4: Please, note that Fieldgroup limitation is outside of `ezplatform-kernel` package. How would an integration test apply to kernel then if we are doing a change to the kernel itself but we can test it only in `ezplatform-permissions` package?

barw4: I would expect that package I'm updating has a representation in a unit test in the exact same package.

ViniTou: >I'm really not sure why would we want to test an edge case. 

Outside of the PR, but one of the reasons is that to not introduce bug again when code is touched to fix another edge case or refactored.

barw4: That's why I've actually tested it in `Mock/ContentTest`. As mentioned below, `FieldGroupLimitation` is outside of this package and it's not the real problem here.

eZ/Publish/API/Repository/Values/Content/Field.php
alongosz: > Well, apparently it can be `null`:
> 
> https://github.com/ezsystems/ezplatform-kernel/blob/6c5d6e44399c073e2954baf8e687cd70721c8d63/eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php#L3863
> 
> 
> Are data providers invalid then?
> There is really no issue besides failing tests, looking at data providers it sure can be `null`.

Usually released API contract is more important that data provider for test mocks. Changing API has implications on 1st and 3rd party code. If someone relies on that method to return `string` and suddenly gets `null` after the update, it's an issue. This change might also affect our other packages when it comes to static analysis. 

That said, **you're raising a valid point** :muscle: - is it a bug on the API level, or is there a bug in the provider - which one is valid? If it's a bug on the API level then all I've said about contract doesn't matter because it is already crashing for 1st or 3rd party code, maybe undiscovered yet.

Let's see the usages:

- [ContentCreateStruct, if no `languageCode` given, uses fallback](https://github.com/ezsystems/ezplatform-kernel/blob/94937f8db831eb4229686f6cfde92e052a658b11/eZ/Publish/Core/Repository/Values/Content/ContentCreateStruct.php#L44-L46)
- [ContentUpdateStruct has no fallback and actually sets null if null given](https://github.com/ezsystems/ezplatform-kernel/blob/94937f8db831eb4229686f6cfde92e052a658b11/eZ/Publish/Core/Repository/Values/Content/ContentUpdateStruct.php#L40-L46)

Let's see how this behaves on an integration level:
```php
$contentUpdateStruct = $contentService->newContentUpdateStruct();

$contentUpdateStruct->setField('name', 'Foo');
self::assertInstanceOf(Field::class, $contentUpdateStruct->fields[0]);

self::assertNull($contentUpdateStruct->fields[0]->getLanguageCode());
```
```
TypeError : Return value of eZ\Publish\API\Repository\Values\Content\Field::getLanguageCode() must be of the type string, null returned
```

Ideally we should use fallback language code, so the API is strict. However I don't see here a clean way to do it, as there's no connection between the Update Struct and a Content the Struct is for.

So the next question would be - how did it happen, since it seems it's `nullable` value? The culprit is here: https://github.com/ezsystems/ezplatform-kernel/commit/01d1b132be9c31a6d9e30ed1c73293358d2c3e35
When specifying strict property I relied on an old ezpublish-kernel code which specified it incorrectly as `string` instead of `string|null`. This is fairly fresh API (v3.3.23), so there's a change the impact of this change won't be too big.

Let's keep your change, it was a bug :1st_place_medal: I'm showing you my thought process here just to emphasize that API changes should be tread lightly and analysed carefully.

eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php
alongosz: @barw4 ok, after more careful investigation I somewhat have an idea about what is happening here, but not fully. I agree that due to `FieldGroupLimitation` being part of Ibexa DXP, outside of the core, it's not possible to cover it with integration-level testing.

What seems to be the bug is that `ContentMapper::getFieldsForUpdate` returns incorrect set of fields for update, right?
I also see why it happens, however I wasn't able to fully reproduce this on API due to issue being more complex than I initially thought.

If `ContentMapper` is broken, you actually need to cover `ContentMapper`, not `ContentService::updateContent`.

_**I need to see an unit test which shows what result is expected from `ContentMapper::getFieldsForUpdate`**_. Please don't do partial mocks of classes under tests, this is anti-pattern we didn't have time to clean up.

alongosz (changes requested):

eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php
barw4: > What seems to be the bug is that ContentMapper::getFieldsForUpdate returns incorrect set of fields for update, right?

@alongosz Yes, this is exactly the case. Afaik we have no dedicated `ContentMapper` test anywhere, so I'll create one instead and leave `Mock\ContentTest` intact.

eZ/Publish/API/Repository/Values/Content/Field.php
barw4: @alongosz thank you Andrzej for the analysis. Of course, I agree with the fact that API changes should be carefully treated, PR is always a good start to a discussion 💟

eZ/Publish/Core/Repository/Tests/Service/Mock/ContentTest.php
alongosz: > > What seems to be the bug is that ContentMapper::getFieldsForUpdate returns incorrect set of fields for update, right?
> 
> @alongosz Yes, this is exactly the case. Afaik we have no dedicated `ContentMapper` test anywhere, so I'll create one instead and leave `Mock\ContentTest` intact.

Hint: `\Ibexa\Tests\Core\Repository\Mapper\ContentMapperTest` in `tests/lib/Repository/Mapper` is a good choice for that test class, to keep -> ibexa upgrade path smooth.

barw4: @alongosz resolved in https://github.com/ezsystems/ezplatform-kernel/pull/368/commits/2c8c2f58b3329cb52005e3a5ec7d9726d411d201

kisztof (approved):

eZ/Publish/API/Repository/Values/Content/Field.php
Steveb-p: PHPStan might indeed detect errors in dependent packages after this change, so everyone should be mindful of the fact that this is incoming.

Steveb-p (approved):

konradoboza (approved):

alongosz (approved):

bogusez (approved):

PR #369, opened 2023-04-14
IBX-5505: Preview response should have Cache-Control header set to private

mateuszbieniek:
| **JIRA issue**                          | [IBX-5505](https://issues.ibexa.co/browse/IBX-5505)
| **Target Ibexa version** | `v3.3+`

In `PreviewController` we remove `Cache-Control` and `Expires` headers from the response - which is bad, as it leads to the situation that Varnish/Fastly will treat this response as catchable and will store it in the cache.
Preview - by definition - allows previewing the current state of the draft, which implies that we want to see change. As the URL used for previewing is always the same for a single version, previewing it multiple times will result in the first preview being cached, and all next previews will not show the current state of the draft, but the one that was cached first.

This PR makes `Response` for preview private, so it won't be cached and - by extension - fixing the issue.

#### Checklist:

### DIFF
--- a/eZ/Publish/Core/MVC/Symfony/Controller/Content/PreviewController.php
+++ b/eZ/Publish/Core/MVC/Symfony/Controller/Content/PreviewController.php
@@ -130,8 +130,7 @@ EOF;
                 throw $e;
             }
         }
-        $response->headers->remove('cache-control');
-        $response->headers->remove('expires');
+        $response->setPrivate();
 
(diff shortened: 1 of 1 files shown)


adamwojs (approved):

ViniTou (approved):

alongosz (approved):

bogusez (approved):
QA Approved on v3.3.32 with .diff / tested with varnish installed on docker container

PR #370, opened 2023-04-24
IBX-5385: Add option content-type to reindex command

papcio122:
| **JIRA issue**                          | [IBX-5385](https://issues.ibexa.co/browse/IBX-5385)
| **Type**                                   | bug/improvement

#### Checklist:

### DIFF
--- a/eZ/Bundle/EzPublishCoreBundle/Command/ReindexCommand.php
+++ b/eZ/Bundle/EzPublishCoreBundle/Command/ReindexCommand.php
@@ -148,7 +148,7 @@ class ReindexCommand extends Command implements BackwardCompatibleCommand
             )->addOption(
                 'content-type',
                 null,
-                InputOption::VALUE_OPTIONAL,
+                InputOption::VALUE_REQUIRED,
                 'Content type identifier to refresh (deleted/updated/added). Implies "no-purge", cannot be combined with "since", "subtree" or "content-ids"'
             )->setHelp(
                 <<<EOT
--- a/eZ/Bundle/EzPublishCoreBundle/Command/ReindexCommand.php
+++ b/eZ/Bundle/EzPublishCoreBundle/Command/ReindexCommand.php
@@ -128,23 +128,28 @@ class ReindexCommand extends Command implements BackwardCompatibleCommand
                 'since',
                 null,
                 InputOption::VALUE_OPTIONAL,
-                'Refresh changes since a time provided in any format understood by DateTime. Implies "no-purge", cannot be combined with "content-ids" or "subtree"'
+                'Refresh changes since a time provided in any format understood by DateTime. Implies "no-purge", cannot be combined with "content-ids", "subtree" or "content-type"'
             )->addOption(
                 'content-ids',
                 null,
                 InputOption::VALUE_OPTIONAL,
-                'Comma-separated list of content ID\'s to refresh (deleted/updated/added). Implies "no-purge", cannot be combined with "since" or "subtree"'
+                'Comma-separated list of content ID\'s to refresh (deleted/updated/added). Implies "no-purge", cannot be combined with "since", "subtree" or "content-type"'
             )->addOption(
                 'subtree',
                 null,
                 InputOption::VALUE_OPTIONAL,
-                'Location ID whose subtree will be indexed (including the Location itself). Implies "no-purge", cannot be combined with "since" or "content-ids"'
+                'Location ID whose subtree will be indexed (including the Location itself). Implies "no-purge", cannot be combined with "since", "content-ids" or "content-type"'
             )->addOption(
                 'processes',
                 null,
                 InputOption::VALUE_OPTIONAL,
                 'Number of child processes to run in parallel for iterations, if set to "auto" it will set to number of CPU cores -1, set to "1" or "0" to disable',
                 'auto'
+            )->addOption(
+                'content-type',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'Content type identifier to refresh (deleted/updated/added). Implies "no-purge", cannot be combined with "since", "subtree" or "content-ids"'
             )->setHelp(
                 <<<EOT
                     The command <info>%command.name%</info> indexes the current configured database in the configured search engine index.
@@ -255,6 +260,10 @@ class ReindexCommand extends Command implements BackwardCompatibleCommand
             $location = $this->locationHandler->load($locationId);
             $count = $this->gateway->countContentInSubtree($location->pathString);
             $generator = $this->gateway->getContentInSubtree($location->pathString, $iterationCount);
+            $purge = false;
+        } elseif ($contentType = $input->getOption('content-type')) {
+            $count = $this->gateway->countContentWithContentTypeIdentifier($contentType);
+            $generator = $this->gateway->getContentWithContentTypeIdentifier($contentType, $iterationCount);
             $purge = false;
         } else {
             $count = $this->gateway->countAllContent();
--- a/eZ/Publish/Core/Search/Legacy/Content/IndexerGateway.php
+++ b/eZ/Publish/Core/Search/Legacy/Content/IndexerGateway.php
@@ -63,6 +63,22 @@ final class IndexerGateway implements SPIIndexerGateway
         return (int)$query->execute()->fetchOne();
     }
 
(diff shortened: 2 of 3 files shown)


eZ/Bundle/EzPublishCoreBundle/Command/ReindexCommand.php
```diff
@@ -128,23 +128,28 @@ protected function configure()
...
             )->addOption(
                 'processes',
                 null,
                 InputOption::VALUE_OPTIONAL,
                 'Number of child processes to run in parallel for iterations, if set to "auto" it will set to number of CPU cores -1, set to "1" or "0" to disable',
                 'auto'
+            )->addOption(
+                'content-type',
+                null,
+                InputOption::VALUE_OPTIONAL,
```
Steveb-p: If an option is passed, it's value should be required.
```suggestion
                InputOption::VALUE_REQUIRED,
```

An array of Content Types might also be considered:

```suggestion
                InputOption::VALUE_REQUIRED|InputOption::VALUE_IS_ARRAY,
```

Although this will require a few changes in code introduced here later on.

eZ/Publish/SPI/Search/Content/IndexerGateway.php
```diff
@@ -40,6 +40,18 @@ public function getContentInSubtree(string $locationPath, int $iterationCount):
      */
     public function countContentInSubtree(string $locationPath): int;
 
+    /**
+     * @throws \Doctrine\DBAL\Exception
+     *
+     * @return \Generator list of Content IDs for each iteration
+     */
+    public function getContentWithContentTypeIdentifier(string $contentTypeIdentifier, int $iterationCount): Generator;
```
Steveb-p: For an interface a wider allowed type should be ok. What we're interested in is to be able to iterate over the results, so we can relax the constraints.

Ensure that what we're returning are integers.

```suggestion
     * @return iterable<int> list of Content IDs for each iteration
     */
    public function getContentWithContentTypeIdentifier(string $contentTypeIdentifier, int $iterationCount): iterable;
```

eZ/Publish/SPI/Search/Content/IndexerGateway.php
```diff
@@ -40,6 +40,18 @@ public function getContentInSubtree(string $locationPath, int $iterationCount):
      */
     public function countContentInSubtree(string $locationPath): int;
 
+    /**
+     * @throws \Doctrine\DBAL\Exception
```
Steveb-p, line 44: As far as the indexing is concerned, I'm not sure if we want to declare that this can throw this exception. The fact that we are using Doctrine is irrelevant for indexing (at least for the interface). @alongosz ?

eZ/Bundle/EzPublishCoreBundle/Command/ReindexCommand.php
```diff
@@ -128,23 +128,28 @@ protected function configure()
...
                 'processes',
                 null,
                 InputOption::VALUE_OPTIONAL,
                 'Number of child processes to run in parallel for iterations, if set to "auto" it will set to number of CPU cores -1, set to "1" or "0" to disable',
                 'auto'
+            )->addOption(
+                'content-type',
+                null,
+                InputOption::VALUE_REQUIRED,
+                'Content type identifier to refresh (deleted/updated/added). Implies "no-purge", cannot be combined with "since", "subtree" or "content-ids"'
```
Steveb-p, line 152: Side note, outside of the scope of this PR:
We are not actually checking that conflicting options are passed.

eZ/Publish/Core/Search/Legacy/Content/IndexerGateway.php
```diff
@@ -63,6 +63,22 @@ public function countContentInSubtree(string $locationPath): int
         return (int)$query->execute()->fetchOne();
     }
 
+    public function getContentWithContentTypeIdentifier(string $contentTypeIdentifier, int $iterationCount): Generator
```
Steveb-p, line 66: The issue reported is relevant. Narrowing down of return type is available for PHP 7.4+, and this version of ezplatform-kernel supports PHP 7.3, which means you cannot use this feature.

```suggestion
    public function getContentWithContentTypeIdentifier(string $contentTypeIdentifier, int $iterationCount): iterable
```

eZ/Publish/SPI/Search/Content/IndexerGateway.php
```diff
@@ -40,6 +40,18 @@ public function getContentInSubtree(string $locationPath, int $iterationCount):
...
+     *
+     * @return iterable<int> list of Content IDs for each iteration
+     */
+    public function getContentWithContentTypeIdentifier(string $contentTypeIdentifier, int $iterationCount): iterable;
+
+    /**
+     * @throws \Doctrine\DBAL\Exception
+     */
+    public function countContentWithContentTypeIdentifier(string $contentTypeIdentifier): int;
```
Steveb-p, line 53: Are we actually ok with introducing new method into this class @alongosz? It is a contract.

eZ/Bundle/EzPublishCoreBundle/Command/ReindexCommand.php
```diff
@@ -128,23 +128,28 @@ protected function configure()
                 'since',
                 null,
                 InputOption::VALUE_OPTIONAL,
-                'Refresh changes since a time provided in any format understood by DateTime. Implies "no-purge", cannot be combined with "content-ids" or "subtree"'
+                'Refresh changes since a time provided in any format understood by DateTime. Implies "no-purge", cannot be combined with "content-ids", "subtree" or "content-type"'
```
alongosz, line 131: Punctuation: use Oxford comma here and in the other related places please
```suggestion
                'Refresh changes since a time provided in any format understood by DateTime. Implies "no-purge", cannot be combined with "content-ids", "subtree", or "content-type"'
```

eZ/Publish/SPI/Search/Content/IndexerGateway.php
alongosz: > As far as the indexing is concerned, I'm not sure if we want to declare that this can throw this exception. The fact that we are using Doctrine is irrelevant for indexing (at least for the interface). @alongosz ?

You're right. On SPI level we can throw SPI or API exceptions.

eZ/Publish/SPI/Search/Content/IndexerGateway.php
alongosz: > Are we actually ok with introducing new method into this class @alongosz? It is a contract.

No, this is strict SPI, any additions to contracts there (except for SPI\Persistence) are hard BC break.

OT: _Though, TBH I'm not sure what idea I had in mind when placing this in SPI. Maybe I thought it's gonna be implemented by each search engine, but AFAICS it should just fetch data related to indexing directly from Repository. Maybe the purpose was rather Legacy Storage strategy..._

eZ/Bundle/EzPublishCoreBundle/Command/ReindexCommand.php
```diff
@@ -256,6 +261,10 @@ protected function indexIncrementally(
             $count = $this->gateway->countContentInSubtree($location->pathString);
             $generator = $this->gateway->getContentInSubtree($location->pathString, $iterationCount);
             $purge = false;
+        } elseif ($contentType = $input->getOption('content-type')) {
+            $count = $this->gateway->countContentWithContentTypeIdentifier($contentType);
+            $generator = $this->gateway->getContentWithContentTypeIdentifier($contentType, $iterationCount);
+            $purge = false;
```
alongosz, line 267: Given you can't add anything to that Gateway without breaking BC promise (see previous comment), I think better solution would be to either use Repository Filtering (`\eZ\Publish\API\Repository\ContentService::find`) or make a dedicated method in `\eZ\Publish\SPI\Persistence\Content\Handler`. The latter option would be more performant and require implementing cache layer (which is good).

alongosz (changes requested):
This improvement is not directly related to a bug fix, but it's rather a new feature making easier to reindex data after a bug fix, right? Moreover AFAICS related CS tickets are 4.x. We fix bugs and performance issues on the oldest supported version, but we don't add features to it.

As a compromise, I would accept this solution for 4.5 instead of 4.6.x-dev, if @webhdx agrees.

Review remarks:

papcio122:
Closing this PR, moving changes to https://github.com/ibexa/core/pull/259 in order to merge this to 4.5

PR #372, opened 2023-05-09
IBX-5502: Added additional tag cleaning to limit down number of orphaned tag entries

mateuszbieniek:
| **JIRA issue**                          | [IBX-5502](https://issues.ibexa.co/browse/IBX-5502)

Due to the how cache is implemented in Symfony-based projects using the Redis adapter (and probably - not only Redis), entries (Members) inside a Tag (Set) are removed only when the tag is purged. Due to that, we can encounter on multiple occasions a situation, when a Tag is purged, resulting in cache items (Keys) being purged as well, but only the Tag that triggered a purge is cleaned up. All other tags that contain entries pointing to cache items that no longer exist will remain there until the affected tag is purged - which often is "never".

This results in orphaned entries in tags (or even in whole orphaned tags) filling up the memory. Memory allocated to tags in Redis is nonevictable, therefore it is freed only when purged on purpose (`maxmemory-policy` setting won't affect orphaned members/sets). On very big and/or constantly changing content structures, this unevictable memory allocated to orphaned tags and entries can significantly lower the amount of memory available for standard cache items to be stored.

Additional Tag purging added in this PR reduces this effect, but - unfortunately - it is not possible to resolve it fully with current Symfony's cache adapters implementation, due to the fact that there is no available interface to remove an entry from a Tag, without purging it whole - and that would result in purging cache items that were nor related to the original action performed on content.

Additional tag purging introduced by PR:
- `ContentHandler::updateContent`:
Multiple content-related entries (`ibx-c-<contentId>-<versionNo>-0`, `ibx-c-<contentId>-<versionNo>-<languageCode>`, `ibx-cvi-<contentId>`,  `ibx-cl-<contentId>-pfd` are orphaned in tags: `l-<locationId>`, `lp-<locationId>`, `c-<contentId>-v-<versionNo>`, `c-<contentId>-v-<prevVersionNo>`. Added purging of those tags to avoid it.

- `TrashHandler::trashSubtree`:
Multiple content-related entries are orphaned (`ibx-c-<contentId>-v<versionNumber>`, `ibx-cl-<contentId>`, `ibx-cvi-<contentId>`, `ibx-l-<locationId>-<languageCode>`, `ibx-c-<contentId>-<versionNo>-0`, `ibx-c-<contentId>-<versionNo>-<languageCode>`, `ibx-cvi-<contentId>` ... (many more) are orphaned in tags: `l-<locationId>`, `c-<contentId>-v-<versionNo>`.

Handling those two occurrences helps to limit greatly the number of orphaned members.

My tests show, that one of the biggest culprits here is `lp-<locationIdOfParent>` tag, which can have multiple orphaned entries - therefore it might be a good idea to purge it as well, to avoid blocked memory in cost of potentially small performance loss - up for discussion.

This PR won't resolve the issue fully but limits the scope of it using available tools. To resolve this fully, we will need (in my opinion):
- an interface to remove entries from sets
- a garbage collector, as entries in tags might become orphaned as well when a key is evicted by Redis when maxmemory limit is reached.
But those two are out of this PR scope.

#### Checklist:

### DIFF
--- a/eZ/Publish/Core/Persistence/Cache/ContentHandler.php
+++ b/eZ/Publish/Core/Persistence/Cache/ContentHandler.php
@@ -333,7 +333,7 @@ class ContentHandler extends AbstractInMemoryPersistenceHandler implements Conte
         $locationTags = array_map(function (Content\Location $location): string {
             return $this->cacheIdentifierGenerator->generateTag(self::LOCATION_IDENTIFIER, [$location->id]);
         }, $locations);
-        $locationPathTags = array_map(function (Content\Location $location) {
+        $locationPathTags = array_map(function (Content\Location $location): string {
             return $this->cacheIdentifierGenerator->generateTag(self::LOCATION_PATH_IDENTIFIER, [$location->id]);
         }, $locations);
 
--- a/eZ/Publish/Core/Persistence/Cache/ContentHandler.php
+++ b/eZ/Publish/Core/Persistence/Cache/ContentHandler.php
@@ -330,7 +330,7 @@ class ContentHandler extends AbstractInMemoryPersistenceHandler implements Conte
         $content = $this->persistenceHandler->contentHandler()->updateContent($contentId, $versionNo, $struct);
         $locations = $this->persistenceHandler->locationHandler()->loadLocationsByContent($contentId);
 
-        $locationTags = array_map(function (Content\Location $location) {
+        $locationTags = array_map(function (Content\Location $location): string {
             return $this->cacheIdentifierGenerator->generateTag(self::LOCATION_IDENTIFIER, [$location->id]);
         }, $locations);
         $locationPathTags = array_map(function (Content\Location $location) {
--- a/eZ/Publish/Core/Persistence/Cache/ContentHandler.php
+++ b/eZ/Publish/Core/Persistence/Cache/ContentHandler.php
@@ -330,7 +330,6 @@ class ContentHandler extends AbstractInMemoryPersistenceHandler implements Conte
         $content = $this->persistenceHandler->contentHandler()->updateContent($contentId, $versionNo, $struct);
         $locations = $this->persistenceHandler->locationHandler()->loadLocationsByContent($contentId);
 
-
         $locationTags = array_map(function (Content\Location $location) {
             return $this->cacheIdentifierGenerator->generateTag(self::LOCATION_IDENTIFIER, [$location->id]);
         }, $locations);
--- a/eZ/Publish/Core/Persistence/Cache/ContentHandler.php
+++ b/eZ/Publish/Core/Persistence/Cache/ContentHandler.php
@@ -328,12 +328,34 @@ class ContentHandler extends AbstractInMemoryPersistenceHandler implements Conte
     {
         $this->logger->logCall(__METHOD__, ['content' => $contentId, 'version' => $versionNo, 'struct' => $struct]);
         $content = $this->persistenceHandler->contentHandler()->updateContent($contentId, $versionNo, $struct);
-        $this->cache->invalidateTags([
-            $this->cacheIdentifierGenerator->generateTag(
+        $locations = $this->persistenceHandler->locationHandler()->loadLocationsByContent($contentId);
+
+
+        $locationTags = array_map(function (Content\Location $location) {
+            return $this->cacheIdentifierGenerator->generateTag(self::LOCATION_IDENTIFIER, [$location->id]);
+        }, $locations);
+        $locationPathTags = array_map(function (Content\Location $location) {
+            return $this->cacheIdentifierGenerator->generateTag(self::LOCATION_PATH_IDENTIFIER, [$location->id]);
+        }, $locations);
+
+        $versionTags = [];
+        $versionTags[] = $this->cacheIdentifierGenerator->generateTag(
+            self::CONTENT_VERSION_IDENTIFIER,
+            [$contentId, $versionNo]
+        );
+        if ($versionNo > 1) {
+            $versionTags[] = $this->cacheIdentifierGenerator->generateTag(
                 self::CONTENT_VERSION_IDENTIFIER,
-                [$contentId, $versionNo]
-            ),
-        ]);
+                [$contentId, $versionNo - 1]
+            );
+        }
+
+        $tags = array_merge(
+            $locationTags,
+            $locationPathTags,
+            $versionTags
+        );
+        $this->cache->invalidateTags($tags);
 
         return $content;
     }
@@ -350,6 +372,7 @@ class ContentHandler extends AbstractInMemoryPersistenceHandler implements Conte
             $contentId,
             APIRelation::FIELD | APIRelation::ASSET
         );
+        $contentLocations = $this->persistenceHandler->locationHandler()->loadLocationsByContent($contentId);
 
         $return = $this->persistenceHandler->contentHandler()->deleteContent($contentId);
 
@@ -365,6 +388,9 @@ class ContentHandler extends AbstractInMemoryPersistenceHandler implements Conte
             $tags = [];
         }
         $tags[] = $this->cacheIdentifierGenerator->generateTag(self::CONTENT_IDENTIFIER, [$contentId]);
+        foreach ($contentLocations as $location) {
+            $tags[] = $this->cacheIdentifierGenerator->generateTag(self::LOCATION_IDENTIFIER, [$location->id]);
+        }
         $this->cache->invalidateTags($tags);
 
         return $return;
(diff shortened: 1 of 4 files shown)


eZ/Publish/Core/Persistence/Cache/ContentHandler.php
```diff
@@ -328,12 +328,33 @@ public function updateContent($contentId, $versionNo, UpdateStruct $struct)
     {
         $this->logger->logCall(__METHOD__, ['content' => $contentId, 'version' => $versionNo, 'struct' => $struct]);
         $content = $this->persistenceHandler->contentHandler()->updateContent($contentId, $versionNo, $struct);
-        $this->cache->invalidateTags([
-            $this->cacheIdentifierGenerator->generateTag(
+        $locations = $this->persistenceHandler->locationHandler()->loadLocationsByContent($contentId);
+
+        $locationTags = array_map(function (Content\Location $location) {
```
konradoboza:
```suggestion
        $locationTags = array_map(function (Content\Location $location): string {
```

konradoboza:
```suggestion
        $locationTags = array_map(function (Content\Location $location): string {
```

eZ/Publish/Core/Persistence/Cache/ContentHandler.php
```diff
@@ -328,12 +328,33 @@ public function updateContent($contentId, $versionNo, UpdateStruct $struct)
...
+            return $this->cacheIdentifierGenerator->generateTag(self::LOCATION_IDENTIFIER, [$location->id]);
+        }, $locations);
+        $locationPathTags = array_map(function (Content\Location $location) {
```
konradoboza:
```suggestion
        $locationPathTags = array_map(function (Content\Location $location): string {
```

eZ/Publish/Core/Persistence/Cache/TrashHandler.php
```diff
@@ -51,12 +55,21 @@ public function trashSubtree($locationId)
             }, $reverseRelations);
         }
 
+        $versionTags = array_map(function (VersionInfo $versionInfo) use ($contentId) {
```
konradoboza:
```suggestion
        $versionTags = array_map(function (VersionInfo $versionInfo) use ($contentId): string {
```

konradoboza (changes requested):
Failing tests need to be investigated. Besides, you mentioned clearing tags with `-languageCode` suffix and I am kinda missing where it's done exactly.

mateuszbieniek:
I apologize - I've missed the CR. 

> Failing tests need to be investigated.

Fixed.

> Besides, you mentioned clearing tags with `-languageCode` suffix and I am kinda missing where it's done exactly.

It is not a tag, but a key. For example, this key is a member of a "Location" tag - those additional tag purgings resolves this key being orphaned when Content is sent to the Trash.

kisztof (changes requested):
Could you implement the code suggestions provided by @konradoboza https://github.com/ezsystems/ezplatform-kernel/pull/372/files#r1189636279

mateuszbieniek:
@kisztof done (test failing not related to pr) and thank you for reminding me. Happy new year!

eZ/Publish/Core/Persistence/Cache/Tests/ContentHandlerTest.php
```diff
@@ -47,7 +49,7 @@ public function providerForUnCachedMethods(): array
             ['setStatus', [2, 0, 1], [['content_version', [2, 1], false]], null, ['c-2-v-1']],
             ['setStatus', [2, 1, 1], [['content', [2], false]], null, ['c-2']],
             ['updateMetadata', [2, new MetadataUpdateStruct()], [['content', [2], false]], null, ['c-2']],
-            ['updateContent', [2, 1, new UpdateStruct()], [['content_version', [2, 1], false]], null, ['c-2-v-1']],
+            //['updateContent', [2, 1, new UpdateStruct()], [['content_version', [2, 1], false]], null, ['c-2-v-1']],
```
konradoboza, line 53: Why is this line commented?

eZ/Publish/Core/Persistence/Cache/Tests/ContentHandlerTest.php
```diff
@@ -303,30 +305,68 @@ public function providerForCachedLoadMethodsMiss(): array
         ];
     }
 
+    /**
+     * @covers \eZ\Publish\Core\Persistence\Cache\ContentHandler::updateContent
+     */
+    public function testUpdateContent()
```
konradoboza: There are plenty of missing return types in here, but increasing code quality even with such small steps is worth the effort to me:
```suggestion
    public function testUpdateContent(): void
```

eZ/Publish/Core/Persistence/Cache/Tests/ContentHandlerTest.php
```diff
@@ -303,30 +305,68 @@ public function providerForCachedLoadMethodsMiss(): array
...
+        $innerContentHandlerMock
+            ->expects($this->once())
+            ->method('updateContent')
+            ->with(2, 1, new UpdateStruct())
+            ->willReturn(new Content());
+
+        $this->cacheIdentifierGeneratorMock
+            ->expects($this->exactly(5))
+            ->method('generateTag')
+            ->withConsecutive(
```
konradoboza: `withConsecutive` and `willReturnOnConsecutiveCalls` are deprecated in the newer versions of PHPUnit. Please consider changing those to `returnValueMap` here as it would be needed on merge-up anyway. For example, you can refer to https://github.com/ibexa/product-catalog/pull/1034/files#diff-df82e5aba4361380977da9b3843954e8614ebbb83c317a6bd45d4b53c0fe55f2R95.

eZ/Publish/Core/Persistence/Cache/TrashHandler.php
```diff
@@ -51,13 +55,23 @@ public function trashSubtree($locationId)
             }, $reverseRelations);
         }
 
+        $versionTags = array_map(function (VersionInfo $versionInfo) use ($contentId): string {
+            return  $this->cacheIdentifierGenerator->generateTag(
```
konradoboza:
```suggestion
            return $this->cacheIdentifierGenerator->generateTag(
```

eZ/Publish/Core/Persistence/Cache/TrashHandler.php
```diff
@@ -37,8 +40,9 @@ public function trashSubtree($locationId)
         $this->logger->logCall(__METHOD__, ['locationId' => $locationId]);
 
         $location = $this->persistenceHandler->locationHandler()->load($locationId);
-        $reverseRelations = $this->persistenceHandler->contentHandler()->loadRelations($location->contentId);
-
+        $contentId = $location->contentId;
+        $reverseRelations = $this->persistenceHandler->contentHandler()->loadRelations($contentId);
+        $versions = $this->persistenceHandler->contentHandler()->listVersions($contentId);
```
konradoboza:
```suggestion
        $contentHandler = $this->persistenceHandler->contentHandler();
        $reverseRelations = $contentHandler->loadRelations($contentId);
        $versions = $contentHandler->listVersions($contentId);
```

konradoboza (commented):
> My tests show, that one of the biggest culprits here is lp-<locationIdOfParent> tag, which can have multiple orphaned entries - therefore it might be a good idea to purge it as well, to avoid blocked memory in cost of potentially small performance loss - up for discussion.

Are you able to tell what performance losses are we observing here?

eZ/Publish/Core/Persistence/Cache/Tests/ContentHandlerTest.php
mateuszbieniek: As you can see this is a copypaste of the line above - I was probably checking something and overlooked it.
Ofc I'll fix that - thank you for spotting this!

mateuszbieniek:
> > My tests show, that one of the biggest culprits here is lp- tag, which can have multiple orphaned entries - therefore it might be a good idea to purge it as well, to avoid blocked memory in cost of potentially small performance loss - up for discussion.
> 
> Are you able to tell what performance losses are we observing here?

To be frank no, but looking at it this is a simple tag. Due to the way how caching works - worst-case scenario it will get recreated the next time it is required, and as it is `lp` - Location Parent (?) - I would not expect anything spectacular, especially - judging from the public issue description - you have to empty trash for this to get orphaned - I won't expect that this will be ever again requested after that.

In addition, freeing this memory to be reused should - especially in the long run - benefit to overall application performance (more memory, less evictions).

eZ/Publish/Core/Persistence/Cache/Tests/ContentHandlerTest.php
mateuszbieniek: Ok, that was intentional - due to added complexcity in `updateContent` method, I had to create specific test case for it, as it cannot be test in `bulk` (same ad `deleteContent`)

mateuszbieniek:
Closing in favor of:
https://github.com/ibexa/core/pull/477

PR #373, opened 2023-05-15
IBX-5663: Added paginated role assignments load method to API

barw4:
| **JIRA issue**                          | [IBX-5663](https://issues.ibexa.co/browse/IBX-5663)

2 new methods were added to `RoleService` in order to improve performance in the Back Office:
- `loadRoleAssignments` - allows setting `$offset` and `$limit` arguments in order to allow pagination. The old method `getRoleAssignments` even if cached could result in a huge performance drop in the Back Office as the amount of assignments can be really large.
-  `countRoleAssignments` - counting role assignments, usable for the above method.

Related PR: https://github.com/ezsystems/ezplatform-admin-ui/pull/2102

#### Checklist:


ViniTou:
Is there a possibility to use proxies for policies instead of introducing a new method and later thinking why there are nulls for some cases?

barw4:
@ViniTou I'll take a look into it.

alongosz (changes requested):
There are a business logic and security issues here. Currently there's no way to distinguish between a `Role` without policies from a `Role` without loaded policies. From API contract perspective information returned like that is not correct and could lead to treating the object as policy-less one, leading to security holes in places where we forgot or 3rd party is not aware that the information loaded is not full.

> Is there a possibility to use proxies for policies instead of introducing a new method and later thinking why there are null for some cases?

Seems like the most reasonable way.

barw4:
@alongosz @ViniTou as policies property in Role is an array and assuming we don't want to change that how would you want to approach building a proxy? ProxyGenerator requires an actual object to be proxied, so would you like to proxy every policy separately? It would require fetching ids of those policies either way so performance gain would be minimal.

ViniTou:
@barw4 
ArrayObject wouldnt probably be some solution, as I am almost sure it would fail at some level, and I think it will count as BC.

Other way, would be to introduce new class for that Role without policies, but probably above will apply as well. 
Atm, no other ideas that wouldnt be some __get hacking things, but in current state I agree this is no go - it creates more problems than it resolves.

Is adding just pagination for role assigments not enough?

barw4:
@ViniTou it should suffice for now, at least according to the public ticket. I've removed `listRoles`.

eZ/Publish/API/Repository/RoleService.php
```diff
@@ -277,6 +277,30 @@ public function loadRoleAssignment(int $roleAssignmentId): RoleAssignment;
...
+     * @return \eZ\Publish\API\Repository\Values\User\RoleAssignment[]
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
+     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
+     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException if the authenticated user is not allowed to read a role
+     */
+    public function loadRoleAssignments(
+        Role $role,
+        int $offset = 0,
+        int $limit = -1
```
alongosz: I would challenge here using `-1` to denote "no limit". I always considered this as a flaw in our API. There's a better way to indicate that there's no limit - `?int $limit = null`. You have to compare it later on anyway.

POV ping @ezsystems/php-dev-team

eZ/Publish/API/Repository/Tests/RoleServiceTest.php
```diff
@@ -1872,6 +1873,78 @@ public function testGetRoleAssignmentsContainExpectedLimitation(array $roleAssig
...
+        $roleAssignments = $roleService->loadRoleAssignments($loadedRole, 0, 1);
+
+        self::assertCount(1, $roleAssignments);
+        self::assertInstanceOf(UserRoleAssignment::class, $roleAssignments[0]);
+    }
+
+    /**
+     * @covers \eZ\Publish\API\Repository\RoleService::countRoleAssignments()
+     *
+     * @depends eZ\Publish\API\Repository\Tests\RoleServiceTest::testCountRoleAssignments
```
alongosz: This is unnecessary. There's no direct data dependency on the state of testCountRoleAssignments. If there's no direct need for it, let's avoid it please, it introduces more issues than it solves when trying to run single tests.

Additionally, if we look at some quality tools, they have problem with properly recognizing such annotation, as this is not the best design by PHPUnit, due to no leading slash.
> [EA] `@depends` referencing to a non-existing or inappropriate entity.

eZ/Publish/API/Repository/Tests/RoleServiceTest.php
```diff
@@ -1872,6 +1873,78 @@ public function testGetRoleAssignmentsContainExpectedLimitation(array $roleAssig
...
+        $newUser = $this->createUser('login', 'Test', 'Test');
+
+        $roleService->assignRoleToUser($role, $newUser);
+
+        // Removing user should subtract '1' from the assignments count
+        $userService->deleteUser($newUser);
+
+        $newCount = $roleService->countRoleAssignments($role);
+
+        self::assertEquals($currentCount, $newCount);
```
alongosz: You're actually not testing here what you want to test. Get `$currentCount` after assigning user and then check if:
```php
 self::assertEquals($currentCount - 1, $newCount);
```

eZ/Publish/Core/Persistence/Legacy/User/Role/Gateway/DoctrineDatabase.php
```diff
@@ -334,6 +334,67 @@ public function loadRoleAssignmentsByRoleId(int $roleId): array
...
+            $query->expr()->eq(
+                'role_id',
+                $query->createPositionalParameter($roleId, ParameterType::INTEGER)
+            )
+        )
+        ->setFirstResult($offset);
+
+        if ($limit > 0) {
+            $query->setMaxResults($limit);
+        }
```
alongosz: `LIMIT 0` is a perfectly valid expression. It tells DBMS to create query plan and execute it without returning its result for fetching. We use this if we want to perform count query earlier and then data query with the same constraints, but we're not interested in the results.

Following my earlier comment, it could be `if (null !== $limit)`

eZ/Publish/Core/Repository/RoleService.php
```diff
@@ -881,25 +877,58 @@ public function getRoleAssignments(APIRole $role): iterable
...
+
+        return $this->buildRoleAssignmentsFromPersistence($role, $persistenceRoleAssignments);
+    }
+
+    /**
+     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
+     */
+    private function buildRoleAssignmentsFromPersistence(
+        APIRole $role,
+        array $persistenceRoleAssignments
```
alongosz, line 907: Static analysis: this needs to have `@param` or `@phpstan-param` definition.

eZ/Publish/Core/Repository/RoleService.php
```diff
@@ -881,25 +877,58 @@ public function getRoleAssignments(APIRole $role): iterable
...
+        return $this->buildRoleAssignmentsFromPersistence($role, $persistenceRoleAssignments);
+    }
+
+    /**
+     * @return \eZ\Publish\API\Repository\Values\User\RoleAssignment[]
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
+     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
+     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException if the authenticated user is not allowed to read a role
+     */
```
alongosz: you don't need to repeat the same doc if it's already on API / interface level definition.

eZ/Publish/Core/Repository/RoleService.php
```diff
@@ -881,25 +877,58 @@ public function getRoleAssignments(APIRole $role): iterable
...
+        return $this->buildRoleAssignmentsFromPersistence($role, $persistenceRoleAssignments);
+    }
+
+    /**
+     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
+     */
+    private function buildRoleAssignmentsFromPersistence(
+        APIRole $role,
+        array $persistenceRoleAssignments
+    ): array {
```
alongosz, line 908: Given the consumers of that method return `iterable`, how about making this `iterable` as well by yielding the results rather than accumulating them in the `$roleAssignments` array?

eZ/Publish/Core/Persistence/Cache/UserHandler.php
```diff
@@ -495,6 +495,30 @@ function () use ($roleId) {
...
+    public function countRoleAssignments(int $roleId): int
+    {
+        $this->logger->logCall(__METHOD__, [
+            'userId' => $roleId,
+        ]);
+
+        return $this->persistenceHandler
+            ->userHandler()
+            ->countRoleAssignments($roleId);
+    }
```
alongosz: These are good candidates to be cached. Otherwise in place of 1 performance issue we're introducing another one.

alongosz (changes requested):
Overall this is a good direction :muscle: 
It needs a little bit of improvement:

eZ/Publish/Core/Persistence/Cache/UserHandler.php
barw4: @alongosz do we want to cache role assignments with `offset` and `limit`-generated keys?

eZ/Publish/API/Repository/Tests/RoleServiceTest.php
barw4: Hmm, I specifically mention in the code that we are adding 1 and subtracting 1 which should give us the initial value (so `$currentCount`).

eZ/Publish/Core/Repository/RoleService.php
barw4: @alongosz the problem is that even if consumers return `iterable`, some places in other packages expect those consumer methods to return a simple array. Not sure if that was a mistake or else.

eZ/Publish/API/Repository/Tests/RoleServiceTest.php
Steveb-p: Then perform assertions in between the changes. It does seem weird at first glance that the count value does not change.

eZ/Publish/Core/Repository/RoleService.php
alongosz: > @alongosz the problem is that even if consumers return `iterable`, some places in other packages expect those consumer methods to return a simple array. Not sure if that was a mistake or else.

@barw4 both `loadRoleAssignments` and `getRoleAssignments` return `iterable`, so I really don't follow. Could you give an example?

eZ/Publish/Core/Persistence/Cache/UserHandler.php
alongosz: > @alongosz do we want to cache role assignments with `offset` and `limit`-generated keys?

Yes, that would be the approach here.

eZ/Publish/Core/Repository/RoleService.php
barw4: @alongosz sure, they return `iterable`, but in fact, both methods return simple arrays, even before my change. For example, an array is required here https://github.com/ezsystems/ezplatform-admin-ui/blob/2.3/src/bundle/Controller/RoleAssignmentController.php#L241. So what I meant by my previous comment is that we would have to convert those `Generator` yield values to a simple array in `getRoleassignments` and `loadRoleAssignments` with `iterator_to_array` or something similar or have I misunderstood you?

Steveb-p: @alongosz basically while our contract specifies we will return iterables, in a few places (in our own code) we operate as if it was always array.

Unless we can identify and fix all those places (including any potential usage of these in third-party code) we should refrain from changing it, especially in 3.x codebase.

alongosz: Seems rather like a consumer bug, but fair enough, if the impact is high, let's keep it as-is :+1:

eZ/Publish/API/Repository/Tests/RoleServiceTest.php
```diff
@@ -1872,6 +1873,81 @@ public function testGetRoleAssignmentsContainExpectedLimitation(array $roleAssig
...
+
+        self::assertEquals($expectedCount, count($roleAssignments));
+
+        // Removing user should subtract '1' from the assignments count
+        $userService->deleteUser($newUser);
+        $expectedCount = $expectedCount - 1;
+
+        $roleAssignments = $roleService->loadRoleAssignments($role);
+
+        self::assertEquals($expectedCount, count($roleAssignments));
```
alongosz: For those we have `assertCount`

eZ/Publish/Core/Persistence/Cache/UserHandler.php
barw4: @alongosz Done in https://github.com/ezsystems/ezplatform-kernel/pull/373/commits/518bcbbf97a7f2691874f4cd8aa41b22cb920294. 

However, I feel like caching count may be not optimal. Essentially, for it to work properly we need to tag count with every assignment's tags so tags added for a single cache count may be huge in numbers (https://github.com/ezsystems/ezplatform-kernel/pull/373/files#diff-8ccb435bd05acfd0a861eabbff63986e588674045a205a0f00c8a46c0a74ff48R551).

eZ/Publish/Core/Persistence/Legacy/User/Role/Gateway/DoctrineDatabase.php
```diff
@@ -334,6 +334,67 @@ public function loadRoleAssignmentsByRoleId(int $roleId): array
...
+        $query = $this->connection->createQueryBuilder();
+
+        $query
+            ->select($this->connection->getDatabasePlatform()->getCountExpression('user_role.id'))
+            ->innerJoin(
+                'user_role',
+                self::CONTENT_OBJECT_TABLE,
+                'content_object',
+                'user_role.contentobject_id = content_object.id'
+            )
```
alongosz: Do you need this join to count user role assignments?

If this method should be related to the list returned via `loadRoleAssignmentsByRoleIdWithOffsetAndLimit`, it should share the same query but use different select expression, add limit/offset when needed and fetching results differently. 

Basically I would expect some private method which builds such query, so we can be sure we're counting and fetching the same result set. Unless I'm mistaken about the purpose here?

eZ/Publish/SPI/Persistence/User/Handler.php
```diff
@@ -208,6 +208,18 @@ public function loadRoleAssignment($roleAssignmentId);
...
 
+    /**
+     * Loads Role's assignments based on provided $offset and $limit arguments.
+     *
+     * @return \eZ\Publish\SPI\Persistence\User\RoleAssignment[]
+     */
+    public function loadRoleAssignmentsByRoleIdWithOffsetAndLimit(int $roleId, int $offset, ?int $limit): array;
+
+    /**
+     * Count Role's assignments taking into consideration related and existing user and user group objects.
```
alongosz: nitpick: consistency - Either "Load" in the line 212 or "Counts" here ;-)

eZ/Publish/Core/Persistence/Cache/UserHandler.php
```diff
@@ -495,6 +497,94 @@ function () use ($roleId) {
...
+            ],
+            $tags
+        );
+
+        $cacheItem->set($count);
+        $cacheItem->tag($tags);
+
+        $this->cache->save($cacheItem);
+
+        return $count;
```
alongosz: It should be enough to tag this count with:
* `$this->cacheIdentifierGenerator->generateTag(self::ROLE_ASSIGNMENT_ROLE_LIST_IDENTIFIER, [$roleId]),`
* `$this->cacheIdentifierGenerator->generateTag(self::ROLE_IDENTIFIER, [$roleId])`

The need to `loadRoleAssignmentsByRoleId($roleId)` here counteracts the whole purpose of caching the count ;-)

alongosz (changes requested):

eZ/Publish/Core/Persistence/Cache/UserHandler.php
barw4: @alongosz it's not enough though. Please notice that role assignments in `loadRoleAssignmentsByRoleId` are tagged as well. If I won't tag count removing any role assignment won't invalidate the count. That's why I've mentioned that caching count is problematic.

> The need to loadRoleAssignmentsByRoleId($roleId) here counteracts the whole purpose of caching the count ;-)

Exactly my point in my previous comment (https://github.com/ezsystems/ezplatform-kernel/pull/373#discussion_r1200618823).

eZ/Publish/Core/Persistence/Legacy/User/Role/Gateway/DoctrineDatabase.php
barw4: > Do you need this join to count user role assignments?

@alongosz Yes, this is necessary imo. In https://github.com/ezsystems/ezplatform-kernel/pull/373/files#diff-2010307b5a19e4fef041125ffd0cca3f4363b79da0c8d718a033390aea07b347R905 we are validating every assignment to check whether the user or user group exists. 

If we delete a user, without deleting a role assignment, the count without join will be bigger by a factor of 1. I could skip building `RoleAssignments` in `RoleService:buildRoleAssignmentsFromPersistence()` though and make the same join in `loadRoleAssignmentsByRoleIdWithOffsetAndLimit` if you wish.

eZ/Publish/Core/Persistence/Cache/UserHandler.php
alongosz: > @alongosz it's not enough though. Please notice that role assignments in `loadRoleAssignmentsByRoleId` are tagged as well. If I won't tag count removing any role assignment won't invalidate the count. That's why I've mentioned that caching count is problematic.
> 
> > The need to loadRoleAssignmentsByRoleId($roleId) here counteracts the whole purpose of caching the count ;-)
> 
> Exactly my point in my previous comment ([#373 (comment)](https://github.com/ezsystems/ezplatform-kernel/pull/373#discussion_r1200618823)).

Indeed. However the solution here is not to avoid caching. Seems we have an issue when it comes to invalidation of role assignment list cache. It occurs both for removing role assignment, but also for deleting user.
I've investigated this a bit and created a suggestion of how to fix it: https://github.com/ezsystems/ezplatform-kernel/pull/374. Tried to fix also side issues. See if that works for you and check please if it works with an actual instance.

I'll update #374 description explaining why and what got fixed either later today or tomorrow.

eZ/Publish/API/Repository/Tests/RoleServiceTest.php
```diff
@@ -1872,6 +1873,81 @@ public function testGetRoleAssignmentsContainExpectedLimitation(array $roleAssig
...
+        $expectedCount = count($roleAssignments);
+
+        // Adding user should add '1' to the assignments count
+        $newUser = $this->createUser('login', 'Test', 'Test');
+        $roleService->assignRoleToUser($role, $newUser);
+        $expectedCount = $expectedCount + 1;
+
+        $roleAssignments = $roleService->loadRoleAssignments($role);
+
+        self::assertEquals($expectedCount, count($roleAssignments));
```
alongosz: On a closer look, I see that you're actually not testing here countRoleAssignment method, but still `loadRoleAssignments`. This is good, we need more coverage for deleting user and unassigning role. Here just test method name needs adjustment. Reflected the change in #374

eZ/Publish/Core/Persistence/Cache/UserHandler.php
alongosz: This whole thing, part from https://github.com/ezsystems/ezplatform-kernel/pull/373/files#r1205653912 discussion, misses in-memory cache hits. Those dedicated methods, like `getCacheValue` are handling that. See #374 for the suggested change.

alongosz (commented):
@barw4 I've updated #374 description and left here also diff note giving more details about what I think we need to change.

eZ/Publish/Core/Persistence/Cache/UserHandler.php
alongosz: Conclusion on that matter after Internal sync with @adamwojs, @barw4, and @webhdx:

we can't cache count without rewriting user group view in AdminUI because it trashes groups like normal Content instead of using dedicated `UserService::deleteUserGroup` API. Rewriting it as a part of this issue is too complex, needs to be planned most likely after 4.6.0 LTS, even if we decide to fix it for 3.3.x.

eZ/Publish/Core/Persistence/Legacy/User/Role/Gateway.php
```diff
@@ -24,7 +24,7 @@ abstract class Gateway
     public const POLICY_LIMITATION_TABLE = 'ezpolicy_limitation';
     public const POLICY_LIMITATION_VALUE_TABLE = 'ezpolicy_limitation_value';
     public const USER_ROLE_TABLE = 'ezuser_role';
-
+    public const CONTENT_OBJECT_TABLE = 'ezcontentobject';
```
alongosz: You already have this const in `\eZ\Publish\Core\Persistence\Legacy\Content\Gateway::CONTENT_ITEM_TABLE`. It's there to be reused in persistence internals ;)

eZ/Publish/Core/Persistence/Legacy/User/Role/Gateway/DoctrineDatabase.php
```diff
@@ -334,6 +334,73 @@ public function loadRoleAssignmentsByRoleId(int $roleId): array
...
+            ->from(
+                self::USER_ROLE_TABLE,
+                'user_role'
+            )->where(
+                $query->expr()->eq(
+                    'role_id',
+                    $query->createPositionalParameter($roleId, ParameterType::INTEGER)
+                )
+            );
```
alongosz: This is almost ok. Now notice that you have 2x almost the same code.

If you extract the common part as:

```php
    /**
     * @param array<string> $columns
     */
    private function buildLoadRoleAssignmentsQuery(array $columns, int $roleId): QueryBuilder
    {
        $query = $this->connection->createQueryBuilder();
        $query
            ->select(...$columns)
            ->from(self::USER_ROLE_TABLE, 'user_role')
            ->innerJoin(
                'user_role',
                self::CONTENT_OBJECT_TABLE,
                'content_object',
                'user_role.contentobject_id = content_object.id'
            )->where(
                $query->expr()->eq(
                    'role_id',
                    $query->createPositionalParameter($roleId, ParameterType::INTEGER)
                )
            );

        return $query;
    }
```

then `countRoleAssignments` could have the form of:
```php
        $query = $this->buildLoadRoleAssignmentsQuery(
            [$this->connection->getDatabasePlatform()->getCountExpression('user_role.id')],
            $roleId
        );

        return (int)$query->execute()->fetchOne();
```

and `loadRoleAssignmentsByRoleIdWithOffsetAndLimit`:
```php
        $query = $this
            ->buildLoadRoleAssignmentsQuery(
                [
                    'user_role.id',
                    'user_role.contentobject_id',
                    'user_role.limit_identifier',
                    'user_role.limit_value',
                    'user_role.role_id',
                ],
                $roleId
            )
            ->setFirstResult($offset);

        if ($limit !== null) {
            $query->setMaxResults($limit);
        }

        return $query
            ->execute()
            ->fetchAllAssociative();
```

It's crucial that pagination and count for pagination re-uses the same query to ensure the stability of the solution :-)

alongosz (changes requested):

eZ/Publish/Core/Persistence/Legacy/User/Role/Gateway/DoctrineDatabase.php
barw4: @alongosz good idea as queries are matching right now, resolved in https://github.com/ezsystems/ezplatform-kernel/pull/373/commits/156d25b81696240e50b7ddb77ff1e0b38b591c42

alongosz (approved):
**Note for reviewers:** persistence cache for `countRoleAssignments` method was not implemented on purpose after a discovery phase and going back and forth on the solution.
It's because backoffice uses Trash and Content API to trash and then delete User Group, instead of User API. This makes role to user group assignment cache for count impossible to invalidate in a clean way.

We need to take another iteration to solve this separately, most likely after 4.6.0 LTS release.

eZ/Publish/API/Repository/Tests/RoleServiceTest.php
```diff
@@ -2959,4 +3057,19 @@ private function createUserGroupVersion1()
 
         return $userGroup;
     }
+
+    private function loadRoleAssignmentForUser(RoleService $roleService, Role $role, User $newUser): UserRoleAssignment
+    {
+        [$userRoleAssignment] = array_values(
+            array_filter(
+                (array)$roleService->getRoleAssignments($role),
+                static function (RoleAssignment $roleAssignment) use ($newUser) {
```
konradoboza:
```suggestion
                static function (RoleAssignment $roleAssignment) use ($newUser): bool {
```

konradoboza (approved):

micszo (approved):
Reproduced the 30 seconds execution timeout with ca. 6k assigned users. Problem was occurring with new cache. Consecutive reloads took 7-8 seconds.
With the fix such role loads like any standard role, in 1-2 seconds in a dev env.

QA Approved on Ibexa Experience 3.3.34-dev with diffs.

alongosz:
Nice work @barw4 :muscle:

PR #374, opened 2023-05-25
IBX-5663: Fixed role assignment cache invalidation

alongosz:
| **JIRA issue**                          | Suggestion for #373 ([IBX-5663](https://issues.ibexa.co/browse/IBX-5663))
| **Target** | #373

This PR aims to rectify role assignment cache invalidation issues discovered in #373. Instead of loading every role assignment to tag assignment count item, the cached `UserHandler::removeRoleAssignment` method has been fixed tag properly a list of the assignment it belongs to. In order to do that I loaded role assignment to get user and role id.

This had unwanted side effect - the method didn't have expected structure anymore due to that extra load, so dedicated test coverage for it was implemented.

I also suggest we change cache key/tag pattern from `ra-%s-count` to `rac-%s`. Base `getCacheValue` then generates more "reasonable" key name (`%s` needs to be at the end of the pattern). Otherwise we need to either make custom handling or use a key e.g. `ra-9-count-9`.

Another change is a fix for role assignment cache invalidation when deleting user. It was simply absent, so if user got deleted, no lists were actually invalidated. At first glance it seems the role assignment data was not removed at all.


alongosz (commented):
@barw4 I've updated #374 description.

Note: indeed there's more to improve according to SonarCloud report.

eZ/Publish/Core/Persistence/Cache/UserHandler.php
```diff
@@ -527,62 +527,25 @@ function () use ($roleId): array {
...
-                $tags[] = $tag;
+            function () use ($roleId): array {
+                return [
+                    $this->cacheIdentifierGenerator->generateTag(self::ROLE_ASSIGNMENT_ROLE_LIST_IDENTIFIER, [$roleId]),
+                    $this->cacheIdentifierGenerator->generateTag(self::ROLE_IDENTIFIER, [$roleId]),
+                ];
+            },
+            function () use ($roleId): array {
+                return [
+                    $this->cacheIdentifierGenerator->generateKey(self::ROLE_IDENTIFIER, [$roleId], true),
```
barw4:
```suggestion
                    $this->cacheIdentifierGenerator->generateKey(self::ROLE_ASSIGNMENTS_COUNT, [$roleId], true),
```
@alongosz I think this is broken, loading roles doesn't work. Shouldn't it be done as above? A typo perhaps?

eZ/Publish/Core/Repository/UserService.php
```diff
@@ -617,7 +617,13 @@ public function deleteUser(APIUser $user): iterable
 
         $this->repository->beginTransaction();
         try {
-            $affectedLocationIds = $this->repository->getContentService()->deleteContent($loadedUser->getVersionInfo()->getContentInfo());
+            foreach ($this->userHandler->loadRoleAssignmentsByGroupId($user->id) as $roleAssignment) {
+                $this->userHandler->removeRoleAssignment($roleAssignment->id);
+            }
+
```
barw4, line 628: @alongosz It works properly for users now, but we should do the same thing for User Groups. 

Deleting users properly invalidates cache now.

barw4 (changes requested):

eZ/Publish/Core/Persistence/Cache/UserHandler.php
alongosz: > @alongosz I think this is broken, loading roles doesn't work. Shouldn't it be done as above? A typo perhaps?

Good catch! Fixed.

eZ/Publish/Core/Repository/UserService.php
alongosz: > @alongosz It works properly for users now, but we should do the same thing for User Groups.
> 
> Deleting users properly invalidates cache now.

Added via [713a3ad](https://github.com/ezsystems/ezplatform-kernel/pull/374/commits/713a3ad703eaf068a6707ecef5fabbbf8dc8c248).

barw4: @alongosz hmm, it looks like it's not working. A count is not invalidated after deleting a user group.

It looks like the `deleteUserGroup` method is not used in the Back Office, but only in REST API. In the Back Office, User Group is treated as a 'normal' content, so it's basically trashed, hence, the `deleteUserGroup` method is not fired.

alongosz: > @alongosz hmm, it looks like it's not working. A count is not invalidated after deleting a user group.
> 
> It looks like the `deleteUserGroup` method is not used in the Back Office, but only in REST API. In the Back Office, User Group is treated as a 'normal' content, so it's basically trashed, hence, the `deleteUserGroup` method is not fired.

What's going to happen if you remove it from trash? Usually AdminUI should use dedicated strategy for user-related items.

barw4: Nothing really happens. It has a dedicated strategy only for Users, not for User Groups.

alongosz: > Nothing really happens. It has a dedicated strategy only for Users, not for User Groups.

Hmm, perhaps it requires fixing on AdminUI side. User Group should be deleted using User API, not Content API. So basically, for the fix you're working on, does this mean that if user group is assigned and gets removed via AdminUI, the count is incorrect and it affects the fix?

barw4: Yes, this is exactly the case. Obviously, it breaks only if we are using caching, so it works properly for the original solution without caching count and paginated list. 

Changing the way how we are deleting User Groups wouldn't be a bit of a design change? We would basically switch design here and this could be potentially breaking for some customers used to having groups in the trash for instance.

alongosz:
@barw4 I've dropped caching count method, but kept assignment removal fix anyway. If you have some more remarks let me know, otherwise you can merge this one into yours so it's easier to review for the others. I'd recommend "rebase and merge" in this case to keep the commits as-is for the review process ;)

barw4 (approved):

barw4:
@alongosz looks good to me, everything seems to be working properly now, thanks Andrzej.

PR #375, opened 2023-06-28
IBX-6017: Implemented `loadVersionInfoListByContentInfo` PAPI method

barw4:
| **JIRA issue**                          | [IBX-6017](https://issues.ibexa.co/browse/IBX-6017)

The newly implemented `loadVersionInfoListByContentInfo` allows loading latest `VersionInfo` of given Contents in batch.

Related `admin-ui` PR: https://github.com/ezsystems/ezplatform-admin-ui/pull/2104

#### Checklist:

### DIFF
--- a/eZ/Publish/API/Repository/ContentService.php
+++ b/eZ/Publish/API/Repository/ContentService.php
@@ -577,4 +577,17 @@ interface ContentService
      *        for a SiteAccess in a current context will be used.
      */
     public function count(Filter $filter, ?array $languages = null): int;
+
+    /**
+     * Bulk-load VersionInfo items by the list of ContentInfo Value Objects.
+     *
+     * @param array<\eZ\Publish\API\Repository\Values\Content\ContentInfo> $contentInfoList
+     *
+     * @return array<int, \eZ\Publish\API\Repository\Values\Content\VersionInfo> List of VersionInfo items with Content Ids as keys
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
+     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
+     */
+    public function loadVersionInfoListByContentInfo(array $contentInfoList): array;
 }
--- a/eZ/Publish/API/Repository/ContentService.php
+++ b/eZ/Publish/API/Repository/ContentService.php
@@ -100,6 +100,19 @@ interface ContentService
     public function loadVersionInfoById(int $contentId, ?int $versionNo = null): VersionInfo;
 
     /**
+     * Bulk-load VersionInfo items by the list of ContentInfo Value Objects.
+     *
+     * @param array<\eZ\Publish\API\Repository\Values\Content\ContentInfo> $contentInfoList
+     *
+     * @return array<int, \eZ\Publish\API\Repository\Values\Content\VersionInfo> List of VersionInfo items with Content Ids as keys
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
+     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
+     */
+    public function loadVersionInfoListByContentInfo(array $contentInfoList): array;
+
+    /**
      * Loads content in a version for the given content info object.
      *
      * If no version number is given, the method returns the current version
@@ -577,17 +590,4 @@ interface ContentService
      *        for a SiteAccess in a current context will be used.
      */
     public function count(Filter $filter, ?array $languages = null): int;
-
-    /**
-     * Bulk-load VersionInfo items by the list of ContentInfo Value Objects.
-     *
-     * @param array<\eZ\Publish\API\Repository\Values\Content\ContentInfo> $contentInfoList
-     *
-     * @return array<int, \eZ\Publish\API\Repository\Values\Content\VersionInfo> List of VersionInfo items with Content Ids as keys
-     *
-     * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
-     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
-     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
-     */
-    public function loadVersionInfoListByContentInfo(array $contentInfoList): array;
 }
--- a/eZ/Publish/API/Repository/Tests/ContentServiceTest.php
+++ b/eZ/Publish/API/Repository/Tests/ContentServiceTest.php
@@ -6685,4 +6685,33 @@ class ContentServiceTest extends BaseContentServiceTest
 
         return $contentWithReverseRelations;
     }
+
+    /**
+     * @covers \eZ\Publish\API\Repository\ContentService::LoadVersionInfoListByContentInfo
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
+     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
+     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
+     */
+    public function testLoadVersionInfoListByContentInfo(): void
+    {
+        $folder1 = $this->createFolder(['eng-GB' => 'Folder1'], 2);
+        $folder2 = $this->createFolder(['eng-GB' => 'Folder2'], 2);
+
+        $versionInfoList = $this->contentService->loadVersionInfoListByContentInfo([
+            $folder1->contentInfo,
+            $folder2->contentInfo,
+        ]);
+
+        self::assertCount(2, $versionInfoList);
+
+        foreach ($versionInfoList as $versionInfo) {
+            $loadedVersionInfo = $this->contentService->loadVersionInfo(
+                $versionInfo->getContentInfo(),
+                $versionInfo->versionNo
+            );
+            self::assertEquals($loadedVersionInfo, $versionInfo);
+        }
+    }
 }
--- a/eZ/Publish/API/Repository/Tests/ContentServiceTest.php
+++ b/eZ/Publish/API/Repository/Tests/ContentServiceTest.php
@@ -6685,33 +6685,4 @@ class ContentServiceTest extends BaseContentServiceTest
 
         return $contentWithReverseRelations;
     }
-
-    /**
-     * @covers \eZ\Publish\API\Repository\ContentService::LoadVersionInfoListByContentInfo
-     *
-     * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
-     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
-     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
-     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
-     */
-    public function testLoadVersionInfoListByContentInfo(): void
-    {
-        $folder1 = $this->createFolder(['eng-GB' => 'Folder1'], 2);
-        $folder2 = $this->createFolder(['eng-GB' => 'Folder2'], 2);
-
-        $versionInfoList = $this->contentService->loadVersionInfoListByContentInfo([
-            $folder1->contentInfo,
-            $folder2->contentInfo,
-        ]);
-
-        self::assertCount(2, $versionInfoList);
-
-        foreach ($versionInfoList as $versionInfo) {
-            $loadedVersionInfo = $this->contentService->loadVersionInfo(
-                $versionInfo->getContentInfo(),
-                $versionInfo->versionNo
-            );
-            self::assertEquals($loadedVersionInfo, $versionInfo);
-        }
-    }
 }
--- a/eZ/Publish/Core/Helper/TranslationHelper.php
+++ b/eZ/Publish/Core/Helper/TranslationHelper.php
@@ -69,7 +69,7 @@ class TranslationHelper
      *
      * @return string
      */
-    private function getTranslatedContentNameByVersionInfo(VersionInfo $versionInfo, $forcedLanguage = null)
+    public function getTranslatedContentNameByVersionInfo(VersionInfo $versionInfo, $forcedLanguage = null)
     {
         foreach ($this->getLanguages($forcedLanguage) as $lang) {
             $translatedName = $versionInfo->getName($lang);
--- a/eZ/Publish/Core/Helper/TranslationHelper.php
+++ b/eZ/Publish/Core/Helper/TranslationHelper.php
@@ -62,15 +62,12 @@ class TranslationHelper
 
     /**
      * Returns content name, translated, from a VersionInfo object.
-     * By default this method uses prioritized languages, unless $forcedLanguage is provided.
-     *
-     * @param \eZ\Publish\API\Repository\Values\Content\VersionInfo $versionInfo
-     * @param string $forcedLanguage
-     *
-     * @return string
+     * By default, this method uses prioritized languages, unless $forcedLanguage is provided.
      */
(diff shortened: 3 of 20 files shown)


eZ/Publish/API/Repository/ContentService.php
```diff
@@ -577,4 +577,17 @@ public function find(Filter $filter, ?array $languages = null): ContentList;
...
+     *
+     * @param array<\eZ\Publish\API\Repository\Values\Content\ContentInfo> $contentInfoList
+     *
+     * @return array<int, \eZ\Publish\API\Repository\Values\Content\VersionInfo> List of VersionInfo items with Content Ids as keys
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\BadStateException
+     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
+     */
+    public function loadVersionInfoListByContentInfo(array $contentInfoList): array;
```
alongosz: It's okay to place it after `loadVersionInfoById`, so all the methods related to VersionInfo loading are next to each other, for better DX of someone who's just looking through the interface.

eZ/Publish/Core/Helper/TranslationHelper.php
```diff
@@ -69,7 +69,7 @@ public function getTranslatedContentName(Content $content, $forcedLanguage = nul
      *
      * @return string
      */
-    private function getTranslatedContentNameByVersionInfo(VersionInfo $versionInfo, $forcedLanguage = null)
+    public function getTranslatedContentNameByVersionInfo(VersionInfo $versionInfo, $forcedLanguage = null)
```
alongosz: Since this was private, you should improve the overall quality of the method.
1. Get rid of `@param` and `@return` type-hints from the PHPDoc
2. Use strict `?string $forcedLanguage = null` parameter
3. Use strict return type `: string`

eZ/Publish/API/Repository/Tests/ContentServiceTest.php
```diff
@@ -6685,4 +6685,33 @@ private function createContentWithReverseRelations(array $drafts)
...
+        self::assertCount(2, $versionInfoList);
+
+        foreach ($versionInfoList as $versionInfo) {
+            $loadedVersionInfo = $this->contentService->loadVersionInfo(
+                $versionInfo->getContentInfo(),
+                $versionInfo->versionNo
+            );
+            self::assertEquals($loadedVersionInfo, $versionInfo);
+        }
+    }
```
alongosz: This is okay, but we're trying to retire legacy integration tests. Given there's a lot of them, any addition increases technical debt. It's fairly simple to add missing pieced to `IbexaKernelTestCase` so it works also for this old version of core.
See: https://github.com/ezsystems/ezplatform-kernel/compare/ibx-6017-implement-loadVersionInfoList...ibx-6017-test-using-ibexa-test-kernel

I added there base class `\Ibexa\Tests\Integration\Core\RepositoryTestCase` and test case `\Ibexa\Tests\Integration\Core\Repository\ContentService\LoadVersionInfoTest`, which is almost exact copy-paste of the test here. Notice that it's no longer one big ContentServiceTest class. That's because those big integration test cases have proven to be difficult to maintain. We need to start splitting them into groups.

eZ/Publish/Core/Persistence/Legacy/Tests/Content/Gateway/DoctrineDatabaseTest.php
```diff
@@ -2116,4 +2116,28 @@ private function assertContentVersionAttributesLanguages(
...
+        $results = $gateway->loadVersionInfoList([11]);
+
+        $orig = include __DIR__ . '/../_fixtures/extract_version_info_from_rows_multiple_versions.php';
+
+        $this->assertEquals(
+            [$orig[1]],
+            $results,
+            'Fixtures differ between what was previously stored(expected) and what it now generates(actual), this hints either some mistake in impl or that the fixture (../_fixtures/extract_content_from_rows_multiple_versions.php) and tests needs to be adapted.'
+        );
+    }
```
alongosz: This test doesn't bring enough value to keep it. It's redundant with an integration one, but actually tests less, because it's ran only against the one database engine we don't use - `sqlite`. Unit tests should never do any integration testing. Tests here are a big issue, we want actually to remove them once we make sure for each case there's corresponding proper integration coverage.

Additionally, if you see CI failures - there's a bug in this code reported by SonarCloud.

eZ/Publish/Core/Repository/ContentService.php
```diff
@@ -2526,4 +2526,22 @@ public function count(Filter $filter, ?array $languages = null): int
 
         return $this->contentFilteringHandler->count($filter);
     }
+
+    public function loadVersionInfoListByContentInfo(array $contentInfoList): array
+    {
+        $contentIds = array_column($contentInfoList, 'id');
```
alongosz: The main issue I have here is that there's no validation of user input, which is error-prone. I managed by accident to pass `VersionInfo` instead of `ContentInfo` and get some, but incorrect, results.

What I'd do here is more validation. Maybe there's better solution for that @ezsystems/php-dev-team? I was tempted to use `Webmozart\Assert\Assert::allIsInstanceOf`, but I actually prefer to still get API `InvalidArgumentException`, hence:

```php
foreach ($contentInfoList as $idx => $contentInfo) {
    if (!$contentInfo instanceof ContentInfo) {
        throw new InvalidArgumentException(
            '$contentInfoList',
            sprintf(
                'Element at position %d is not an instance of %s',
                $idx,
                $contentInfo
            )
        );
    }
}
```

alongosz: Moreover you're here relying on magic `id` property which is known to cause performance issue when scaled up.
Instead:
```php
$contentIds = array_map(
    static function (ContentInfo $contentInfo): int {
        return $contentInfo->getId();
    },
    $contentInfoList
);
```

eZ/Publish/SPI/Persistence/Content/Handler.php
```diff
@@ -364,4 +364,14 @@ public function deleteTranslationFromContent($contentId, $languageCode);
...
      */
     public function deleteTranslationFromDraft($contentId, $versionNo, $languageCode);
+
+    /**
+     * @param array<int> $contentIds
+     *
+     * @return array<\eZ\Publish\SPI\Persistence\Content\VersionInfo>
+     *
+     * @throws \eZ\Publish\Core\Base\Exceptions\InvalidArgumentException
+     * @throws \eZ\Publish\Core\Base\Exceptions\NotFoundException
```
alongosz, line 374: Please always use API contract exception, like `\eZ\Publish\API\Repository\Exceptions\NotFoundException` when reporting thrown exceptions.

eZ/Publish/Core/Persistence/Cache/ContentHandler.php
```diff
@@ -580,4 +580,37 @@ private function getCacheTagsForVersion(VersionInfo $versionInfo, array $tags =
 
         return $getContentInfoTagsFn($contentInfo, $tags);
     }
+
+    /**
+     * @throws \eZ\Publish\API\Repository\Exceptions\InvalidArgumentException
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
+     */
```
alongosz: You don't need to repeat this, it's inherited from the parent.

eZ/Publish/Core/Persistence/Legacy/Content/Gateway.php
```diff
@@ -459,4 +459,12 @@ abstract public function deleteTranslationFromVersion(
         int $versionNo,
         string $languageCode
     ): void;
+
+    /**
+     * @param array<int> $contentIds
+     *
+     * @throws \Doctrine\DBAL\Driver\Exception
+     * @throws \Doctrine\DBAL\Exception
```
alongosz: This should never be thrown by the ExceptionConversion layer, which is the last one aliasing this abstract. It's there for that reason. You can report here `\eZ\Publish\Core\Base\Exceptions\DatabaseException` thrown. If it indeed throws at least one of the mentioned exceptions, ExceptionConversion layer should take care about that.

eZ/Publish/Core/Persistence/Legacy/Content/Handler.php
```diff
@@ -904,4 +904,33 @@ static function ($lang) use ($languageCode) {
...
     }
+
+    /**
+     * @throws \eZ\Publish\Core\Base\Exceptions\NotFoundException
+     */
+    public function loadVersionInfoList(array $contentIds): array
+    {
+        $rows = $this->contentGateway->loadVersionInfoList($contentIds);
+        $mappedRows = array_map(
+            static function ($row) {
```
alongosz:
```suggestion
            static function (array $row): array {
```

eZ/Publish/Core/Persistence/Legacy/Tests/Content/ContentHandlerTest.php
```diff
@@ -1465,6 +1465,63 @@ public function testSetStatus()
         );
     }
 
+    /**
+     * @covers \eZ\Publish\Core\Persistence\Legacy\Content\Handler::loadVersionInfoList
+     *
+     * @throws \eZ\Publish\Core\Base\Exceptions\NotFoundException
```
alongosz:
```suggestion
     * @throws \eZ\Publish\API\Repository\Exceptions\Exception
```

alongosz (changes requested):
~Why is this targeted to 3.3 LTS if CS issue is for 4.5.0? This adds unnecessary maintenance costs for QA, cross merge-up, and maintenance.~ (resolved)

Remarks:

adamwojs:
@barw4 Could you please apply feedback from @alongosz ?

barw4:
> Why is this targeted to 3.3 LTS if CS issue is for 4.5.0? This adds unnecessary maintenance costs for QA, cross merge-up, and maintenance. The review anyway:

@alongosz we usually validate whether the same problem occurs in v3.3 if it occurs in v4.5 (and in most cases it does). I believe plenty of installations will benefit from it on v3.3 as it's still LTS even if no one reported it. I can scrap this PR if you wish but I see nothing wrong in improving performance for the active LTS version.

Steveb-p:
> > Why is this targeted to 3.3 LTS if CS issue is for 4.5.0? This adds unnecessary maintenance costs for QA, cross merge-up, and maintenance. The review anyway:
> 
> @alongosz we usually validate whether the same problem occurs in v3.3 if it occurs in v4.5 (and in most cases it does). I believe plenty of installations will benefit from it on v3.3 as it's still LTS even if no one reported it. I can scrap this PR if you wish but I see nothing wrong in improving performance for the active LTS version.

In theory 3.3 should not receive new functionality. This is done both to incentivize upgrading to a higher version and limit costs of maintenance, as Andrew pointed out.

However, given how we've dealt with it in the past, this performance fix can be accepted, as most of the work is done and I assume you will take care of merging it up later. Ultimately the decision is made by weighting our maintenance costs against benefits.

alongosz:
> > Why is this targeted to 3.3 LTS if CS issue is for 4.5.0? This adds unnecessary maintenance costs for QA, cross merge-up, and maintenance. The review anyway:
> 
> @alongosz we usually validate whether the same problem occurs in v3.3 if it occurs in v4.5 (and in most cases it does). I believe plenty of installations will benefit from it on v3.3 as it's still LTS even if no one reported it. I can scrap this PR if you wish but I see nothing wrong in improving performance for the active LTS version.

~if this was an update of internal business logic, that would be just a bug fix and no-brainer. But here you're introducing new API. It always impacts QA, maintenance, and Customers' upgrade process. For that reason I try to avoid packing features into old LTS, unless it's related to CS Customer version. It's fine if you believe it's needed, but in that case we need green light from either @adamwojs or @webhdx so they accept the risk and costs.~
Update: resolved.

alongosz:
@barw4 after brief discussion with @webhdx we can indeed add this to 3.3 given the amount of work you've put into this and benefits for LTS :muscle:
Please proceed with the remaining review remarks: https://github.com/ezsystems/ezplatform-kernel/pull/375#pullrequestreview-1531431049. Feel free to reach out to me in case of questions :)

barw4:
@alongosz thank you for the heads-up, on it

eZ/Publish/Core/Helper/TranslationHelper.php
barw4: @alongosz this one got me confused. Shouldn't we keep those old type-hints as a part of BC policy, even if method itself was private? I mean, this method is already used in the `getTranslatedContentName` method and the `getTranslatedContentName` doesn't have strict type-hints.

eZ/Publish/Core/Persistence/Legacy/Tests/Content/Gateway/DoctrineDatabaseTest.php
barw4: Regarding this bug, changing it to `include_once` resulted in some other exceptions, so I kept it as in the methods above. Nevertheless, I removed this method.

eZ/Publish/Core/Repository/ContentService.php
barw4: Good point definitely. The same problem can occur in the `loadContentListByContentInfo`.

eZ/Publish/SPI/Persistence/Content/Handler.php
barw4: Oops, sorry, PHPStorm auto completion

barw4:
@alongosz almost all your remarks are applied, pinging for re-review

eZ/Publish/Core/Helper/TranslationHelper.php
alongosz: > @alongosz this one got me confused. Shouldn't we keep those old type-hints as a part of BC policy, even if method itself was private? I mean, this method is already used in the `getTranslatedContentName` method and the `getTranslatedContentName` doesn't have strict type-hints.

@barw4 It always depends on a specific case. If the method was private, there's nothing wrong with changing type hints as long as other methods which are invoking it are ready for that. You can even completely get rid of that method and replace it with something else. If still it's doubtful (because of e.g. the usage in `getTranslatedContentName`), we need to look at that usage and consequences of the change. If anyone passes anything else there, then it's a serious issue that can be detected thanks to the change. The code won't work if you pass something that isn't "stringable" as both an argument and `VersionInfo::getName` result (meaning it would've already crashed for someone doing that).

Someone could still argue that wrong input shouldn't result in a crash [1]. Fortunately we're in not any danger here. The thing is - this class doesn't have strict types declared. That means that even if we passed or returned e.g. `int(1)` it still would've been cast to `string('1')`.

Last but not least, we need to assess overall risks and impact to be able gradually reduce technical debt. In this case those are minimal and it's possible that at some point the entire class will get an overhaul, within reason. Now, that the method has become, public we're gonna need to be more careful about any changes to its signature, so right now is the best moment to do it.

-----------
[1] 
> On two occasions I have been asked, - "Pray, Mr. Babbage, if you put into the machine wrong figures, will the right answers come out?"
I am not able rightly to apprehend the kind of confusion of ideas that could provoke such a question

~ Charles Babbage

alongosz (commented):
Replied to the last issue here: https://github.com/ezsystems/ezplatform-kernel/pull/375#discussion_r1267343591

eZ/Publish/Core/Helper/TranslationHelper.php
barw4: @alongosz sounds fair to me, thank you for the explanation, resolved in https://github.com/ezsystems/ezplatform-kernel/pull/375/commits/b495c5c164577d506cfaba803c07c36033919bcc

alongosz (approved):

konradoboza (approved):

kisztof (approved):

micszo (approved):
Tested with ca. 5000k landing pages with 100+ fields. Used tree built with folders with depth 3 (landing pages on 3rd level). One parent had 2, 5, 10, 50, 100 and 500 items.

Tried extreme Content tree config with `load_more_limit` 500 and `children_load_max_limit` 5000.
With the fix observed a gain in time from 18-19 sec to ca. 11 sec.

With default Content tree config (`load_more_limit` 30 and `children_load_max_limit` 200) the impact was not as big but also clear.
With the fixed observed a gain in time from 11-12 sec to 8-9 sec.

QA Approved on Ibexa Experience 3.3.34-dev.

PR #380, opened 2023-08-01
IBX-4031: Forced copy non-translatable field values from main content to current draft during publish

barw4:
| **JIRA issue**                          | [IBX-4031](https://issues.ibexa.co/browse/IBX-4031)

Following PR assures that a content that is being published will have updated values for its non-translatable fields taken from main published version.

Related PR: https://github.com/ezsystems/ezplatform-content-forms/pull/71

#### Checklist:


eZ/Publish/API/Repository/Tests/ContentServiceTest.php
```diff
@@ -6524,6 +6524,130 @@ public function testCopyTranslationsFromInvalidPublishedContentToDraft()
...
+        $repository = $this->getRepository();
+        $permissionResolver = $repository->getPermissionResolver();
+        $contentTypeService = $repository->getContentTypeService();
+
+        $typeCreate = $contentTypeService->newContentTypeCreateStruct('nontranslatable');
+
+        $typeCreate->mainLanguageCode = 'eng-GB';
+        $typeCreate->remoteId = '1234567890abcdef';
+        $typeCreate->urlAliasSchema = 'url|scheme';
+        $typeCreate->nameSchema = 'name|scheme';
```
alongosz: This doesn't make practical sense. It would generate fixed `url%7Cscheme` URL Alias. It's irrelevant for the test, but it could be confusing.
It should be something like `<title>` for this particular CT.

eZ/Publish/API/Repository/Tests/ContentServiceTest.php
```diff
@@ -6524,6 +6524,130 @@ public function testCopyTranslationsFromInvalidPublishedContentToDraft()
...
+        $bodyFieldCreate->isSearchable = true;
+        $bodyFieldCreate->defaultValue = null;
+
+        $typeCreate->addFieldDefinition($bodyFieldCreate);
+
+        $contentTypeDraft = $contentTypeService->createContentType(
+            $typeCreate,
+            [$contentTypeService->loadContentTypeGroupByIdentifier('Media')],
+        );
+        $contentTypeService->publishContentTypeDraft($contentTypeDraft);
```
alongosz: could you extract creating content type to some separate method? Cognitive complexity here is very high which makes it unreadable.

Nattfarinn (approved):

ViniTou (approved):

ViniTou:
It seems that for some reason, maybe totally unrelated, integration tests are failing at random here. Could you check if those changes could introduce that behavior?  

Also, not sure, but maybe it is reasonable to move that new integration test into new integration setup based on TestKernel? @alongosz

eZ/Publish/API/Repository/Tests/ContentServiceTest.php
```diff
@@ -6524,6 +6524,130 @@ public function testCopyTranslationsFromInvalidPublishedContentToDraft()
...
+
+        // Publishing ger-DE draft with the empty non-translatable field
+        $this->contentService->publishVersion($gerContent->getVersionInfo());
+
+        // Loading main content
+        $mainPublishedContent = $this->contentService->loadContent($engContent->id);
+        $bodyFieldValue = $mainPublishedContent->getField('body')->getValue();
+
+        self::assertSame($expectedBodyValue, $bodyFieldValue->text);
+    }
```
alongosz: > maybe it is reasonable to move that new integration test into new integration setup based on TestKernel?

yes, it should be a test in the new namespace using new IbexaTestKernel. It could be placed in [`./tests/integration/Core/Repository/ContentService/`](https://github.com/ibexa/core/tree/main/tests/integration/Core/Repository/ContentService/) to avoid conflicts, since integration tests don't need to follow strictly `ClassNameTest.php` pattern.

@barw4 we need to start separating and refactoring those test cases anyway. ContentServiceTest is almost 7k lines long and it starts to be difficult for PHPStorm to load efficiently all symbols. You can base the test on `./tests/integration/Core/RepositoryTestCase.php` but I think you need to rebase against fresh 1.3 branch first.

eZ/Publish/Core/Repository/ContentService.php
```diff
@@ -1468,6 +1469,71 @@ public function publishVersion(APIVersionInfo $versionInfo, array $translations
         return $content;
     }
 
+    protected function copyNonTranslatableFieldsFromPublishedVersion(APIContent $currentVersionContent): void
```
alongosz, line 1512:
```suggestion
    private function copyNonTranslatableFieldsFromPublishedVersion(APIContent $currentVersionContent): void
```

eZ/Publish/Core/Repository/ContentService.php
```diff
@@ -1468,6 +1469,71 @@ public function publishVersion(APIVersionInfo $versionInfo, array $translations
...
+    {
+        $versionInfo = $currentVersionContent->getVersionInfo();
+        $contentType = $currentVersionContent->getContentType();
+
+        $mainContent = $this->internalLoadContentById($versionInfo->getContentInfo()->getId());
```
alongosz: 1. What's going to happen if the loaded Content doesn't have any published Version yet? See conditions in `copyTranslationsFromPublishedVersion`.
2. Naming: following method name and the remark above this could be `$publishedContent`, right?

eZ/Publish/Core/Repository/ContentService.php
alongosz, line 1512: This method should report throwing `\eZ\Publish\API\Repository\Exceptions\NotFoundException`.

eZ/Publish/Core/Repository/ContentService.php
```diff
@@ -1468,6 +1469,71 @@ public function publishVersion(APIVersionInfo $versionInfo, array $translations
...
+        $mainContentFieldsInMainLanguage = $mainContent->getFieldsByLanguage(
+            $mainContent->getVersionInfo()->getContentInfo()->getMainLanguageCode()
+        );
+
+        $fieldValues = [];
+        $spiFields = [];
```
alongosz: Naming: `$persistenceFields`. That was never correct in all currently existing usages.

eZ/Publish/Core/Repository/ContentService.php
```diff
@@ -1468,6 +1469,71 @@ public function publishVersion(APIVersionInfo $versionInfo, array $translations
...
+        foreach ($currentVersionContent->getFields() as $field) {
+            $fieldDefinition = $contentType->getFieldDefinition($field->fieldDefIdentifier);
+            $fieldValues[$fieldDefinition->identifier][$field->languageCode] = $field->getValue();
```
alongosz: This line can be moved after if-continue block below. Reason: there's no way to know if `getValue` is heavy or not as it can be custom, so let's not do it if it's not needed.

eZ/Publish/Core/Repository/ContentService.php
```diff
@@ -1468,6 +1469,71 @@ public function publishVersion(APIVersionInfo $versionInfo, array $translations
...
+                || $field->languageCode === $versionInfo->initialLanguageCode
+            ) {
+                continue;
+            }
+
+            $fieldType = $this->fieldTypeRegistry->getFieldType(
+                $fieldDefinition->fieldTypeIdentifier
+            );
+
+            $newValue = $mainContentFieldsInMainLanguage[$field->fieldDefIdentifier]->value;
```
alongosz: no magic please
```suggestion
            $newValue = $mainContentFieldsInMainLanguage[$field->fieldDefIdentifier]->getValue();
```

eZ/Publish/Core/Repository/ContentService.php
```diff
@@ -1468,6 +1469,71 @@ public function publishVersion(APIVersionInfo $versionInfo, array $translations
...
+
+        $fieldValues = [];
+        $spiFields = [];
+        foreach ($currentVersionContent->getFields() as $field) {
+            $fieldDefinition = $contentType->getFieldDefinition($field->fieldDefIdentifier);
+            $fieldValues[$fieldDefinition->identifier][$field->languageCode] = $field->getValue();
+
+            if (
+                $fieldDefinition->isTranslatable
+                || $field->languageCode === $versionInfo->initialLanguageCode
```
alongosz: Q: what if Main Language code is different than `$versionInfo->initialLanguageCode`? Does that require copying as well? Not sure, asking.

eZ/Publish/Core/Repository/ContentService.php
```diff
@@ -1468,6 +1469,71 @@ public function publishVersion(APIVersionInfo $versionInfo, array $translations
...
+            );
+        }
+
+        $updateStruct = new SPIContentUpdateStruct();
+        $updateStruct->name = $this->nameSchemaService->resolveNameSchema(
+            $currentVersionContent,
+            $fieldValues,
+            $versionInfo->languageCodes,
+            $contentType
+        );
```
alongosz, line 1573: Is it necessary to resolve name again when dealing with non-translatable fields?

barw4: I think it was necessary as non-translatable field could be a part of name schema. However, this is irrelevant after https://github.com/ezsystems/ezplatform-kernel/pull/380/commits/036c05d50cc0464dbaabbf15f8090bf925d972e3

eZ/Publish/Core/Repository/ContentService.php
barw4: Irrelevant after https://github.com/ezsystems/ezplatform-kernel/commit/036c05d50cc0464dbaabbf15f8090bf925d972e3

barw4: @alongosz I don't think this should be moved. We have to resolve a name again for content in the `$updateStruct` based on those `$fieldValues`:
```
$updateStruct->name = $this->nameSchemaService->resolveNameSchema(..)
```

eZ/Publish/Core/Repository/ContentService.php
barw4: Actually, we still have to use `SPIContentUpdateStruct`, rollbacking.

eZ/Publish/Core/Repository/ContentService.php
barw4: @alongosz yes, I think it requires copying. `$versionInfo->initialLanguageCode` is a language that for example we chose translating content into and when `mainLanguageCode` is different it means that content having this `mainLanguageCode` has original non-translatable translation that we need to copy.

eZ/Publish/API/Repository/Tests/ContentServiceTest.php
barw4: @alongosz should be all good now!

tests/integration/Core/Repository/ContentService/CopyNonTranslatableFieldsFromPublishedVersionTest.php
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * @covers \eZ\Publish\API\Repository\ContentService
+ */
+final class CopyNonTranslatableFieldsFromPublishedVersionTest extends RepositoryTestCase
+{
+    private const GER_DE = 'ger-DE';
+    private const ENG_US = 'eng-US';
+    private const CONTENT_TYPE_IDENTIFIER = 'nontranslatable';
+
+    public function testCopyNonTranslatableFieldsFromPublishedVersionToDraft(): void
```
alongosz, line 29:
```suggestion
    /**
     * @throws \eZ\Publish\API\Repository\Exceptions\Exception
     */
    public function testCopyNonTranslatableFieldsFromPublishedVersionToDraft(): void
```

eZ/Publish/Core/Repository/ContentService.php
alongosz: [This comment](https://github.com/ezsystems/ezplatform-kernel/pull/380/files#r1284226846) has not been resolved.

tests/integration/Core/Repository/ContentService/CopyNonTranslatableFieldsFromPublishedVersionTest.php
```diff
@@ -0,0 +1,156 @@
...
+            'StringLengthValidator' => [
+                'minStringLength' => 0,
+                'maxStringLength' => 0,
+            ],
+        ];
+        $titleFieldCreate->fieldSettings = [];
+        $titleFieldCreate->isSearchable = true;
+        $titleFieldCreate->defaultValue = 'default title';
+
+        $typeCreate->addFieldDefinition($titleFieldCreate);
```
alongosz: This fragment feels still a bit too similar to L126-148. What if we extracted variable parts and created separate method for that, so the entire block L102-148 is more compact?

```php
        $fieldDefinitionPosition = 1;
        $typeCreate->addFieldDefinition(
            $this->buildFieldDefinitionCreateStructForNonTranslatableContentType(
                $fieldDefinitionPosition,
                'title',
                ['eng-GB' => 'Title'],
                true,
                true,
                'default title'
            )
        );

        $typeCreate->addFieldDefinition(
            $this->buildFieldDefinitionCreateStructForNonTranslatableContentType(
                ++$fieldDefinitionPosition,
                'body',
                ['eng-GB' => 'Body'],
                false,
                false
            )
        );
```
and the method itself:
```php
// private const TEXT_LINE_FIELD_TYPE_IDENTIFIER = 'ezstring';

    /**
     * @param array<string, string> $names
     */
    private function buildFieldDefinitionCreateStructForNonTranslatableContentType(
        int $position,
        string $fieldIdentifier,
        array $names,
        bool $isTranslatable,
        bool $isRequired,
        ?string $defaultValue = null
    ): FieldDefinitionCreateStruct {
        $contentTypeService = self::getContentTypeService();

        $fieldDefinitionCreateStruct = $contentTypeService->newFieldDefinitionCreateStruct(
            $fieldIdentifier,
            self::TEXT_LINE_FIELD_TYPE_IDENTIFIER
        );

        $fieldDefinitionCreateStruct->names = $names;
        $fieldDefinitionCreateStruct->descriptions = $names;
        $fieldDefinitionCreateStruct->fieldGroup = 'content';
        $fieldDefinitionCreateStruct->position = $position;
        $fieldDefinitionCreateStruct->isTranslatable = $isTranslatable;
        $fieldDefinitionCreateStruct->isRequired = $isRequired;
        $fieldDefinitionCreateStruct->isInfoCollector = false;
        $fieldDefinitionCreateStruct->validatorConfiguration = [
            'StringLengthValidator' => [
                'minStringLength' => 0,
                'maxStringLength' => 0,
            ],
        ];
        $fieldDefinitionCreateStruct->fieldSettings = [];
        $fieldDefinitionCreateStruct->isSearchable = true;
        $fieldDefinitionCreateStruct->defaultValue = $defaultValue;

        return $fieldDefinitionCreateStruct;
    }
```

I'm keeping common parts hardcoded. In the future either `RepositoryTestCase` or dedicated VO builder could have more flexible (less hardcoded) logic.

That should also conveniently get rid of SonarCloud Quality Gate failure.

alongosz (commented):
@barw4 looks very good. I have some small remarks and an idea which should workaround SonarCloud code duplication issue for now:

tests/integration/Core/Repository/ContentService/CopyNonTranslatableFieldsFromPublishedVersionTest.php
barw4: @alongosz sure, sounds good to me, please see https://github.com/ezsystems/ezplatform-kernel/pull/380/commits/7635e1413108327cba78236465620408490b55dd

tests/integration/Core/Repository/ContentService/CopyNonTranslatableFieldsFromPublishedVersionTest.php
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace Ibexa\Tests\Integration\Core\Repository\ContentService;
+
+use Datetime;
```
alongosz: @barw4 
```suggestion
use DateTime;
```

alongosz (approved):

eZ/Publish/Core/Repository/ContentService.php
```diff
@@ -1505,6 +1506,88 @@ public function publishVersion(APIVersionInfo $versionInfo, array $translations
...
+    {
+        $versionInfo = $currentVersionContent->getVersionInfo();
+        $contentType = $currentVersionContent->getContentType();
+
+        $publishedContent = $this->internalLoadContentById($versionInfo->getContentInfo()->getId());
+        $publishedVersionInfo = $publishedContent->getVersionInfo();
+
+        if (
+            !$publishedVersionInfo->isPublished()
+            ||
```
konradoboza:
```suggestion
            !$publishedVersionInfo->isPublished() ||
```

eZ/Publish/Core/Repository/ContentService.php
```diff
@@ -1505,6 +1506,88 @@ public function publishVersion(APIVersionInfo $versionInfo, array $translations
...
+
+        $fieldValues = [];
+        $persistenceFields = [];
+        foreach ($currentVersionContent->getFields() as $field) {
+            $fieldDefinition = $contentType->getFieldDefinition($field->fieldDefIdentifier);
+            $fieldValues[$fieldDefinition->identifier][$field->languageCode] = $field->getValue();
+
+            if (
+                $fieldDefinition->isTranslatable
+                || $field->languageCode === $versionInfo->initialLanguageCode
```
konradoboza, line 1539:
```suggestion
                $fieldDefinition->isTranslatable ||
                $field->languageCode === $versionInfo->initialLanguageCode
```

tests/integration/Core/Repository/ContentService/CopyNonTranslatableFieldsFromPublishedVersionTest.php
```diff
@@ -0,0 +1,173 @@
...
+        $contentTypeService = self::getContentTypeService();
+
+        $typeCreate = $contentTypeService->newContentTypeCreateStruct(self::CONTENT_TYPE_IDENTIFIER);
+
+        $typeCreate->mainLanguageCode = 'eng-GB';
+        $typeCreate->remoteId = '1234567890abcdef';
+        $typeCreate->urlAliasSchema = '<title>';
+        $typeCreate->nameSchema = '<title>';
+        $typeCreate->names = [
+            'eng-GB' => 'Nontranslatable content type',
```
konradoboza:
```suggestion
            'eng-GB' => 'Non-translatable content type',
```

konradoboza (approved):
Optional nitpicking:

eZ/Publish/Core/Repository/ContentService.php
barw4: @konradoboza In most (or in all) places we tend to use the `||` operator in a new line. https://github.com/ibexa/core/blob/main/src/lib/MVC/Symfony/Security/EventListener/SecurityListener.php#L229

eZ/Publish/Core/Repository/ContentService.php
barw4: @konradoboza same as above

eZ/Publish/Core/Repository/ContentService.php
konradoboza: That still looks better, feel free to adjust. As mentioned, this isn't a must.

eZ/Publish/Core/Repository/ContentService.php
konradoboza: Fine with me.

eZ/Publish/Core/Repository/ContentService.php
barw4: https://pear.php.net/manual/en/rfc.cs-enhancements.splitlongstatements.php - seems like this is most common practice

konradoboza: It's still different than having logical operator in its own line, but as mentioned - that's not a must.

kisztof (approved):

bogusez:
Regression tests passed:
https://github.com/ibexa/oss/pull/109

bogusez (approved):

PR #382, opened 2023-08-08
EZP-29574: [Tests] Moved testLoadUserByTokenThrowsNotFoundException after tests depending of testLoadUserByToken

mateuszdebinski:
| **JIRA issue**                          | [EZP-29574](https://issues.ibexa.co/browse/EZP-29574)

After re-analyzing the tests that sometimes throw the problem, I probably found the cause. `testLoadUserByTokenThrowsNotFoundException` was fired right after `testLoadUserByToken` and has `$repository = $this->getRepository();` which clears the database. In `testLoadUserByTokenThrowsNotFoundException`, we also created a user token with a time set to the current time that is not needed for this test. Therefore the exception about not finding the token was thrown randomly if the seconds changed between `testLoadUserByTokenThrowsNotFoundException` and `testUpdateUserToken`

#### Checklist:

### DIFF
--- a/eZ/Publish/API/Repository/Tests/UserServiceTest.php
+++ b/eZ/Publish/API/Repository/Tests/UserServiceTest.php
@@ -2972,29 +2972,6 @@ class UserServiceTest extends BaseTest
     }
 
     /**
-     * Test trying to load User by invalid Token.
-     *
-     * @covers \eZ\Publish\API\Repository\UserService::loadUserByToken
-     */
-    public function testLoadUserByTokenThrowsNotFoundException()
-    {
-        $this->expectException(NotFoundException::class);
-
-        $repository = $this->getRepository();
-        $userService = $repository->getUserService();
-
-        $user = $this->createUserVersion1();
-
-        $userTokenUpdateStruct = new UserTokenUpdateStruct();
-        $userTokenUpdateStruct->hashKey = md5('hash');
-        $userTokenUpdateStruct->time = new DateTime();
-
-        $userService->updateUserToken($user, $userTokenUpdateStruct);
-
-        $userService->loadUserByToken('not_existing_token');
-    }
-
-    /**
      * Test updating User Token.
      *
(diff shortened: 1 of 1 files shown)


eZ/Publish/API/Repository/Tests/UserServiceTest.php
```diff
@@ -3045,6 +3022,21 @@ public function testExpireUserToken($userToken)
...
+     * @covers \eZ\Publish\API\Repository\UserService::loadUserByToken
+     */
+    public function testLoadUserByTokenThrowsNotFoundException()
+    {
+        $this->expectException(NotFoundException::class);
+
+        $repository = $this->getRepository();
+        $userService = $repository->getUserService();
+
+        $userService->loadUserByToken('not_existing_token');
```
alongosz: `expectExpection` statement should be placed right before the statement which is expected to trigger that exception

eZ/Publish/API/Repository/Tests/UserServiceTest.php
```diff
@@ -3045,6 +3022,20 @@ public function testExpireUserToken($userToken)
         $userService->loadUserByToken($userToken);
     }
 
+    /**
+     * Test trying to load User by invalid Token.
+     *
+     * @covers \eZ\Publish\API\Repository\UserService::loadUserByToken
+     */
+    public function testLoadUserByTokenThrowsNotFoundException()
```
alongosz: Return type - missed this initially, my bad
```suggestion
    public function testLoadUserByTokenThrowsNotFoundException(): void
```

alongosz (approved):
Small step for an Engineer, big step for Kernel :trophy:

konradoboza (approved):

ciastektk (approved):

ViniTou (approved):

Nattfarinn (approved):

kisztof (approved):

PR #385, opened 2023-10-02
IBX-3957: Made NOP URL aliases not reusable and original

barw4:
| **JIRA issue**                          | [IBX-3957](https://issues.ibexa.co/browse/IBX-3957)
| **BC breaks**                          | ?

Reasoning: because `nop` aliases are **not** original and they are reusable creating another alias with the same name will break such an alias, however, imo it should be incremented instead. This will prevent creating exactly the same alias with the same parent and as a result, breaking the old one.

#### Checklist:

### DIFF
--- a/eZ/Publish/API/Repository/Tests/URLAliasServiceTest.php
+++ b/eZ/Publish/API/Repository/Tests/URLAliasServiceTest.php
@@ -1603,7 +1603,6 @@ class URLAliasServiceTest extends BaseTest
     public function testRenamingParentContentDoesntBreakChildAlias(): void
     {
         $repository = $this->getRepository();
-        $locationService = $repository->getLocationService();
         $urlAliasService = $repository->getURLAliasService();
         $contentService = $repository->getContentService();
 
--- a/eZ/Publish/API/Repository/Tests/URLAliasServiceTest.php
+++ b/eZ/Publish/API/Repository/Tests/URLAliasServiceTest.php
@@ -1605,50 +1605,45 @@ class URLAliasServiceTest extends BaseTest
         $repository = $this->getRepository();
         $locationService = $repository->getLocationService();
         $urlAliasService = $repository->getURLAliasService();
-        $contentTypeService = $repository->getContentTypeService();
         $contentService = $repository->getContentService();
 
         $languageCode = 'eng-GB';
-        $contentType = $contentTypeService->loadContentTypeByIdentifier('folder');
 
         // 1. Create parent folder
-        $folderCreateStruct = $contentService->newContentCreateStruct($contentType, $languageCode);
-        $folderCreateStruct->setField('name', 'A');
-        $folderDraft = $contentService->createContent($folderCreateStruct, [
-            $locationService->newLocationCreateStruct(2),
-        ]);
-        $folder = $contentService->publishVersion($folderDraft->getVersionInfo());
-        $folderLocation = $locationService->loadLocation($folder->contentInfo->getMainLocationId());
+        $folder = $this->createFolder([$languageCode => 'a'], 2);
+        $folderLocationId = $folder->contentInfo->getMainLocationId();
 
         // 2. Create child folder
-        $childCreateStruct = $contentService->newContentCreateStruct($contentType, $languageCode);
-        $childCreateStruct->setField('title', 'B');
-        $childDraft = $contentService->createContent($folderCreateStruct, [
-            $locationService->newLocationCreateStruct($folderLocation->id),
-        ]);
-        $child = $contentService->publishVersion($childDraft->getVersionInfo());
-        $childLocation = $locationService->loadLocation($child->contentInfo->getMainLocationId());
+        $child = $this->createFolder([$languageCode => 'b'], $folderLocationId);
+        $childLocationId = $child->contentInfo->getMainLocationId();
+        $childLocation = $locationService->loadLocation($childLocationId);
 
         // 3. Create custom URL alias for child folder
         $urlAliasService->createUrlAlias($childLocation, '/c/b', $languageCode);
         $lookup = $urlAliasService->lookup('/c/b');
 
         self::assertSame('/c/b', $lookup->path);
+        self::assertSame($childLocationId, $lookup->destination);
 
         // 4. Rename "A" to "C"
         $folderDraft = $contentService->createContentDraft($folder->contentInfo);
         $folderUpdateStruct = $contentService->newContentUpdateStruct();
-        $folderUpdateStruct->setField('name', 'C');
+        $folderUpdateStruct->setField('name', 'c');
         $renamedFolder = $contentService->updateContent($folderDraft->getVersionInfo(), $folderUpdateStruct);
         $contentService->publishVersion($renamedFolder->getVersionInfo());
 
         // Loading aliases shouldn't throw a `BadStateException`
-        $urlAliasService->listLocationAliases($childLocation);
+        $childLocationAliases = $urlAliasService->listLocationAliases($childLocation);
+
+        self::assertCount(1, $childLocationAliases);
+        self::assertSame('/c/b', $childLocationAliases[0]->path);
 
         // Renamed content should have '/c2' path alias
-        $lookupRenamed = $urlAliasService->lookup('C2');
+        $lookupRenamed = $urlAliasService->lookup('c2');
+        $originalLookup = $urlAliasService->lookup('/c/b');
 
-        self::assertSame('/C2', $lookupRenamed->path);
+        self::assertSame($childLocationId, $originalLookup->destination);
+        self::assertSame('/c2', $lookupRenamed->path);
     }
 
     /**
--- a/eZ/Publish/API/Repository/Tests/URLAliasServiceTest.php
+++ b/eZ/Publish/API/Repository/Tests/URLAliasServiceTest.php
@@ -1615,8 +1615,8 @@ class URLAliasServiceTest extends BaseTest
 
         // 2. Create child folder
         $child = $this->createFolder([$languageCode => 'b'], $folderLocationId);
-        $childLocationId = $child->contentInfo->getMainLocationId();
-        $childLocation = $locationService->loadLocation($childLocationId);
+        $childLocation = $child->getVersionInfo()->getContentInfo()->getMainLocation();
+        $childLocationId = $childLocation->id;
 
         // 3. Create custom URL alias for child folder
         $urlAliasService->createUrlAlias($childLocation, '/c/b', $languageCode);
--- a/eZ/Publish/API/Repository/Tests/URLAliasServiceTest.php
+++ b/eZ/Publish/API/Repository/Tests/URLAliasServiceTest.php
@@ -1600,6 +1600,57 @@ class URLAliasServiceTest extends BaseTest
         );
     }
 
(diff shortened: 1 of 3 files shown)


alongosz:
@barw4 as discussed internally, please add some integration coverage for the case this issue solves so I can analyse this further. At this point I'm not sure what would be the side effects here. Looks like we're just missing some coverage.

barw4:
@alongosz an integration test is added, please see https://github.com/ezsystems/ezplatform-kernel/pull/385/commits/2b9b4b78fe0898f2ddb8f1d73234517c5a26e144

eZ/Publish/Core/Persistence/Legacy/Content/UrlAlias/Handler.php
```diff
@@ -234,7 +234,6 @@ private function internalPublishUrlAliasForLocation(
             // 2. existing location or custom alias entry
             // 3. history entry
```
konradoboza: Not sure about the initial intention behind this, but this comment should be updated too concerning your changes.

eZ/Publish/API/Repository/Tests/URLAliasServiceTest.php
```diff
@@ -1600,6 +1600,57 @@ public function testDeleteCorruptedUrlAliases()
...
+        ]);
+        $folder = $contentService->publishVersion($folderDraft->getVersionInfo());
+        $folderLocation = $locationService->loadLocation($folder->contentInfo->getMainLocationId());
+
+        // 2. Create child folder
+        $childCreateStruct = $contentService->newContentCreateStruct($contentType, $languageCode);
+        $childCreateStruct->setField('title', 'B');
+        $childDraft = $contentService->createContent($folderCreateStruct, [
+            $locationService->newLocationCreateStruct($folderLocation->id),
+        ]);
```
alongosz: This part doesn't add up, you don't have `title` field in folder. I got confused at first why this test passes, but notice that you're actually not using `childCreateStruct` when creating content.

There's `createFolder` helper method available here so you can make that code very compact.

eZ/Publish/API/Repository/Tests/URLAliasServiceTest.php
```diff
@@ -1600,6 +1600,57 @@ public function testDeleteCorruptedUrlAliases()
...
+        $contentType = $contentTypeService->loadContentTypeByIdentifier('folder');
+
+        // 1. Create parent folder
+        $folderCreateStruct = $contentService->newContentCreateStruct($contentType, $languageCode);
+        $folderCreateStruct->setField('name', 'A');
+        $folderDraft = $contentService->createContent($folderCreateStruct, [
+            $locationService->newLocationCreateStruct(2),
+        ]);
+        $folder = $contentService->publishVersion($folderDraft->getVersionInfo());
+        $folderLocation = $locationService->loadLocation($folder->contentInfo->getMainLocationId());
```
alongosz: you're loading location now just to use it later as `$folderLocation->id`. This is not necessary. Just do `$folder->getVersionInfo()->getContentInfo()->getMainLocationId()` instead please.

eZ/Publish/API/Repository/Tests/URLAliasServiceTest.php
```diff
@@ -1600,6 +1600,57 @@ public function testDeleteCorruptedUrlAliases()
...
+            $locationService->newLocationCreateStruct($folderLocation->id),
+        ]);
+        $child = $contentService->publishVersion($childDraft->getVersionInfo());
+        $childLocation = $locationService->loadLocation($child->contentInfo->getMainLocationId());
+
+        // 3. Create custom URL alias for child folder
+        $urlAliasService->createUrlAlias($childLocation, '/c/b', $languageCode);
+        $lookup = $urlAliasService->lookup('/c/b');
+
+        self::assertSame('/c/b', $lookup->path);
```
alongosz, line 1624: Please use everywhere either capitalized or small letters. Kernel tests treat `C` and `c` differently because there's probably no configuration to normalize that (on full stack app you'll always get small caps in urls).

alongosz, line 1624: Wouldn't it also make sense here to check that `$lookup` points to a proper content item?

kisztof (approved):

eZ/Publish/API/Repository/Tests/URLAliasServiceTest.php
```diff
@@ -1600,6 +1600,57 @@ public function testDeleteCorruptedUrlAliases()
...
+        $renamedFolder = $contentService->updateContent($folderDraft->getVersionInfo(), $folderUpdateStruct);
+        $contentService->publishVersion($renamedFolder->getVersionInfo());
+
+        // Loading aliases shouldn't throw a `BadStateException`
+        $urlAliasService->listLocationAliases($childLocation);
+
+        // Renamed content should have '/c2' path alias
+        $lookupRenamed = $urlAliasService->lookup('C2');
+
+        self::assertSame('/C2', $lookupRenamed->path);
```
alongosz: So now I know what's happening here. The end result is unfortunately the opposite of what the PR describes. A **reusabe** entry is the same entry either used for multiple purposes or that was repurposed. The changes you made here made effectively NOP entries NOT reusable. They're left behind as-is and the secondary entry gets created and named C2 according to the unique naming URL Alias logic.

The flow after the changes:

1. `/a/b` -> normal URL alias, `a` and `b` corresponding content items exist
1. `/c/b` -> custom URL alias, `c` is `NOP` because there's no `c` corresponding content
1. Rename `a` to `c`
1. `/c2` now points to renamed content, `/c` is still a NOP entry therefore it wasn't reused.

What we'd want if we wanted to keep re-usability logic would be:

3. Rename `a` to `c`
4. `/c/b` becomes normal URL alias since `c` corresponds now to proper content node
5. `/a/b` -> is a historical entry redirecting to `/c/b`.

The issue with that is that renaming `a` to `c` causes custom URL alias being overridden and incorrectly cleaned up.
```
 id | parent | text |  action   | alias_redirects | is_alias | is_original | link 
----+--------+------+-----------+-----------------+----------+-------------+------
 43 |     42 | b    | eznode:63 |               0 |        0 |           1 |   43
 45 |     44 | b    | eznode:63 |               0 |        1 |           1 |   45
 42 |      0 | c    | eznode:62 |               1 |        0 |           1 |   42
 46 |      0 | a    | eznode:62 |               0 |        0 |           0 |   42
```
Notice that `id=44` referenced in parent column is missing. It's supposed to be cleaned up via `\eZ\Publish\Core\Persistence\Legacy\Content\UrlAlias\Gateway::cleanupAfterPublish` but that neven happens because it's an edge case not considered before.

----

That being said, maybe your change is good, but what we're doing is actually making NOP not reusable, which is behavior change. Going with the fix for cleanup, even if possible, might lead to another issue - I'm not sure how system is going to behave if custom `/c/b` points to completely different Location. Then making it not reusable would make more sense.

Maybe this is something we can discuss with @adamwojs and @ezsystems/php-dev-team before we decide what direction we should take? Or is it clear for everyone and we can move forward with @barw4's fix and just change the wording?

eZ/Publish/API/Repository/Tests/URLAliasServiceTest.php
```diff
@@ -1600,6 +1600,57 @@ public function testDeleteCorruptedUrlAliases()
...
+        $folderUpdateStruct = $contentService->newContentUpdateStruct();
+        $folderUpdateStruct->setField('name', 'C');
+        $renamedFolder = $contentService->updateContent($folderDraft->getVersionInfo(), $folderUpdateStruct);
+        $contentService->publishVersion($renamedFolder->getVersionInfo());
+
+        // Loading aliases shouldn't throw a `BadStateException`
+        $urlAliasService->listLocationAliases($childLocation);
+
+        // Renamed content should have '/c2' path alias
+        $lookupRenamed = $urlAliasService->lookup('C2');
```
alongosz: please also check that `/c/b` lookup works as you'd expect at this point.

eZ/Publish/API/Repository/Tests/URLAliasServiceTest.php
```diff
@@ -1600,6 +1600,57 @@ public function testDeleteCorruptedUrlAliases()
...
+
+        // 4. Rename "A" to "C"
+        $folderDraft = $contentService->createContentDraft($folder->contentInfo);
+        $folderUpdateStruct = $contentService->newContentUpdateStruct();
+        $folderUpdateStruct->setField('name', 'C');
+        $renamedFolder = $contentService->updateContent($folderDraft->getVersionInfo(), $folderUpdateStruct);
+        $contentService->publishVersion($renamedFolder->getVersionInfo());
+
+        // Loading aliases shouldn't throw a `BadStateException`
+        $urlAliasService->listLocationAliases($childLocation);
```
alongosz: can we actually check if the list is as expected? Not sure how big that snippet would be and if we're not gonna run into (Postgre)SQL returned records issues. But still sth to consider.

eZ/Publish/API/Repository/Tests/URLAliasServiceTest.php
barw4: This is obviously a naming mistake, you are correct. I relied too much on that comment above the `if` (https://github.com/ezsystems/ezplatform-kernel/pull/385/files#diff-d640f4389daf07a7bd0a474b4181c7b97c2c3208d172269c69fa59151e1fe014L232) and confused myself. My intention was that an alias would get incremented with `uniqueCounter++` and I therefore treated it as 'reusable', because it gets incremented (reused, right?😺). Sorry about that

barw4: Test improved in https://github.com/ezsystems/ezplatform-kernel/pull/385/commits/bd4088eba026b215debabf3a623d8e64c848828b

alongosz: No problem, URL alias nomenclature  was never easy or straightforward ;)
Now we need to decide if that's the change that we really want. Will discuss this with the Team and probably Adam.

eZ/Publish/API/Repository/Tests/URLAliasServiceTest.php
```diff
@@ -1600,6 +1600,52 @@ public function testDeleteCorruptedUrlAliases()
...
+        $languageCode = 'eng-GB';
+
+        // 1. Create parent folder
+        $folder = $this->createFolder([$languageCode => 'a'], 2);
+        $folderLocationId = $folder->contentInfo->getMainLocationId();
+
+        // 2. Create child folder
+        $child = $this->createFolder([$languageCode => 'b'], $folderLocationId);
+        $childLocationId = $child->contentInfo->getMainLocationId();
+        $childLocation = $locationService->loadLocation($childLocationId);
```
alongosz: Side note: you don't need to load that directly:
```suggestion
        $childLocation = $child->getVersionInfo()->getContentInfo()->getMainLocation()
        $childLocationId = $childLocation->id; 
```
(or use later on `$childLocation->id` directly).

alongosz (commented):
This is ok now, but I need to talk to Adam to make final decision, before giving approval.

alongosz:
@adamwojs @webhdx we need to sync on that, there's behavior change here.

barw4:
Closing in favor of https://github.com/ibexa/core/pull/350 (`v5`)

PR #388, opened 2023-10-12
IBX-6631: Enriched `TrashItem` with `removedLocationContentIdMap`

barw4:
| **JIRA issue**                          | [IBX-6631](https://issues.ibexa.co/browse/IBX-6631)
| **BC breaks**                          |no

Related PR: https://github.com/ezsystems/date-based-publisher/pull/267

One assert was removed because crucial values are compared in the same method which makes this assert obsolete. Obviously, it also fails as loaded `TrashItem` will always have an empty `removedLocationContentIdMap` whether the one built after trashing - won't.

#### Checklist:

### DIFF
--- a/eZ/Publish/API/Repository/Values/Content/TrashItem.php
+++ b/eZ/Publish/API/Repository/Values/Content/TrashItem.php
@@ -19,4 +19,15 @@ abstract class TrashItem extends Location
      * @var \DateTime
      */
     protected $trashed;
+
+    /** @var array<int, int> */
+    protected $removedLocationContentIdMap = [];
+
+    /**
+     * @return array<int, int>
+     */
+    public function getRemovedLocationContentIdMap(): array
+    {
+        return $this->removedLocationContentIdMap;
+    }
 }
--- a/eZ/Publish/Core/Persistence/Legacy/Content/Location/Trash/Handler.php
+++ b/eZ/Publish/Core/Persistence/Legacy/Content/Location/Trash/Handler.php
@@ -74,21 +74,15 @@ class Handler implements BaseTrashHandler
         $this->contentHandler = $contentHandler;
     }
 
-    /**
-     * Loads the data for the trashed location identified by $id.
-     * $id is the same as original location (which has been previously trashed).
-     *
-     * @param int $id
-     *
-     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
-     *
-     * @return \eZ\Publish\SPI\Persistence\Content\Location\Trashed
-     */
-    public function loadTrashItem($id)
+    public function loadTrashItem(int $id, array $trashedLocationsContentMap = []): Trashed
     {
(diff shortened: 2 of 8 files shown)


eZ/Publish/Core/Persistence/Legacy/Content/Location/Trash/Handler.php
```diff
@@ -109,6 +109,7 @@ public function trashSubtree($locationId)
         $locationRows = $this->locationGateway->getSubtreeContent($locationId);
         $isLocationRemoved = false;
         $parentLocationId = null;
+        $trashedLocationsContentMap = [];
```
alongosz: let's be consistent:
```suggestion
        $removedLocationsContentMap = [];
```
(or the other way, other variables starting with trashed instead).

alongosz (commented):
@barw4 this is very good, but as usual, I need some integration test coverage to see that those subtree IDs land there properly in that API TrashItem Value ;)

barw4:
@alongosz test added in https://github.com/ezsystems/ezplatform-kernel/pull/388/commits/48730c11ce932de1ae3d0c6173b299d1c56ce353

alongosz (approved):

konradoboza (approved):

Steveb-p (approved):

ciastektk (approved):

eZ/Publish/API/Repository/Values/Content/TrashItem.php
```diff
@@ -19,4 +19,15 @@ abstract class TrashItem extends Location
...
      */
     protected $trashed;
+
+    /** @var array<int, int> */
+    protected $removedLocationContentIdMap = [];
+
+    /**
+     * @return array<int, int>
+     */
+    public function getRemovedLocationContentIdMap(): array
```
Steveb-p, line 29: In https://github.com/ezsystems/date-based-publisher/pull/267 I see that you are using this map to create content deletion notifications, and perform actual deletions.

Note that Content can have multiple locations, which means the same Content ID might occur in this map multiple times. Is this your intent?

barw4: @Steveb-p Yes, it can appear multiple times. Why would that be a problem? If one schedule entry is deleted, the second one won't be found and hence, won't be deleted. No exception is being thrown.

Steveb-p: If that is the case, then it is ok. I assumed it was a case that was not taken into account and the same action was taken for the same Content ID more than once.

You may consider swapping the map order. By having it be `array<ContentID, array<LocationID>>` (array with `ContentID` as keys and an array of `LocationID`s as value) you would have a similar effect, with Location IDs being grouped together. Only a suggestion though, of course.

tomaszszopinski (approved):

PR #389, opened 2023-10-19
IBX-6504: Gracefully handled URL generation in `RoutingExtension`

barw4:
| **JIRA issue**                          | [IBX-6504](https://issues.ibexa.co/browse/IBX-6504)

Related PR: https://github.com/ezsystems/ezplatform-page-fieldtype/pull/251

#### Checklist:

### DIFF
--- a/eZ/Bundle/EzPublishCoreBundle/Resources/config/services.yml
+++ b/eZ/Bundle/EzPublishCoreBundle/Resources/config/services.yml
@@ -191,7 +191,7 @@ services:
     ezpublish.param_converter.location:
         class: eZ\Bundle\EzPublishCoreBundle\Converter\LocationParamConverter
         arguments:
-            $locationService: "@ezpublish.siteaccessaware.service.location"
+            $locationService: '@ezpublish.siteaccessaware.service.location'
             $contentPreviewHelper: '@ezpublish.content_preview_helper'
         tags:
             - { name: request.param_converter, priority: "%ezpublish.param_converter.location.priority%", converter: ez_location_converter }
--- a/eZ/Bundle/EzPublishCoreBundle/Resources/config/services.yml
+++ b/eZ/Bundle/EzPublishCoreBundle/Resources/config/services.yml
@@ -191,7 +191,8 @@ services:
     ezpublish.param_converter.location:
         class: eZ\Bundle\EzPublishCoreBundle\Converter\LocationParamConverter
         arguments:
-            - "@ezpublish.siteaccessaware.service.location"
+            $locationService: "@ezpublish.siteaccessaware.service.location"
+            $contentPreviewHelper: '@ezpublish.content_preview_helper'
         tags:
             - { name: request.param_converter, priority: "%ezpublish.param_converter.location.priority%", converter: ez_location_converter }
 
--- a/eZ/Publish/Core/MVC/Symfony/Templating/Twig/Extension/RoutingExtension.php
+++ b/eZ/Publish/Core/MVC/Symfony/Templating/Twig/Extension/RoutingExtension.php
@@ -114,15 +114,25 @@ class RoutingExtension extends AbstractExtension
     {
         if ($object instanceof Location) {
             $routeName = UrlAliasRouter::URL_ALIAS_ROUTE_NAME;
+            $forcedLanguageCode = $this->getForcedLanguageCodeBasedOnPreview();
+            if ($forcedLanguageCode !== null) {
+                $parameters += [
+                    'forcedLanguageCode' => $forcedLanguageCode,
+                ];
+            }
             $parameters += [
                 'locationId' => $object->id,
-                'forcedLanguage' => $this->getForcedLanguageCodeBasedOnPreview(),
             ];
         } elseif ($object instanceof Content || $object instanceof ContentInfo) {
             $routeName = UrlAliasRouter::URL_ALIAS_ROUTE_NAME;
+            $forcedLanguageCode = $this->getForcedLanguageCodeBasedOnPreview();
+            if ($forcedLanguageCode !== null) {
+                $parameters += [
+                    'forcedLanguageCode' => $forcedLanguageCode,
+                ];
+            }
             $parameters += [
                 'contentId' => $object->id,
-                'forcedLanguage' => $this->getForcedLanguageCodeBasedOnPreview(),
             ];
         } elseif ($object instanceof RouteReference) {
             $routeName = $object->getRoute();
@@ -142,7 +152,7 @@ class RoutingExtension extends AbstractExtension
      */
     private function getForcedLanguageCodeBasedOnPreview(): ?string
     {
-        if ($this->contentPreviewHelper->isPreviewActive() === false) {
+        if ($this->contentPreviewHelper->isPreviewActive() !== true) {
             return null;
         }
 
--- a/eZ/Publish/Core/MVC/Symfony/Templating/Twig/Extension/RoutingExtension.php
+++ b/eZ/Publish/Core/MVC/Symfony/Templating/Twig/Extension/RoutingExtension.php
@@ -9,46 +9,43 @@ declare(strict_types=1);
 namespace eZ\Publish\Core\MVC\Symfony\Templating\Twig\Extension;
 
 use eZ\Publish\API\Repository\Exceptions\NotFoundException;
-use eZ\Publish\API\Repository\LocationService;
 use eZ\Publish\API\Repository\Values\Content\Content;
 use eZ\Publish\API\Repository\Values\Content\ContentInfo;
 use eZ\Publish\API\Repository\Values\Content\Location;
-use eZ\Publish\Core\Helper\ContentPreviewHelper;
 use eZ\Publish\Core\MVC\Symfony\Routing\Generator\RouteReferenceGeneratorInterface;
 use eZ\Publish\Core\MVC\Symfony\Routing\RouteReference;
 use eZ\Publish\Core\MVC\Symfony\Routing\UrlAliasRouter;
+use Psr\Log\LoggerAwareInterface;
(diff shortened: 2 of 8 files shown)


kisztof (approved):

eZ/Publish/Core/MVC/Symfony/Templating/Twig/Extension/RoutingExtension.php
```diff
@@ -78,14 +88,30 @@ public function getPath(object $name, array $parameters = [], bool $relative = f
...
+            return $this->generateUrlForObject($object, $parameters, $referenceType);
+        } catch (NotFoundException $e) {
+            return '';
+        } catch (Throwable $e) {
+            $this->logger->warning(
+                'Url could not be generated.',
+                ['exception' => $e]
+            );
+
+            return '';
```
Steveb-p: Is it acceptable for us to **always** discard the error? Invalid route usually throws an error in Symfony (and it used to do so for us). This is technically a change in behavior.

Can we look for a different, alternative approach? A parameter, a state?

We could create a Twig token to start and stop a "relaxed" rendering?

```twig
{% relaxed_start %}
    {{ ez_route(...) }}
{% relaxed_end %}
```

barw4: I've discussed this with @alongosz and he wanted to keep it that way. 

Adding relaxed rendering would mean we have to change all the templates utilizing `ez_route` now, right?

alongosz: > Is it acceptable for us to **always** discard the error? Invalid route usually throws an error in Symfony (and it used to do so for us). This is technically a change in behavior.
> 
> Can we look for a different, alternative approach? A parameter, a state?
> 
> We could create a Twig token to start and stop a "relaxed" rendering?
> 
> ```
> {% relaxed_start %}
>     {{ ez_route(...) }}
> {% relaxed_end %}
> ```

Hey @Steveb-p this was my idea. If this should be handled in Twig in a better way, we're of course opened to suggestions. I look at this from API perspective and from API perspective trying to render URL in case of API NotFound Exception doesn't make sense. The URL for main language will be incorrect in that context.

konradoboza: Concerning the above, don't you think that `Url could not be generated` is not informative enough? Wouldn't it be better to try logging the actual exception messages coming directly from `UrlGeneratorInterface::generate` method?

Steveb-p: > Adding relaxed rendering would mean we have to change all the templates utilizing `ez_route` now, right?

Yes, that would be the effect.

> I look at this from API perspective and from API perspective trying to render URL in case of API NotFound Exception doesn't make sense.

I understand why we need it, I am just not happy about how broad the effect of this change will be. We are basically making `ez_path` and family functions somewhat unsafe, different from what they used to be.

I could agree to having this function behave differently if a Content object is not found, or a Content that the Location points to is not found. That makes sense to me. However, swallowing any error is too much. Normal route references should still throw a hard error.

I think our solution should be to only handle the special case in question, and allow any other errors and exceptions to propagate normally. They are potentially a logic error and should not be hidden away.

alongosz: > I could agree to having this function behave differently if a Content object is not found, or a Content that the Location points to is not found. That makes sense to me. However, swallowing any error is too much. Normal route references should still throw a hard error.
> 
> I think our solution should be to only handle the special case in question, and allow any other errors and exceptions to propagate normally. They are potentially a logic error and should not be hidden away.

@Steveb-p so, if we handle just `NotFound` and leave out logging + handling Throwable, would that be okay for now?

Steveb-p: I believe that would be the best approach. We can handle our own errors, but we shouldn't touch the ones generated from normal Symfony routes, in my opinion.

alongosz: > I believe that would be the best approach. We can handle our own errors, but we shouldn't touch the ones generated from normal Symfony routes, in my opinion.

@barw4 ok, let's cut the last catch and logger changes as you initially suggested yourself. You were right :grinning:

barw4: @alongosz @Steveb-p resolved in  https://github.com/ezsystems/ezplatform-kernel/pull/389/commits/791b764e227720160ca4d6309a28c8a4b0018576

Steveb-p (approved):

konradoboza (approved):

alongosz (approved):

micszo:
Fix works when LP preview is empty (no block or no base lang). Otherwise exception persists but template changes (ezplatform-kernel/eZ/Bundle/EzPublishCoreBundle/Resources/views/pagelayout.html.twig -> ezplatform-page-fieldtype/src/bundle/Resources/views/default_layout.html.twig).

barw4:
@alongosz @konradoboza @Steveb-p @kisztof as @micszo noticed with the certain case the preview still wasn't able to be loaded. This is due to block rendering in a new language and loading old values if we are basing a new translation on an old translation. https://github.com/ezsystems/ezplatform-page-fieldtype/pull/251 + https://github.com/ezsystems/ezplatform-kernel/pull/389/commits/124723a0dd881b0ab0d983f132674124f17cf6e7 should handle this case

micszo:
This unblocked the mentioned flow. 🙂

alongosz:
@barw4 so if page is translated into many languages that would hit performance badly as all 30 translations would be loaded. I'm not sure if there's API issue here or we use it incorrectly, but that doesn't feel like a good direction performance wise. Not sure what to suggest without looking deeper into this. What is exactly the error there and what and where we're trying to load that we shouldn't?

cc @ezsystems/php-dev-team

barw4:
@alongosz we have multiple calls like this (with `Language::ALL`) throughout the whole codebase so I'm a bit surprised this has become a problem of performance just now. Especially, if this will only in reality affect PB's previews. In the same request, we are having the exact same calls from other packages as well (for example https://github.com/ezsystems/ezplatform-page-fieldtype/blob/2.3/src/lib/FieldType/Page/ParameterProvider.php#L116).

To answer your question, this is basically the same case and comes down to loading Location, but now this is caused by blocks rendering https://github.com/ezsystems/ezplatform-page-fieldtype/blob/2.3/src/bundle/Resources/views/default_layout.html.twig#L8, specifically https://github.com/ezsystems/ezplatform-page-fieldtype/blob/2.3/src/lib/FragmentRenderer/BlockRenderOptionsFragmentRenderer.php#L121 and later https://github.com/ezsystems/ezplatform-page-fieldtype/blob/2.3/src/lib/FragmentRenderer/BlockRenderOptionsFragmentRenderer.php#L148. Situation is pretty much similar as `ez_path` - we cannot load Location in given Site Access when `alwaysAvailable` flag is `false` without an error.

eZ/Bundle/EzPublishCoreBundle/Resources/config/services.yml
```diff
@@ -191,7 +191,8 @@ services:
     ezpublish.param_converter.location:
         class: eZ\Bundle\EzPublishCoreBundle\Converter\LocationParamConverter
         arguments:
-            - "@ezpublish.siteaccessaware.service.location"
+            $locationService: "@ezpublish.siteaccessaware.service.location"
+            $contentPreviewHelper: '@ezpublish.content_preview_helper'
```
alongosz: CS:
```suggestion
            $locationService: '@ezpublish.siteaccessaware.service.location'
            $contentPreviewHelper: '@ezpublish.content_preview_helper'
```

alongosz (approved):
> @alongosz we have multiple calls like this (with `Language::ALL`) throughout the whole codebase so I'm a bit surprised this has become a problem of performance just now. Especially, if this will only in reality affect PB's previews. In the same request, we are having the exact same calls from other packages as well (for example https://github.com/ezsystems/ezplatform-page-fieldtype/blob/2.3/src/lib/FieldType/Page/ParameterProvider.php#L116).
> 
> To answer your question, this is basically the same case and comes down to loading Location, but now this is caused by blocks rendering https://github.com/ezsystems/ezplatform-page-fieldtype/blob/2.3/src/bundle/Resources/views/default_layout.html.twig#L8, specifically https://github.com/ezsystems/ezplatform-page-fieldtype/blob/2.3/src/lib/FragmentRenderer/BlockRenderOptionsFragmentRenderer.php#L121 and later https://github.com/ezsystems/ezplatform-page-fieldtype/blob/2.3/src/lib/FragmentRenderer/BlockRenderOptionsFragmentRenderer.php#L148. Situation is pretty much similar as `ez_path` - we cannot load Location in given Site Access when `alwaysAvailable` flag is `false` without an error.

I'm not fully convinced by "we're already doing it like that" argument, however maybe it's fine for now if @micszo tries to benchmark it a bit with e.g. content item with 30 translations? AFAIR test-fixtures already contain a lot of languages, creating such content might be a bit time consuming, so maybe some sort of command to loop over languages and create translations? ;)

konradoboza (approved):

micszo:
> however maybe it's fine for now if @micszo tries to benchmark it a bit with e.g. content item with 30 translations? 

Tested the flow by adding 30 translations to a Landing page with simple non-empty preview. Did not observe any significant increase in time when editing, previewing or publishing.

micszo (approved):
QA Approved on Ibexa Commerce 3.3.36-dev.

PR #390, opened 2023-10-26
IBX-6833: Copying empty fields during translation copying shouldn't be skipped

barw4:
| :ticket: Issue |  [IBX-6833](https://issues.ibexa.co/browse/IBX-6833) |
|----------------|-----------|

#### Description:

When one is copying translations using the `copyTranslationsFromPublishedVersion` method during publishing empty values are left out and not stored in the database which could result in entries not appearing in `ezcontentobject_attribute` table resulting in missing translations and broken language masks.

#### For QA:

_Maintainer note:_

Requires extensive multilingual content updating testing, ideally in multiple tabs or even by at least 2 separate editors at the same time (though for that unrelated bugs might occur). Possible regressions: [EZEE-2661](https://issues.ibexa.co/browse/EZEE-2661) but doesn't really require scheduler interaction, rather checking if translations were copied properly.

### DIFF
--- a/eZ/Publish/Core/Repository/ContentService.php
+++ b/eZ/Publish/Core/Repository/ContentService.php
@@ -1294,7 +1294,8 @@ class ContentService implements ContentServiceInterface
     protected function internalUpdateContent(
         APIVersionInfo $versionInfo,
         APIContentUpdateStruct $contentUpdateStruct,
-        ?array $fieldIdentifiersToValidate = null
+        ?array $fieldIdentifiersToValidate = null,
+        bool $doAddEmpty = false
     ): Content {
         $contentUpdateStruct = clone $contentUpdateStruct;
 
@@ -1385,7 +1386,7 @@ class ContentService implements ContentServiceInterface
                 );
                 $fieldValues[$fieldDefinition->identifier][$languageCode] = $fieldValue;
 
-                if ($isRetained || $isCopied || ($isLanguageNew && $isEmpty) || $isProcessed) {
+                if ($isRetained || $isCopied || ($isLanguageNew && $isEmpty && !$doAddEmpty) || $isProcessed) {
                     continue;
                 }
 
@@ -1692,7 +1693,7 @@ class ContentService implements ContentServiceInterface
             $updateStruct->setField($fallbackField->fieldDefIdentifier, $fallbackField->value, $fallbackField->languageCode);
         }
 
-        $this->internalUpdateContent($versionInfo, $updateStruct);
+        $this->internalUpdateContent($versionInfo, $updateStruct, null, true);
     }
 
     protected function fieldValuesAreEqual(FieldType $fieldType, Value $value1, Value $value2): bool
--- a/eZ/Publish/Core/Repository/ContentService.php
+++ b/eZ/Publish/Core/Repository/ContentService.php
@@ -1295,7 +1295,7 @@ class ContentService implements ContentServiceInterface
         APIVersionInfo $versionInfo,
         APIContentUpdateStruct $contentUpdateStruct,
         ?array $fieldIdentifiersToValidate = null,
-        bool $doAddEmpty = false
+        bool $copyEmptyField = false
     ): Content {
         $contentUpdateStruct = clone $contentUpdateStruct;
 
@@ -1386,7 +1386,7 @@ class ContentService implements ContentServiceInterface
                 );
                 $fieldValues[$fieldDefinition->identifier][$languageCode] = $fieldValue;
 
-                if ($isRetained || $isCopied || ($isLanguageNew && $isEmpty && !$doAddEmpty) || $isProcessed) {
+                if ($isRetained || $isCopied || ($isLanguageNew && $isEmpty && !$copyEmptyField) || $isProcessed) {
                     continue;
                 }
 
--- a/tests/integration/Core/Repository/ContentService/CopyTranslationsFromPublishedVersionTest.php
+++ b/tests/integration/Core/Repository/ContentService/CopyTranslationsFromPublishedVersionTest.php
@@ -1,0 +1,115 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace Ibexa\Tests\Integration\Core\Repository\ContentService;
+
+use DateTime;
+use eZ\Publish\API\Repository\Values\ContentType\FieldDefinitionCreateStruct;
+use eZ\Publish\Core\Repository\Values\Content\ContentUpdateStruct;
+use Ibexa\Tests\Integration\Core\RepositoryTestCase;
+
+/**
+ * @covers \eZ\Publish\API\Repository\ContentService
+ */
+final class CopyTranslationsFromPublishedVersionTest extends RepositoryTestCase
+{
+    private const ENG_LANGUAGE_CODE = 'eng-GB';
+    private const GER_LANGUAGE_CODE = 'ger-DE';
+    private const US_LANGUAGE_CODE = 'eng-US';
+    private const CONTENT_TYPE_IDENTIFIER = 'custom';
+
+    /**
+     * @throws \eZ\Publish\API\Repository\Exceptions\Exception
+     */
+    public function testCopyTranslationsFromPublishedVersionCopiesEmptyValues(): void
+    {
+        $this->createContentType();
+
+        $contentService = self::getContentService();
+        $contentTypeService = self::getContentTypeService();
+        $locationService = self::getLocationService();
+
+        // Creating and publishing content in eng-GB language
+        $contentType = $contentTypeService->loadContentTypeByIdentifier(self::CONTENT_TYPE_IDENTIFIER);
+        $mainLanguageCode = self::ENG_LANGUAGE_CODE;
+        $contentCreateStruct = $contentService->newContentCreateStruct($contentType, $mainLanguageCode);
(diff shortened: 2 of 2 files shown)


kisztof (approved):

ViniTou (commented):
I can not wrap my head around it, why storing empty values is somehow breaking our system? How it is handled when you just creating version with empty values?

Also, which of those is not enough so the `isCopied` flag is not set to true and you need to pass additional argument?

```php
                } elseif (!$isFieldUpdated && $isLanguageNew && !$fieldDefinition->isTranslatable) {
                    $isCopied = true;
                }
 ```

**Comment by barw4** [comment]
> I can not wrap my head around it, why storing empty values is somehow breaking our system? 

@ViniTou The problem is that they are not stored.

> Also, which of those is not enough so the isCopied flag is not set to true and you need to pass additional argument?

The `isCopied` flag is irrelevant here. Our system fails when `($isLanguageNew && $isEmpty) === true` and that's what happens if we are copying translations from published version using the `copyTranslationsFromPublishedVersion ` method. In such case, empty value is skipped and therefore an empty field that ought to be translated is not injected into a db.

**Comment by ViniTou** [comment]
Ok, then in which case (outside of copying translation) `$isLanguageNew && $isEmpty` is happening and storing is skipped? Like it seems that there is/was valid case for skipping empty values without breaking the system.

>The isCopied flag is irrelevant here.

I know, it was more about, that maybe we can set it using some other logic to not skip adding without introducing new argument to method signature.

**Comment by barw4** [comment]
> Ok, then in which case (outside of copying translation) $isLanguageNew && $isEmpty is happening and storing is skipped? Like it seems that there is/was valid case for skipping empty values without breaking the system.

For example when we are creating a new version in a new language and we are not basing it on another translation so that values are empty.

> I know, it was more about, that maybe we can set it using some other logic to not skip adding without introducing new argument to method signature.

The `isCopied` flag is mostly used for non-translatable fields so I cannot see how would it help in this case:
```
 elseif (!$isFieldUpdated && $isLanguageNew && !$fieldDefinition->isTranslatable) {
                    $isCopied = true;
                }
```

File: tests/integration/Core/Repository/ContentService/CopyTranslationsFromPublishedVersionTest.php [hunk 1]
```diff
@@ -0,0 +1,115 @@
...
+        ]);
+        $contentUpdateStruct->setField('title', null);
+        $usContent = $contentService->updateContent($usDraft->getVersionInfo(), $contentUpdateStruct);
+        $publishedUsContent = $contentService->publishVersion($usContent->getVersionInfo(), [self::US_LANGUAGE_CODE]);
+
+        $gerFieldValueInUsContent = $publishedUsContent->getField('title', self::GER_LANGUAGE_CODE)
+            ->getValue()
+            ->text;
+
+        self::assertSame('', $gerFieldValueInUsContent);
```
alongosz: 1. `->text` is a polymorphic call and this will most likely make PHPStan fail in ibexa/core without extra assertions.
2. Let's add sanity test for eng-US field value as well.

1 & 2:
```suggestion
        $usFieldValueInUsContent = $publishedUsContent->getField('title', self::US_LANGUAGE_CODE)->getValue();
        self::assertInstanceOf(TextLine\Value::class, $usFieldValueInUsContent);
        self::assertSame('', $usFieldValueInUsContent->text);

        $gerFieldValueInUsContent = $publishedUsContent->getField('title', self::GER_LANGUAGE_CODE)->getValue();
        self::assertInstanceOf(TextLine\Value::class, $gerFieldValueInUsContent);
        self::assertSame('', $gerFieldValueInUsContent->text);
```

eZ/Publish/Core/Repository/ContentService.php
```diff
@@ -1294,7 +1294,8 @@ public function updateContent(APIVersionInfo $versionInfo, APIContentUpdateStruc
     protected function internalUpdateContent(
         APIVersionInfo $versionInfo,
         APIContentUpdateStruct $contentUpdateStruct,
-        ?array $fieldIdentifiersToValidate = null
+        ?array $fieldIdentifiersToValidate = null,
+        bool $doAddEmpty = false
```
alongosz: Naming: how about `$copyEmptyField`?

However, what are the side effects of always copying empty field instead of just when copying translations?
Or maybe we could properly prepare update struct prior calling internalUpdateContent?

The pre-existing code is just very complex here and hard to understand (not your fault :-) ).

What makes me wonder though is how emptiness is related to the issue at hand. IBX-6833 descriptions states that in general, a translation is lost. In such case does this change solve the issue if it's applied to empty fields only? If we test this against more direct case mentioned in the ticket
> Create a new content of the newly created Content Type with random values at two mentioned fields.

will that sill work somehow?

barw4: @alongosz I'll try to explain step by step what's going on behind the scenes.

After publishing version 2 having `eng-GB` and `ger-DE` translations we are at the following scenario:

 version | data_text | language_code
| ------------- | ------------- | -------------- |
| 1 |	one	| eng-GB |
| 1 |	two	| eng-GB |
| 2 |	one	| eng-GB |
| 2 |	two	| eng-GB |
| 2 |		| ger-DE |
| 2 |	one	| ger-DE |
| 3 |	one	| eng-GB |
| 3 |	two	| eng-GB |
| 3 |		| fre-FR |
| 3 |	one	| fre-FR |

Please note, that the 3rd version has only `eng-GB` and `fre-FR` translations because the 3rd version was created before we published the 2nd version.

Now, we publish the 3rd version with `eng-GB` and `fre-FR` languages.

 version | data_text | language_code
| ------------- | ------------- | -------------- |
| 1 |	one	| eng-GB |
| 1 |	two	| eng-GB |
| 2 |	one	| eng-GB |
| 2 |	two	| eng-GB |
| 2 |		| ger-DE |
| 2 |	one	| ger-DE |
| 3 |	one	| eng-GB |
| 3 |	two	| eng-GB |
| 3 |		| fre-FR |
| 3 |	one	| fre-FR |

Records are the same in both tables. Now, because the `ger-DE` translation is missing the `language_mask` in the `ezcontentobject` table is wrong because it contains only `eng-GB` and `fre-FR` translations, therefore, one - the `ger-DE` translation goes missing.

That's basically the issue from the ticket description.

Now, what we actually need to do to overcome this issue is somehow influence this `if`:
```php
if ($isRetained || $isCopied || ($isLanguageNew && $isEmpty && !$doAddEmpty) || $isProcessed) {
    continue;
}
```
and make the following conditional `false`: `($isLanguageNew && $isEmpty && !$doAddEmpty)` in case of empty fields because we would want those fields to be created but **only when we are copying translations**. 

In the case of our scenario the `$isLanguageNew` variable is `true` and `isEmpty` is `true` as well, when publishing the 3rd version. In such a case the loop finishes at the `continue` preventing empty fields from creation.

Now to answer your questions:
> What makes me wonder though is how emptiness is related to the issue at hand. IBX-6833 descriptions states that in general, a translation is lost. In such case does this change solve the issue if it's applied to empty fields only? If we test this against more direct case mentioned in the ticket

The emptiness is the main issue here because the loop finishes at the `continue`. Currently in this PR **we are in fact applying this to empty fields only** because of `($isLanguageNew && $isEmpty && !$doAddEmpty)`.

> If we test this against more direct case mentioned in the ticket
Create a new content of the newly created Content Type with random values at two mentioned fields.
will that sill work somehow?

It will work properly if a translatable field is not empty. Only empty fields are causing issues.

> However, what are the side effects of always copying an empty field instead of just when copying translations?

I'm not sure if I understood you correctly but this won't make it work in case of always copying an empty field. The 3rd version does not have any information about the 2nd version, it's starting from 1st version therefore only the `copyTranslationsFromPublishedVersion` will allow us to get a necessary data from an earlier, published version.

It's kind of a complicated topic, if you wish we can sync regarding this.

eZ/Publish/Core/Repository/ContentService.php
```diff
@@ -1385,7 +1386,7 @@ protected function internalUpdateContent(
                 );
                 $fieldValues[$fieldDefinition->identifier][$languageCode] = $fieldValue;
 
-                if ($isRetained || $isCopied || ($isLanguageNew && $isEmpty) || $isProcessed) {
+                if ($isRetained || $isCopied || ($isLanguageNew && $isEmpty && !$copyEmptyField) || $isProcessed) {
```
Nattfarinn, line 1389:
```suggestion
                if ($isRetained || $isCopied || $isLanguageNew || $isProcessed) {
```

`internalUpdateContent` is a method used only it two places and, correct me if I am wrong, the change does not seem to affect `ContentService::contentUpdate` much/at all. The parameter isn't exposed beyond protected scope so it basically is a hardcoded behaviour.

barw4: @Nattfarinn no, this won't make it. You can immediately see that integration tests fail one by one if we apply this change. 

We have to skip this `if` logic only when we are copying translations from already published version.

webhdx (approved):
It's quite a drastic change and we definitely need a solid QA round here to detect possible side effects. I also feel like fields were skipped for a reason but adding them anyway seems like a reasonable solution to keep data integrity.

alongosz (approved):
After digging into this deeper, I don't see any apparent issues with this change, though there might be some unforeseen side-effects. Indeed we need to vary that internal method based on either content update or copying fields during publishing, which reuses that method.

Probably more refactoring would be better suited now that we know these operations are actually not the same, but this - what @barw4 did now - seems like the simplest way to achieve it.

I've updated PR description with some extra tips for QA to investigate.

tomaszszopinski (approved):
QA approved on Ibexa DXP oss 3.3.x-dev

PR #391, opened 2023-11-07
IBX-6592: Content Object State assignment should rely on `Location` instead of `ContentInfo`

barw4:
| **JIRA issue**                          | [IBX-6592](https://issues.ibexa.co/browse/IBX-6592)

Currently, Content Object State assignment is validated, as permissions go, utilizing `ContentInfo` which in my opinion is wrong, as that leaves Subtree Limitation pretty much useless, and yet, this limitation is available for a choosing when we are dealing with the `state/assign` policy. 

To keep BC I've provided a new argument that is `?Location = null`. This `Location` would be later checked against Subtree Limitation when assigning an Object State to a certain Content.

Related PR: https://github.com/ezsystems/ezplatform-admin-ui/pull/2112

#### Checklist:

### DIFF
--- a/eZ/Publish/API/Repository/ObjectStateService.php
+++ b/eZ/Publish/API/Repository/ObjectStateService.php
@@ -7,6 +7,7 @@
 namespace eZ\Publish\API\Repository;
 
 use eZ\Publish\API\Repository\Values\Content\ContentInfo;
+use eZ\Publish\API\Repository\Values\Content\Location;
 use eZ\Publish\API\Repository\Values\ObjectState\ObjectState;
 use eZ\Publish\API\Repository\Values\ObjectState\ObjectStateCreateStruct;
 use eZ\Publish\API\Repository\Values\ObjectState\ObjectStateGroup;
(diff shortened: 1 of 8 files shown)


ViniTou:
Correct me if I get this wrong.
Now (maybe it also works like this right now, don't know) it will be possible to set global object state just with access to specific subtree? Like what if content is in more then one location?

Nattfarinn (changes requested):
Changes to `ContentService` interface are not applied to Event layer. Please update `setContentState` method implementation and add new field to `BeforeSetContentStateEvent` and `SetContentStateEvent` events.

eZ/Publish/API/Repository/ObjectStateService.php
```diff
@@ -172,12 +173,13 @@ public function deleteObjectState(ObjectState $objectState): void;
...
-     * @param \eZ\Publish\API\Repository\Values\Content\ContentInfo $contentInfo
-     * @param \eZ\Publish\API\Repository\Values\ObjectState\ObjectStateGroup $objectStateGroup
-     * @param \eZ\Publish\API\Repository\Values\ObjectState\ObjectState $objectState
      */
-    public function setContentState(ContentInfo $contentInfo, ObjectStateGroup $objectStateGroup, ObjectState $objectState): void;
+    public function setContentState(
+        ContentInfo $contentInfo,
+        ObjectStateGroup $objectStateGroup,
+        ObjectState $objectState,
+        ?Location $location = null
```
Nattfarinn, line 181: Is it necessary? This Location parameter does not affect method logic and is just workaround for `PermissionResolver`. It is also security risk, as **not** setting this parameter will omit some of limitations checks.

Nattfarinn (changes requested):

Nattfarinn:
Related comment: https://github.com/ezsystems/ezplatform-admin-ui/pull/2112#issuecomment-1835737817

PR #394, opened 2023-11-14
IBX-6494: Fixed copying of non-translatable fields to later versions

barw4:
| **JIRA issue**                          | [IBX-6494](https://issues.ibexa.co/browse/IBX-6494)

The issue was present before the `copyNonTranslatableFieldsFromPublishedVersion` was introduced. However, this method would allow us to handle also the case given in the public ticket.

_Maintainer update_

#### QA

This fix makes non-translatable fields to be copied regardless of if they're coming from higher or lower version number, so we need some extensive testing both for scenario in the ticket and some other ideas of how Editor could manipulate content and possibly break it. The version number simple math comparison  (`>`) was there _probably_ since eZ Publish. However I think this was never a good determinant of if something should be overridden.

_Side note: would be nice to have in such case a version comparison response which would result in an UI dialog stating that the field changed in some other version, prompting user to choose proper value. But we're nowhere near that yet._

#### Checklist:

### DIFF
--- a/eZ/Publish/Core/Repository/ContentService.php
+++ b/eZ/Publish/Core/Repository/ContentService.php
@@ -1517,10 +1517,7 @@ class ContentService implements ContentServiceInterface
         $publishedContent = $this->internalLoadContentById($versionInfo->getContentInfo()->getId());
         $publishedVersionInfo = $publishedContent->getVersionInfo();
 
-        if (
-            !$publishedVersionInfo->isPublished()
-            || ($versionInfo->versionNo >= $publishedVersionInfo->versionNo)
-        ) {
+        if (!$publishedVersionInfo->isPublished()) {
             return;
         }
 
--- a/eZ/Publish/Core/Repository/ContentService.php
+++ b/eZ/Publish/Core/Repository/ContentService.php
@@ -1521,9 +1521,8 @@ class ContentService implements ContentServiceInterface
             return;
         }
 
-        $publishedContentFieldsInMainLanguage = $publishedContent->getFieldsByLanguage(
-            $publishedContent->getVersionInfo()->getContentInfo()->getMainLanguageCode()
-        );
+        $mainLanguageCode = $publishedContent->getVersionInfo()->getContentInfo()->getMainLanguageCode();
+        $publishedContentFieldsInMainLanguage = $publishedContent->getFieldsByLanguage($mainLanguageCode);
 
         $fieldValues = [];
         $persistenceFields = [];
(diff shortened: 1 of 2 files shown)


eZ/Publish/Core/Repository/ContentService.php
```diff
@@ -1517,10 +1517,7 @@ protected function copyNonTranslatableFieldsFromPublishedVersion(APIContent $cur
         $publishedContent = $this->internalLoadContentById($versionInfo->getContentInfo()->getId());
         $publishedVersionInfo = $publishedContent->getVersionInfo();
 
-        if (
-            !$publishedVersionInfo->isPublished()
```
ViniTou, line 1521: Isnt that going to break some cases when things are not copied for translations (\Ibexa\Core\Repository\ContentService::copyTranslationsFromPublishedVersion) and then some other fields will be copied here?

Or isnt it going to break publishing older version, with overriding all fields with current ones?

barw4: @ViniTou 
> Isnt that going to break some cases when things are not copied for translations (\Ibexa\Core\Repository\ContentService::copyTranslationsFromPublishedVersion) and then some other fields will be copied here?

Hmm, I cannot see why removing this condition would break that. The updated method only deals with nontranslatable fields and those are not handled in `copyTranslationsFromPublishedVersion`.

> Or isnt it going to break publishing older version, with overriding all fields with current ones?

I've just tested this and it seems that version 2 (3 is published) properly overridden version 3 so values from version 3 weren't copied into earlier version 2 during publishing if that's what you had in mind.

ViniTou: >The updated method only deals with nontranslatable fields and those are not handled in copyTranslationsFromPublishedVersion.

Yes, my point was, that right now some fields could be copied/not copied there, and be copied/not copied here as the logic now differs.

But I guess, maybe QA will stumble on specific case as I cant come up with proper testing scenario for it.

ViniTou (approved):

kisztof (approved):

eZ/Publish/Core/Repository/ContentService.php
```diff
@@ -1517,10 +1517,7 @@ protected function copyNonTranslatableFieldsFromPublishedVersion(APIContent $cur
...
-            || ($versionInfo->versionNo >= $publishedVersionInfo->versionNo)
```
alongosz, line 1522: AFAIR this logic was there since eZ Publish, so we indeed need extensive QA

alongosz (approved):
@barw4 could you rebase this PR so we can see the current state of CI.

barw4:
> @barw4 could you rebase this PR so we can see the current state of CI.

Done, will send to QA after it's green

konradoboza (approved):

Nattfarinn (approved):

tomaszszopinski:
Note from QA: Ive observed a behaviour that might be a bug; 
**Steps to reproduce:**
1. Create foo content item (from JIRA)
2. Add translation to this content item (.ie ger-DE)
3. Edit eng-GB translation (change image) 
4. Publish
5. Image in ger-DE translation is not changing (it changes only after editing & publishing)

Without this PR this behaviour is not happening.

https://github.com/ezsystems/ezplatform-kernel/assets/59650405/fb95f1a4-fa9d-4865-913d-c6d9ebf8b659

barw4:
@alongosz @ViniTou @konradoboza @Nattfarinn calling for a second review.

As @tomaszszopinski noticed the first iteration of this PR fixed the issue from the ticket but broke the original idea for this method (older versions with different languages were not updated properly with non-translatable fields). It's really hard to explain but the newest change will allow changing a value of a non-translatable field with an older published field's value only if a current version is older than a published one. Test was added that validates this update.

alongosz (approved):

konradoboza (approved):

tomaszszopinski (approved):

PR #397, opened 2023-12-14
IBX-5388: Fixed performance issues of content updates after field changes

Nattfarinn:
| **JIRA issue**                          | [IBX-5388](https://issues.ibexa.co/browse/IBX-5388)
| **BC breaks**                          | no (or soft)

When new field was added we iterated over each content and applied field changes with default values to each existing version. When we removed the field we iterated over each content and removed data from every version as well. This cause major performance issue as all these changes happened in runtime each time you tried to publish Content Type and grow exponentially with increasing number of content items leading to memory issues, transaction nesting issues and exceeded execution times.

This PR introduces two keystones: 
 - on field removed: there is no need to update content items, we can just ignore additional data within persistence layer during content load,
 - on field added: there is no need to update content items with default field values, as we can add these fields in runtime to object without touching persistence (like "virtual" fields).

This makes publishing Content Type process much faster and doesn't make content loading process slower (loading times differences are unnoticeable). Change itself is almost transparent to the user and developer and these virtual fields does not need any special handling for day-to-day operations. Once the content is published, copied or anyhow saved, virtual fields become persisted and will be indistinguishable from any other field. But there are some behavior changes, but with low impact:
 - virtual fields are not stored: this means fields do not have an `id` (it remains `null`),
 - virtual fields are dynamic:  this means default value is not static anymore. Changes to default values within Content Type field definitions will be immediately reflected by all content items that still rely on virtual fields.

#### Checklist:


eZ/Publish/API/Repository/Tests/NonRedundantFieldSetTest.php
```diff
@@ -152,13 +152,13 @@ public function testCreateContentEmptyValuesTranslationNotStored()
      *
      * @param \eZ\Publish\API\Repository\Values\Content\Content $content
      */
-    public function testCreateContentEmptyValuesTranslationNotStoredFields(Content $content)
+    public function testCreateContentEmptyValuesTranslationVirtualFields(Content $content)
```
alongosz:
```suggestion
    public function testCreateContentEmptyValuesTranslationVirtualFields(Content $content): void
```

eZ/Publish/API/Repository/Tests/NonRedundantFieldSetTest.php
```diff
@@ -262,22 +266,26 @@ public function testCreateContentTwoLanguagesSecondTranslationNotStored()
      *
      * @param \eZ\Publish\API\Repository\Values\Content\Content $content
      */
-    public function testCreateContentTwoLanguagesSecondTranslationNotStoredFields(Content $content)
+    public function testCreateContentTwoLanguagesSecondTranslationVirtualFields(Content $content)
```
alongosz:
```suggestion
    public function testCreateContentTwoLanguagesSecondTranslationVirtualFields(Content $content): void
```

eZ/Publish/Core/Persistence/Legacy/Content/FieldHandler.php
```diff
@@ -151,6 +151,10 @@ protected function getEmptyField(FieldDefinition $fieldDefinition, $languageCode
     public function createExistingFieldsInNewVersion(Content $content)
```
alongosz: Nice to have, since you've touched the method: 
```suggestion
    public function createExistingFieldsInNewVersion(Content $content): void
```

eZ/Publish/Core/Persistence/Legacy/Content/Mapper.php
```diff
@@ -39,16 +41,25 @@ class Mapper
...
+     * @var \eZ\Publish\SPI\Persistence\Content\Type\Handler
+     */
+    private $contentTypeHandler;
+
     /**
      * Creates a new mapper.
      *
      * @param \eZ\Publish\Core\Persistence\Legacy\Content\FieldValue\ConverterRegistry $converterRegistry
      * @param \eZ\Publish\SPI\Persistence\Content\Language\Handler $languageHandler
      */
```
alongosz: Nice to have: you can drop this entire block as it provides no value.
```suggestion
```

eZ/Publish/Core/Persistence/Legacy/Tests/Content/MapperTest.php
```diff
@@ -251,6 +323,30 @@ public function testExtractContentFromRowsMultipleVersions()
...
+        int $expectedConverterCalls = null
+    ): Registry {
+        $convMock = $this->createMock(Converter::class);
+        $convMock->expects(
+            $expectedConverterCalls === null
+            ? $this->any()
+            : $this->exactly($expectedConverterCalls)
+        )
+            ->method('toFieldValue')
+            ->will($this->returnValue(new FieldValue()));
```
Steveb-p: This should be sufficient, as it basically does the same thing.
```suggestion
            ->willReturn(new FieldValue());
```

eZ/Publish/Core/Persistence/Legacy/Content/Mapper.php
```diff
@@ -182,33 +193,50 @@ public function convertToStorageValue(Field $field)
...
             if (!isset($contentInfos[$contentId])) {
                 $contentInfos[$contentId] = $this->extractContentInfoFromRow($row, $prefix);
             }
             if (!isset($versionInfos[$contentId])) {
                 $versionInfos[$contentId] = [];
             }
 
-            $versionId = (int)$row['ezcontentobject_version_id'];
+            if (!isset($fieldDefinitions[$contentId][$versionId])) {
+                $contentType = $this->contentTypeHandler->load($contentTypeId);
```
alongosz: I'm a bit worried about content type loading per each loop iteration. In most cases it will be the same contentTypeId for multiple rows (regardless if we load a single content item or a content item list), because of cross product on ezcontent: object, version, and attribute.

I know that we have in-memory cache on top of persistence cache, but I also remember that there were some issues there and in some cases that cache ended up disabled. If you checked that this is indeed in-memory cached, then maybe it's unnecessary - but I would make sure that this self-contained logic doesn't do multiple fetches. We also want to avoid too many queries to persistence cache because it tends to be problematic as well for redis.

eZ/Publish/Core/Persistence/Legacy/Content/Mapper.php
```diff
@@ -227,6 +255,17 @@ public function extractContentFromRows(array $rows, array $nameRows, $prefix = '
                 $content->versionInfo->names = $names;
                 $content->versionInfo->contentInfo = $contentInfo;
                 $content->fields = array_values($fields[$contentId][$versionId]);
+
+                /** @var string $languageCode */
```
alongosz: Is this PHPDoc necessary? It seems that the type of `$languageCode` is easily inferred by static analysis

eZ/Publish/Core/Persistence/Legacy/Content/Mapper.php
```diff
@@ -182,33 +193,50 @@ public function convertToStorageValue(Field $field)
     public function extractContentFromRows(array $rows, array $nameRows, $prefix = 'ezcontentobject_')
```
alongosz: I see we have missing reporting of `@throws` here and while touching the method we could also improve its signature:
```php
     *
     */
    public function extractContentFromRows(array $rows, array $nameRows, string $prefix = 'ezcontentobject_'): array
```

alongosz: With these changes to the method, its cognitive complexity grew out of proportion. I'd suggest refactoring it so it's more readable, splitting logically-encapsulated parts into separate methods.

What I potentially see could be refactored is:
1. **L223-230 to the form of**:
     ```php
      if (!isset($fieldDefinitions[$contentId][$versionId])) {
          $fieldDefinitions[$contentId][$versionId] = $this->loadCachedFieldDefinitionsPerLanguage($languageCode, $contentTypeId);
      }
     ```
     which might include earlier in-memory caching remark (again, unless you're certain it doesn't do to many non-in-memory calls).

     **Alternatively** I would move that loading prior **L212** - processing `foreach($rows)` - however that would  require some preliminary iteration over rows, which has its own flaws (e.g. we can't make it a generator).

2. **L245-271** could hopefully be easily refactored into a separate method, given it operates on the data that were build prior processing and there's nothing further after that block.
    2a) L250-252 can be refactored into:
    ```php
                // Fallback to just the main language name if versioned name data is missing
                $names = $versionedNameData[$contentId][$versionInfo->versionNo] ?? [
                    $contentInfo->mainLanguageCode => $contentInfo->name,
                ];
                
    ```


Any extra or completely different ideas for making this method more compact are welcomed.

eZ/Publish/Core/Persistence/Legacy/Content/StorageHandler.php
```diff
@@ -79,7 +79,7 @@ public function copyFieldData(VersionInfo $versionInfo, Field $field, Field $ori
     public function getFieldData(VersionInfo $versionInfo, Field $field)
     {
         $storage = $this->storageRegistry->getStorage($field->type);
-        if ($storage->hasFieldData()) {
+        if ($storage->hasFieldData() && $field->id !== null) {
```
alongosz: > [EA] This condition execution costs less than the previous one. 
```suggestion
        if ($field->id !== null && $storage->hasFieldData()) {
```

eZ/Publish/Core/Persistence/Legacy/Content/Type/Update/Handler/DoctrineDatabase.php
```diff
@@ -23,21 +22,14 @@ final class DoctrineDatabase extends Handler
...
-        $this->contentUpdater = $contentUpdater;
     }
 
     public function updateContentObjects(Type $fromType, Type $toType): void
     {
-        $this->contentUpdater->applyUpdates(
-            $fromType->id,
-            $this->contentUpdater->determineActions($fromType, $toType)
-        );
+        // Do nothing, content objects are no longer updated
```
alongosz, line 32: Actually, given you dropped the method call in `eZ/Publish/Core/Persistence/Legacy/Content/Type/Handler.php:595` it's not necessary to remove that "feature" from this gateway.
What I would suggest is deprecating the method on `\eZ\Publish\Core\Persistence\Legacy\Content\Type\Update\Handler` stating that it's due to its low scalability and performance.

eZ/Publish/Core/Persistence/Legacy/Tests/Content/MapperTest.php
```diff
@@ -169,36 +169,110 @@ public function testExtractContentFromRows()
...
+    public function testExtractContentFromRowsWithRemovedFieldDefinitions(): void
+    {
+        $rowsFixture = $this->getContentExtractFixture();
+        $nameRowsFixture = $this->getNamesExtractFixture();
+
+        $contentType = $this->getContentTypeFromRows($rowsFixture);
+        $contentType->fieldDefinitions = array_filter($contentType->fieldDefinitions, static function (Content\Type\FieldDefinition $fieldDefinition) {
+            // ref. fixtures, ezauthor
+            return $fieldDefinition->id !== 185;
+        });
```
alongosz: CS: line too long:
```suggestion
        $contentType->fieldDefinitions = array_filter(
            $contentType->fieldDefinitions,
            static function (Content\Type\FieldDefinition $fieldDefinition) {
                // ref. fixtures, ezauthor
                return $fieldDefinition->id !== 185;
            }
        );
```

eZ/Publish/Core/Persistence/Legacy/Tests/Content/MapperTest.php
```diff
@@ -169,36 +169,110 @@ public function testExtractContentFromRows()
...
+
+    public function testExtractContentFromRowsWithRemovedFieldDefinitions(): void
+    {
+        $rowsFixture = $this->getContentExtractFixture();
+        $nameRowsFixture = $this->getNamesExtractFixture();
+
+        $contentType = $this->getContentTypeFromRows($rowsFixture);
+        $contentType->fieldDefinitions = array_filter($contentType->fieldDefinitions, static function (Content\Type\FieldDefinition $fieldDefinition) {
+            // ref. fixtures, ezauthor
+            return $fieldDefinition->id !== 185;
```
alongosz: how exactly the fixtures affect the test that you need to remove it?

eZ/Publish/Core/Persistence/Legacy/Tests/Content/MapperTest.php
```diff
@@ -251,6 +323,30 @@ public function testExtractContentFromRowsMultipleVersions()
         );
     }
 
+    /**
+     * @param string[] $fields
+     */
+    private function getFieldRegistry(
+        array $fields = [],
+        int $expectedConverterCalls = null
```
alongosz:
```suggestion
        ?int $expectedConverterCalls = null
```

eZ/Publish/Core/Persistence/Legacy/Tests/Content/MapperTest.php
```diff
@@ -251,6 +323,30 @@ public function testExtractContentFromRowsMultipleVersions()
         );
     }
 
+    /**
+     * @param string[] $fields
+     */
+    private function getFieldRegistry(
+        array $fields = [],
```
alongosz: naming: the expected contents of that variable is not self-explanatory enough:
How about:
```suggestion
        array $fieldTypeIdentifiers = [],
```

eZ/Publish/Core/Persistence/Legacy/Tests/Content/MapperTest.php
```diff
@@ -251,6 +323,30 @@ public function testExtractContentFromRowsMultipleVersions()
...
+            $expectedConverterCalls === null
+            ? $this->any()
+            : $this->exactly($expectedConverterCalls)
+        )
+            ->method('toFieldValue')
+            ->will($this->returnValue(new FieldValue()));
+
+        $converters = [];
+        foreach ($fields as $field) {
+            $converters[$field] = $convMock;
```
alongosz: Edit: seems reasoning for this change got lost when editing comment... Congrats Microsoft GitHub UI.

1. any and exactly are static method and should be called statically. I don't insist this change for other places (e.g. `$this->assert...`) because only here we have unrelated vicinity. Though we should always use `self::`.
2. will -> willReturn already explained by Steve earlier I see
3. We try not to abbreviate variable names due to readability (unless full name makes it less readable, which is not the case here).

```suggestion
        $converterMock = $this->createMock(Converter::class);
        $converterMock->expects(
            $expectedConverterCalls === null
            ? self::any()
            : self::exactly($expectedConverterCalls)
        )
            ->method('toFieldValue')
            ->willReturn(new FieldValue());

        $converters = [];
        foreach ($fields as $field) {
            $converters[$field] = $converterMock;
```

eZ/Publish/SPI/Tests/FieldType/BaseIntegrationTest.php
```diff
@@ -445,10 +445,10 @@ public function testLoadFieldType($content)
     {
         $this->assertSame(
             $this->getTypeName(),
-            $content->fields[2]->type
+            $content->fields[1]->type
         );
 
-        return $content->fields[2];
+        return $content->fields[1];
```
alongosz: What happened here?

alongosz (commented):
The direction here makes perfect sense. We need to test this extensively of course:

Code-wise remarks:

eZ/Publish/Core/Persistence/Legacy/Content/Mapper.php
Nattfarinn: I've refactored it to additionally use internal cache (I wish PHP 7.3 had `??=` operator):

```php
    $contentTypes = [];
    // ...
    $contentType = $contentTypes[$contentTypeId] = $contentTypes[$contentTypeId]
        ?? $this->contentTypeHandler->load($contentTypeId);
```

eZ/Publish/Core/Persistence/Legacy/Content/Type/Update/Handler/DoctrineDatabase.php
Nattfarinn: It is necessary because otherwise it leads to circular reference (dependency on `ContentUpdater` must be dropped).

eZ/Publish/Core/Persistence/Legacy/Tests/Content/MapperTest.php
Nattfarinn: I reuse existing fixtures to test if content objects are built without fields removed from content type, in this case it is field `185` of `ezauthor` type.

eZ/Publish/Core/Persistence/Legacy/Content/Mapper.php
```diff
@@ -176,64 +181,146 @@ public function convertToStorageValue(Field $field)
...
+        $languageCodes = [];
+
         foreach ($nameRows as $row) {
-            $contentId = (int)$row['ezcontentobject_name_contentobject_id'];
-            $versionNo = (int)$row['ezcontentobject_name_content_version'];
-            $versionedNameData[$contentId][$versionNo][$row['ezcontentobject_name_content_translation']] = $row['ezcontentobject_name_name'];
+            $contentId = (int)$row["{$prefix}name_contentobject_id"];
+            $versionNo = (int)$row["{$prefix}name_content_version"];
+            $versionedNameData[$contentId][$versionNo][$row["{$prefix}name_content_translation"]] = $row["{$prefix}name_name"];
+            $languageCodes[] = $row["{$prefix}name_content_translation"];
```
alongosz: > `$languageCodes`: Array is only updated, but never queried

AFAIR in previous version it was used later on in the part which I asked to be encapsulated as `loadCachedVersionFieldDefinitionsPerLanguage`. Maybe you can still pass it there and re-use it? if not, then the variable itself is needed.

eZ/Publish/Core/Persistence/Legacy/Content/Mapper.php
```diff
@@ -176,64 +181,146 @@ public function convertToStorageValue(Field $field)
...
+            $contentId = (int)$row["{$prefix}id"];
+            $versionId = (int)$row["{$prefix}version_id"];
+            $contentTypeId = (int)$row["{$prefix}contentclass_id"];
+            $languageMask = (int)$row["{$prefix}version_language_mask"];
+
+            if (!isset($fieldDefinitions[$contentId][$versionId])) {
+                $contentType = $contentTypes[$contentTypeId] = $contentTypes[$contentTypeId]
+                    ?? $this->contentTypeHandler->load($contentTypeId);
+                foreach ($contentType->fieldDefinitions as $fieldDefinition) {
+                    $languageCodes = $this->extractLanguageCodesFromMask($languageMask, $allLanguages);
```
alongosz: So, if it's better to compute it on-the-fly instead of reusing mentioned `$languageCodes` prepared earlier, I think this can be done right below L306, right? The result is going to be deterministic for each iteration of that inner foreach?

eZ/Publish/Core/Persistence/Legacy/Content/Mapper.php
```diff
@@ -176,64 +181,146 @@ public function convertToStorageValue(Field $field)
...
+
                 $results[] = $content;
             }
         }
 
         return $results;
     }
 
+    /**
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
```
alongosz, line 346: missing `@return` type array shape and `@param array $rows` argument array-shape. 
We don't have PHPStan in kernel 1.3 to report that :(

eZ/Publish/Core/Persistence/Legacy/Content/Mapper.php
```diff
@@ -176,64 +181,146 @@ public function convertToStorageValue(Field $field)
...
+        $contentTypes = [];
+        $allLanguages = $this->loadAllLanguagesWithIdKey();
+
+        foreach ($rows as $row) {
+            $contentId = (int)$row["{$prefix}id"];
+            $versionId = (int)$row["{$prefix}version_id"];
+            $contentTypeId = (int)$row["{$prefix}contentclass_id"];
+            $languageMask = (int)$row["{$prefix}version_language_mask"];
+
+            if (!isset($fieldDefinitions[$contentId][$versionId])) {
```
alongosz: Wouldn't it be better to do early continue and one less indent later?
```php
if (isset($fieldDefinitions[$contentId][$versionId])) {
    continue;
}
```

eZ/Publish/Core/Persistence/Legacy/Content/Mapper.php
```diff
@@ -176,64 +181,146 @@ public function convertToStorageValue(Field $field)
...
+
+    /**
+     * @return \eZ\Publish\SPI\Persistence\Content[]
+     */
+    private function buildContentObjects(
+        array $contentInfos,
+        array $versionInfos,
+        array $fields,
+        array $fieldDefinitions,
+        array $versionedNameData
```
alongosz: PHPDoc: Missing array-shapes for arguments.

eZ/Publish/Core/Persistence/Legacy/Content/Mapper.php
```diff
@@ -176,64 +181,146 @@ public function convertToStorageValue(Field $field)
...
+    }
+
+    /**
+     * @return \eZ\Publish\SPI\Persistence\Content[]
+     */
+    private function buildContentObjects(
+        array $contentInfos,
+        array $versionInfos,
+        array $fields,
+        array $fieldDefinitions,
```
alongosz: Naming: it confused me for a sec. For this method it is `$missingFieldDefinitions`, right? Unless I misunderstood what's happening here.

alongosz (dismissed):

kisztof (approved):

eZ/Publish/Core/Persistence/Legacy/Content/Mapper.php
```diff
@@ -16,12 +16,20 @@
...
 /**
  * Mapper for Content Handler.
  *
  * Performs mapping of Content objects.
+ *
+ * @phpstan-type TVersionedLanguageFieldDefinitionsMap array<int, array<int, array<string, array<int, \eZ\Publish\SPI\Persistence\Content\Type\FieldDefinition>>>>
+ * @phpstan-type TVersionedFieldMap array<int, array<int, array<int, \eZ\Publish\SPI\Persistence\Content\Field>>>
+ * @phpstan-type TVersionedNameMap array<int, array<int, array<string, array<int, string>>>>
+ * @phpstan-type TContentInfoMap array<int, \eZ\Publish\SPI\Persistence\Content\ContentInfo>
+ * @phpstan-type TVersionInfoMap array<int, array<int, \eZ\Publish\SPI\Persistence\Content\VersionInfo>>
```
Steveb-p: You may consider splitting these into multiple lines to increase readability, to make it easier to figure out what elements lie at which depth :)

```suggestion
 * @phpstan-type TVersionedLanguageFieldDefinitionsMap array<
       int, array<
           int, array<
               string, array<
                   int, \eZ\Publish\SPI\Persistence\Content\Type\FieldDefinition,
               >
           >
       >
   >
 * @phpstan-type TVersionedFieldMap array<
     int, array<
          int, array<
              int, \eZ\Publish\SPI\Persistence\Content\Field,
          >
      >
  >
 * @phpstan-type TVersionedNameMap array<
     int, array<
         int, array<
             string, array<int, string>
         >
     >
 >
 * @phpstan-type TContentInfoMap array<int, \eZ\Publish\SPI\Persistence\Content\ContentInfo>
 * @phpstan-type TVersionInfoMap array<
     int, array<
         int, \eZ\Publish\SPI\Persistence\Content\VersionInfo,
     >
 >
```

Or

```suggestion
 * @phpstan-type TVersionedLanguageFieldDefinitionsMap array<
       int, 
       array<
           int, 
           array<
               string, 
               array<int, \eZ\Publish\SPI\Persistence\Content\Type\FieldDefinition>
           >
       >
   >
 * @phpstan-type TVersionedFieldMap array<
     int, 
     array<
          int, 
          array<int, \eZ\Publish\SPI\Persistence\Content\Field>
      >
  >
 * @phpstan-type TVersionedNameMap array<
     int, 
     array<
         int, 
         array<
             string, array<int, string>
         >
     >
 >
 * @phpstan-type TContentInfoMap array<int, \eZ\Publish\SPI\Persistence\Content\ContentInfo>
 * @phpstan-type TVersionInfoMap array<
     int, 
     array<int, \eZ\Publish\SPI\Persistence\Content\VersionInfo>
 >
```

Any that makes it easier to read, of course. It highly depends on a person :)

eZ/Publish/Core/Persistence/Legacy/Content/FieldHandler.php
```diff
@@ -148,9 +148,13 @@ protected function getEmptyField(FieldDefinition $fieldDefinition, $languageCode
      *
      * @param \eZ\Publish\SPI\Persistence\Content $content
      */
-    public function createExistingFieldsInNewVersion(Content $content)
+    public function createExistingFieldsInNewVersion(Content $content): void
     {
         foreach ($content->fields as $field) {
+            if ($field->id === null) {
```
Steveb-p, line 154:
```suggestion
            if ($field->getId() === null) {
```
or
```suggestion
            if ($field->isVirtual()) {
```

eZ/Publish/Core/Persistence/Legacy/Content/Mapper.php
```diff
@@ -176,64 +187,156 @@ public function convertToStorageValue(Field $field)
...
+            $contentTypeId = (int)$row["{$prefix}contentclass_id"];
+            $languageMask = (int)$row["{$prefix}version_language_mask"];
+
+            if (isset($fieldDefinitions[$contentId][$versionId])) {
+                continue;
+            }
+
+            $languageCodes = $this->extractLanguageCodesFromMask($languageMask, $allLanguages);
+            $contentType = $contentTypes[$contentTypeId] = $contentTypes[$contentTypeId]
+                ?? $this->contentTypeHandler->load($contentTypeId);
```
Steveb-p: I was looking at what we are doing here. I had to look into operator precedence, since I didn't immediately recall if assignment operator has lower priority over null-coalescing.

https://www.php.net/manual/en/language.operators.precedence.php

```suggestion
            $contentTypes[$contentTypeId] ??= $this->contentTypeHandler->load($contentTypeId);
            $contentType = $contentTypes[$contentTypeId];
```
or
```suggestion
            $contentType = $contentTypes[$contentTypeId] ??= $this->contentTypeHandler->load($contentTypeId);
```

eZ/Publish/Core/Persistence/Legacy/Content/StorageHandler.php
```diff
@@ -79,7 +79,7 @@ public function copyFieldData(VersionInfo $versionInfo, Field $field, Field $ori
     public function getFieldData(VersionInfo $versionInfo, Field $field)
     {
         $storage = $this->storageRegistry->getStorage($field->type);
-        if ($storage->hasFieldData()) {
+        if ($field->id !== null && $storage->hasFieldData()) {
```
Steveb-p, line 82:
```suggestion
        if ($field->isVirtual() === false && $storage->hasFieldData()) {
```
or
```suggestion
        if ($field->getId() !== null && $storage->hasFieldData()) {
```

eZ/Publish/Core/Persistence/Legacy/Tests/Content/MapperTest.php
```diff
@@ -169,36 +169,108 @@ public function testExtractContentFromRows()
...
+            'ezkeyword',
+        ], count($rowsFixture) - 2);
+
+        $mapper = new Mapper($reg, $this->getLanguageHandler(), $contentTypeHandlerMock);
         $result = $mapper->extractContentFromRows($rowsFixture, $nameRowsFixture);
 
+        $expectedContent = $this->getContentExtractReference();
+        $expectedContent->fields = array_values(array_filter($expectedContent->fields, static function (Field $field) {
+            return $field->fieldDefinitionId !== 185;
+        }));
```
Steveb-p:
```suggestion
        $expectedContent->fields = array_values(
            array_filter($expectedContent->fields, static function (Field $field): bool {
                return $field->fieldDefinitionId !== 185;
            })
        );
```

eZ/Publish/Core/Persistence/Legacy/Tests/Content/MapperTest.php
```diff
@@ -169,36 +169,108 @@ public function testExtractContentFromRows()
...
+
+    public function testExtractContentFromRowsWithRemovedFieldDefinitions(): void
+    {
+        $rowsFixture = $this->getContentExtractFixture();
+        $nameRowsFixture = $this->getNamesExtractFixture();
+
+        $contentType = $this->getContentTypeFromRows($rowsFixture);
+        $contentType->fieldDefinitions = array_filter(
+            $contentType->fieldDefinitions,
+            static function (Content\Type\FieldDefinition $fieldDefinition) {
```
Steveb-p:
```suggestion
            static function (Content\Type\FieldDefinition $fieldDefinition): bool {
```

eZ/Publish/Core/Persistence/Legacy/Tests/Content/MapperTest.php
```diff
@@ -251,6 +321,30 @@ public function testExtractContentFromRowsMultipleVersions()
...
+     */
+    private function getFieldRegistry(
+        array $fieldTypeIdentifiers = [],
+        ?int $expectedConverterCalls = null
+    ): Registry {
+        $converterMock = $this->createMock(Converter::class);
+        $converterMock->expects(
+            $expectedConverterCalls === null
+            ? self::any()
+            : self::exactly($expectedConverterCalls)
```
Steveb-p: I usually indent those, but that is a personal preference, so feel free to ignore.
```suggestion
                ? self::any()
                : self::exactly($expectedConverterCalls)
```

eZ/Publish/Core/Persistence/Legacy/Tests/Content/MapperTest.php
```diff
@@ -637,4 +733,39 @@ static function ($languageCode) use ($languages) {
 
         return $this->languageHandler;
     }
+
+    /**
+     * @return Content\Type\Handler|\PHPUnit\Framework\MockObject\MockObject
```
Steveb-p:
```suggestion
     * @return \eZ\Publish\SPI\Persistence\Content\Type\Handler&\PHPUnit\Framework\MockObject\MockObject
```

eZ/Publish/Core/Persistence/Legacy/Tests/Content/StorageHandlerTest.php
```diff
@@ -127,6 +128,34 @@ public function testGetFieldDataNotAvailable()
...
+
+        $storageMock->expects(self::never())
+            ->method('hasFieldData');
+
+        $storageMock->expects(self::never())
+            ->method('getFieldData');
+
+        $storageRegistryMock->expects(self::once())
+            ->method('getStorage')
+            ->with($this->equalTo('foobar'))
```
Steveb-p:
```suggestion
            ->with(self::equalTo('foobar'))
```
or (since it's exactly the same as)
```suggestion
            ->with('foobar')
```

You can also use
```suggestion
            ->with(self::identicalTo('foobar'))
```
which is as if you used `assertSame` assertion (with strict comparison and all that).

Steveb-p (approved):
LGTM. All my comments are optional notes that you can ignore as you see fit, since they are mostly opinion-based.

One thing I would very much like to add, but would require effort (probably a lot), is an integration test for it. It would be extremely valuable in my opinion, as we are dealing with a behavior that is very specific and depends on interaction between multiple components.

eZ/Publish/API/Repository/Values/Content/Field.php
```diff
@@ -85,4 +86,9 @@ public function getFieldTypeIdentifier(): string
     {
         return $this->fieldTypeIdentifier;
     }
+
+    public function isVirtual(): bool
```
Steveb-p, line 93: `@phpstan-assert-if-true` can be used to indicate a relationship between this method and others.

https://phpstan.org/writing-php-code/narrowing-types#custom-type-checking-functions-and-methods

```suggestion
    /**
     * @phpstan-assert-if-true !null $this->getId()
     */
    public function isVirtual(): bool
```

micszo (changes requested):
Following issue found in image asset.

Steps:

0. image CT (default), image asset CT (ezstring, ezimageasset)
1. create image
2. create image asset, select from repository
3. edit image CT, add single relation field or keywords 
4. edit image asset
5. open image editor
6. flip image
7. save as / apply changes to all

Actual result: In UI error occurs "Internal sever error". Exception occurs in log:

`php.CRITICAL: Uncaught Error: reset(): Argument #1 ($array) must be of type array, null given {"exception":"[object] (TypeError(code: 0): reset(): Argument #1 ($array) must be of type array, null given at /Users/michalszoltysek/Projects/workspace/ibexa_website_4/vendor/ezsystems/ezplatform-kernel/eZ/Publish/Core/FieldType/Relation/SearchField.php:32)"} []`

or

`php.CRITICAL: Uncaught Error: implode(): Argument #1 ($pieces) must be of type array, string given {"exception":"[object] (TypeError(code: 0): implode(): Argument #1 ($pieces) must be of type array, string given at /Users/michalszoltysek/Projects/workspace/ibexa_website_4/vendor/ezsystems/ezplatform-kernel/eZ/Publish/Core/FieldType/Keyword/SearchField.php:29)"} []`

micszo (changes requested):
Retest status.

- ezobjectrelation: OK
- ezkeyword: NOK - the same exception.

pspanja:
For what it's worth, there was a way to process Content update asynchronously, removed as deprecated in the PR: https://github.com/ezsystems/ezpublish-kernel/pull/2976

IMHO, this should be properly handled by asynchronous processing and not by introducing inconsistencies to the Repository API.

micszo (approved):
QA Approved on Ibexa Commerce 3.3.38-dev.

eZ/Publish/Core/Persistence/Legacy/Tests/Content/MapperTest.php
```diff
@@ -676,6 +713,20 @@ protected function getRelationCreateStructFixture()
         return $struct;
     }
 
+    protected function getEventDispatcher(): EventDispatcherInterface
+    {
+        $eventDispatcher = new EventDispatcher();
+        $eventDispatcher->addSubscriber(
+            new ResolveVirtualFieldSubscriber(
+                $this->getValueConverterRegistryMock(),
+                new StorageRegistry([]),
```
webhdx: Why using the class directly instead of mocking it?

eZ/Publish/Core/Search/Legacy/Tests/Content/AbstractTestCase.php
```diff
@@ -113,4 +118,18 @@ protected function getConverterRegistry()
...
         return $this->converterRegistry;
     }
+
+    protected function getEventDispatcher(): EventDispatcherInterface
+    {
+        $eventDispatcher = new EventDispatcher();
+        $eventDispatcher->addSubscriber(
+            new ResolveVirtualFieldSubscriber(
+                $this->getConverterRegistry(),
+                new StorageRegistry([]),
```
webhdx: I'd rather have it mocked.

Steveb-p:
> For what it's worth, there was a way to process Content update asynchronously, removed as deprecated in the PR: [ezsystems/ezpublish-kernel#2976](https://github.com/ezsystems/ezpublish-kernel/pull/2976)
> 
> IMHO, this should be properly handled by asynchronous processing and not by introducing inconsistencies to the Repository API.

Yes, ideally an update operation would be pushed into asynchronous queue. However, that does not prevent the object in question (in this case, Content Object that is related to Content Type being changed) from being accessed. 

Unless you intend to push the Content Type update into asynchronous processing too, but that doesn't really solve the issue - having a database transaction (to prevent application crashing on access) that spans across thousands of rows, that takes multiple minutes to complete, is likely to be an issue in itself.

So from my point of view, the best way of solving it is updating the Content Object when it's being accessed, in case the access operation happens before the Content Object update is run. Whether it is updated by a queue in the background is irrelevant, because the issue is accessing the item during that operation.

eZ/Publish/Core/Persistence/Legacy/Tests/Content/MapperTest.php
Steveb-p: If the storage isn't used, and we don't need stub functionality (checking method calls), then it's fine to use the class directly. Unless there is an interface declared for `StorageRegistry`, in which case I'd also prefer a mock of that interface.

eZ/Publish/Core/Persistence/Legacy/Tests/Content/StorageHandlerTest.php
```diff
@@ -127,6 +128,34 @@ public function testGetFieldDataNotAvailable()
...
+        $field->type = 'foobar';
+        $field->value = new FieldValue();
+
+        $handler = $this->getStorageHandler();
+        $handler->getFieldData($this->getVersionInfoMock(), $field);
+    }
+
+    /**
+     * @covers \eZ\Publish\Core\Persistence\Legacy\Content\StorageHandler::getFieldData
+     */
```
Steveb-p: Using phpdoc `@covers` annotation on a method is highly discouraged by PHPUnit.

The effect of this annotation is that it limits coverage report (which we don't really use) to this method only, while there is potential for private/protected methods to be called too.

I'd remove it.

```suggestion
```

eZ/Publish/Core/Persistence/Legacy/Tests/Content/StorageHandlerTest.php
```diff
@@ -127,6 +128,34 @@ public function testGetFieldDataNotAvailable()
...
+    public function testGetFieldDataNotAvailableForVirtualField()
```
Steveb-p:
```suggestion
    public function testGetFieldDataNotAvailableForVirtualField(): void
```

eZ/Bundle/EzPublishDebugBundle/Twig/DebugTemplate.php
```diff
@@ -19,7 +19,7 @@ class DebugTemplate extends Template
 {
     private $fileSystem;
 
-    public function display(array $context, array $blocks = [])
+    public function display(array $context, array $blocks = []): void
```
alongosz: This has been fixed already on 1.3 via https://github.com/ezsystems/ezplatform-kernel/pull/406

eZ/Publish/Core/Persistence/Legacy/Content/Mapper.php
```diff
@@ -16,15 +16,50 @@
...
+             string, array<int, string>
+         >
+     >
+ >
+ * @phpstan-type TContentInfoMap array<int, \eZ\Publish\SPI\Persistence\Content\ContentInfo>
+ * @phpstan-type TVersionInfoMap array<
+     int, array<
+         int, \eZ\Publish\SPI\Persistence\Content\VersionInfo,
+     >
+ >
```
alongosz: Malformed PHPDoc block
```suggestion
 *
 * @phpstan-type TVersionedLanguageFieldDefinitionsMap array<
 *     int, array<
 *         int, array<
 *             string, array<
 *                 int, \eZ\Publish\SPI\Persistence\Content\Type\FieldDefinition,
 *             >
 *         >
 *     >
 * >
 * @phpstan-type TVersionedFieldMap array<
 *   int, array<
 *        int, array<
 *            int, \eZ\Publish\SPI\Persistence\Content\Field,
 *        >
 *    >
 * >
 * @phpstan-type TVersionedNameMap array<
 *   int, array<
 *       int, array<
 *           string, array<int, string>
 *       >
 *   >
 * >
 * @phpstan-type TContentInfoMap array<int, \eZ\Publish\SPI\Persistence\Content\ContentInfo>
 * @phpstan-type TVersionInfoMap array<
 *   int, array<
 *       int, \eZ\Publish\SPI\Persistence\Content\VersionInfo,
 *   >
 * >
```

eZ/Publish/Core/Persistence/Legacy/Content/Mapper.php
```diff
@@ -16,15 +16,50 @@
...
  */
 class Mapper
 {
+    public const EMPTY_FIELD_ID = -1;
```
alongosz: This doesn't seem to be used anywhere.

eZ/Publish/Core/Persistence/Legacy/Tests/Content/MapperTest.php
```diff
@@ -169,36 +179,126 @@ public function testExtractContentFromRows()
...
+            $reg,
+            $this->getLanguageHandler(),
+            $contentTypeHandlerMock,
+            $this->getEventDispatcher()
+        );
+        $result = $mapper->extractContentFromRows($rowsFixture, $nameRowsFixture);
+
+        $expected = [$this->getContentExtractReference()];
+
+        $this->assertEquals(
```
alongosz:
```suggestion
        self::assertEquals(
```

eZ/Publish/Core/Persistence/Legacy/Tests/Content/MapperTest.php
```diff
@@ -169,36 +179,126 @@ public function testExtractContentFromRows()
...
+            'ezdatetime',
+            'ezkeyword',
+        ], count($rowsFixture) - 1);
+
+        $mapper = new Mapper(
+            $reg,
+            $this->getLanguageHandler(),
+            $contentTypeHandlerMock,
+            $this->getEventDispatcher()
+        );
```
alongosz, line 202: Question: any reason these snippets in:
* `testExtractContentFromRows`
* `testExtractContentFromRowsWithNewFieldDefinitions`
* `testExtractContentFromRowsWithRemovedFieldDefinitions`

are only slightly different? Maybe they can be the same and extracted to common method, like `$mapper = $this->buildMapper(...`
Or is this on purpose?

eZ/Publish/Core/Persistence/Legacy/Tests/Content/MapperTest.php
```diff
@@ -637,4 +786,39 @@ static function ($languageCode) use ($languages) {
 
         return $this->languageHandler;
     }
+
+    /**
+     * @return \eZ\Publish\SPI\Persistence\Content\Type\Handler&\PHPUnit\Framework\MockObject\MockObject
+     */
+    protected function getContentTypeHandler()
```
alongosz: Missing strict type:

```suggestion
    protected function getContentTypeHandler(): Handler
```
(convention: we type-hint intersection for mocks via PHPDoc, but strictly type the actual type being mocked)

eZ/Publish/Core/Persistence/Legacy/Tests/Content/StorageHandlerTest.php
alongosz: > Using phpdoc `@covers` annotation on a method is highly discouraged by PHPUnit.
> 
> The effect of this annotation is that it limits coverage report (which we don't really use) to this method only, while there is potential for private/protected methods to be called too.
> 
> I'd remove it.

In this case there should be
```php
/**
 * @covers \eZ\Publish\Core\Persistence\Legacy\Content\StorageHandler
 */
```
above the entire class.

eZ/Publish/Core/settings/storage_engines/legacy/content.yml
```diff
@@ -9,6 +9,16 @@ services:
...
             - "@ezpublish.persistence.legacy.field_value_converter.registry"
             - "@ezpublish.spi.persistence.legacy.language.handler"
+            - '@ezpublish.spi.persistence.legacy.content_type.handler'
+            - '@Symfony\Contracts\EventDispatcher\EventDispatcherInterface'
+
+    Ibexa\Core\Persistence\Legacy\Content\Mapper\ResolveVirtualFieldSubscriber:
+        arguments:
+            - "@ezpublish.persistence.legacy.field_value_converter.registry"
+            - "@ezpublish.persistence.external_storage_registry"
+            - "@ezpublish.persistence.legacy.content.gateway"
```
alongosz: Inconsistent quoting, also please use named arguments in new DIC, it's easier to maintain.
```suggestion
            $converterRegistry: '@ezpublish.persistence.legacy.field_value_converter.registry'
            $storageRegistry: '@ezpublish.persistence.external_storage_registry'
            $contentGateway: '@ezpublish.persistence.legacy.content.gateway'
```

src/contracts/Event/Mapper/ResolveMissingFieldEvent.php
```diff
@@ -0,0 +1,73 @@
...
+    private $content;
+
+    /** @var \eZ\Publish\SPI\Persistence\Content\Type\FieldDefinition */
+    private $fieldDefinition;
+
+    /** @var string */
+    private $languageCode;
+
+    /** @var array */
+    private $context;
```
alongosz: This array needs a shape. If it can be anything it should be stated explicitly: `array<mixed>`. But maybe it's meant to be `array<string, mixed>` or `array<string, scalar>`? Or something more specific? TL;DR;

src/contracts/Event/Mapper/ResolveMissingFieldEvent.php
```diff
@@ -0,0 +1,73 @@
...
+
+    /** @var \eZ\Publish\SPI\Persistence\Content\Field|null */
+    private $field;
+
+    public function __construct(
+        Content $content,
+        FieldDefinition $fieldDefinition,
+        string $languageCode,
+        array $context = []
```
alongosz, line 38: and similarly here

src/contracts/FieldType/DefaultDataFieldStorage.php
```diff
@@ -0,0 +1,26 @@
...
+interface DefaultDataFieldStorage
+{
+    /**
+     * Populates $field value property with default data based on the external data.
+     * $field->value is a {@link \eZ\Publish\SPI\Persistence\Content\FieldValue} object.
+     * This value holds the data as a {@link \eZ\Publish\Core\FieldType\Value} based object, according to
+     * the field type (e.g. for TextLine, it will be a {@link \eZ\Publish\Core\FieldType\TextLine\Value} object).
+     *
+     * @param \eZ\Publish\SPI\Persistence\Content\VersionInfo $versionInfo
+     * @param \eZ\Publish\SPI\Persistence\Content\Field $field
```
alongosz: This is redundant and can be skipped
```suggestion
```

src/contracts/FieldType/DefaultDataFieldStorage.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace Ibexa\Contracts\FieldType;
```
alongosz: This namespace is not correct. `./src/contracts` is mapped to `Ibexa\Contracts\Core\`.
Moreover I would move it one level down, to:
```suggestion
namespace Ibexa\Contracts\Core\FieldType\Storage;
```

src/contracts/FieldType/DefaultDataFieldStorage.php
```diff
@@ -0,0 +1,26 @@
...
+use eZ\Publish\SPI\Persistence\Content\Field;
+use eZ\Publish\SPI\Persistence\Content\VersionInfo;
+
+interface DefaultDataFieldStorage
+{
+    /**
+     * Populates $field value property with default data based on the external data.
+     * $field->value is a {@link \eZ\Publish\SPI\Persistence\Content\FieldValue} object.
+     * This value holds the data as a {@link \eZ\Publish\Core\FieldType\Value} based object, according to
+     * the field type (e.g. for TextLine, it will be a {@link \eZ\Publish\Core\FieldType\TextLine\Value} object).
```
alongosz: Sadly, this doesn't render well by our chosen API reference tool - phpDocumentor, because it doesn't handle `@link` for documentation elements (only hyperlinks).
image

It renders better when changed to:
```suggestion
     * Populates $field value property with default data based on the external data.
     *
     * <code>$field->value</code> is a {@see \eZ\Publish\SPI\Persistence\Content\FieldValue} object.
     * This value holds the data as a {@see \eZ\Publish\Core\FieldType\Value} based object, according to
     * the field type (e.g. for TextLine, it will be a {@see \eZ\Publish\Core\FieldType\TextLine\Value} object).
```
1. Note that by convention the first line of method's PHPDoc is an intro and should be followed by an empty line - this is also reflected by the way phpDocumentor renders doc - the first line is actually above the method as its brief intro).
2. References to documentation elements need to be referenced via `@see`.

The end result:
image

src/lib/Persistence/Legacy/Content/Mapper/ResolveVirtualFieldSubscriber.php
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    /**
+     * @throws \eZ\Publish\Core\Persistence\Legacy\Content\FieldValue\Converter\Exception\NotFound
+     */
+    public function persistExternalStorageField(ResolveMissingFieldEvent $event): void
+    {
+        $field = $event->getField();
+
+        if ($field && $field->id !== null) {
```
alongosz: $field is not of a boolean type
```suggestion
        if ($field !== null && $field->id !== null) {
```

src/lib/Persistence/Legacy/Content/Mapper/ResolveVirtualFieldSubscriber.php
```diff
@@ -0,0 +1,215 @@
...
+        );
+
+        $event->setField($field);
+    }
+
+    public function resolveVirtualExternalStorageField(ResolveMissingFieldEvent $event): void
+    {
+        $field = $event->getField();
+
+        if ($field && $field->id !== null) {
```
alongosz:
```suggestion
        if ($field !== null && $field->id !== null) {
```

src/lib/Persistence/Legacy/Content/Mapper/ResolveVirtualFieldSubscriber.php
```diff
@@ -0,0 +1,215 @@
...
+        );
+
+        $event->setField($field);
+
+        // Do not persist the external storage field
+        $event->stopPropagation();
+    }
+
+    /**
+     * @throws \eZ\Publish\Core\Persistence\Legacy\Content\FieldValue\Converter\Exception\NotFound
```
alongosz: We rather avoid reporting non-contract exceptions:
```suggestion
     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
```

src/lib/Persistence/Legacy/Content/Mapper/ResolveVirtualFieldSubscriber.php
```diff
@@ -0,0 +1,215 @@
...
+        $field->type = $fieldDefinition->fieldType;
+        $field->value = $this->getDefaultValue($fieldDefinition);
+        $field->languageCode = $languageCode;
+        $field->versionNo = $versionInfo->versionNo;
+
+        return $field;
+    }
+
+    /**
+     * @throws \eZ\Publish\Core\Persistence\Legacy\Content\FieldValue\Converter\Exception\NotFound
```
alongosz:
```suggestion
     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
```

alongosz (commented):
Some minor remarks:

src/contracts/Event/Mapper/ResolveMissingFieldEvent.php
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * @return array<mixed>
+     */
+    public function getContext(): array
+    {
+        return $this->context;
+    }
+
+    public function setField(?Field $field): void
```
Steveb-p, line 70: Is it valid to call `ResolveMissingFieldEvent::setField(null)`?

alongosz (approved):

src/contracts/Event/Mapper/ResolveMissingFieldEvent.php
Nattfarinn: Yes, it is intentional and there are valid use cases for that.

PR #399, opened 2024-01-05
IBX-7485: Implemented custom Filesystem that handles corrupted filepaths

barw4:
### **Proof of concept PR** 

| **JIRA issue**                          | [IBX-7485](https://issues.ibexa.co/browse/IBX-7485)

This PR overrides `league/flysystem`'s `Filesystem` class so it handles already existing files with broken paths gracefully.

I intentionally overrode only some of the methods of the original class as we still should disallow uploading files with broken names, this should fix the issue with broken names already existing in the filesystem/db.

Another idea (skipping broken files): https://github.com/ezsystems/ezplatform-kernel/pull/400

#### Checklist:

### DIFF
--- a/eZ/Publish/Core/IO/Filesystem.php
+++ b/eZ/Publish/Core/IO/Filesystem.php
@@ -1,0 +1,79 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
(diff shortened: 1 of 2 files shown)


alongosz (changes requested):
We rely on that filesystem in many places (nfs, multirepo) and regression coverage is still not what we aim to, so it's best to avoid touching it.

Favoring #400 over this.

PR #401, opened 2024-01-09
IBX-7172: Fixed Repository Filtering by multiple ObjectStateId criteria

vidarl:
| **JIRA issue**                          | [IBX-7172](https://issues.ibexa.co/browse/IBX-7172)

If user has policy with limitation on multiple object states, the api call for `$locationService->loadLocationChildren($location)` creates (`eZ\Publish\Core\Persistence\Legacy\Filter\CriterionQueryBuilder\Content\ObjectStateIdQueryBuilder`) an incorrect sql for checking the object states.

The sql looks as follows :
```
SELECT
  DISTINCT COUNT(DISTINCT location.node_id)
FROM
  ezcontentobject_tree location
  INNER JOIN ezcontentobject content ON content.id = location.contentobject_id
  INNER JOIN ezcontentobject_version version ON (
    content.id = version.contentobject_id
  )
  AND (
    content.current_version = version.version
  )
  AND (version.status = :dcValue1)
  INNER JOIN ezcobj_state_link object_state_link ON content.id = object_state_link.contentobject_id
  INNER JOIN ezcontent_language language ON language.id & version.language_mask = language.id
WHERE
  (
    location.parent_node_id IN (:dcValue2)
  )
  AND (
    (
      language.locale IN (:dcValue3)
    )
    OR (version.language_mask & 1 = 1)
  )
  AND (
    (
      content.section_id IN (:dcValue4)
    )
    AND (
      (
        object_state_link.contentobject_state_id IN (:dcValue5)
      )
      AND (
        object_state_link.contentobject_state_id IN (:dcValue6)
      )
    )
  )
```

Obviously, `object_state_link.contentobject_state_id` cannot be equal both `:dcValue5` and `:dcValue6` at the same time.
Instead the objet_state_link table should be joined once for each id and the sql should read:
```
SELECT
  DISTINCT COUNT(DISTINCT location.node_id)
FROM
  ezcontentobject_tree location
  INNER JOIN ezcontentobject content ON content.id = location.contentobject_id
  INNER JOIN ezcontentobject_version version ON (
    content.id = version.contentobject_id
  )
  AND (
    content.current_version = version.version
  )
  AND (version.status = :dcValue1)
  INNER JOIN ezcobj_state_link object_state_link_1 ON content.id = object_state_link.contentobject_id
  INNER JOIN ezcobj_state_link object_state_link_2 ON content.id = object_state_link.contentobject_id
  INNER JOIN ezcontent_language language ON language.id & version.language_mask = language.id
WHERE
  (
    location.parent_node_id IN (:dcValue2)
  )
  AND (
    (
      language.locale IN (:dcValue3)
    )
    OR (version.language_mask & 1 = 1)
  )
  AND (
    (
      content.section_id IN (:dcValue4)
    )
    AND (
      (
        object_state_link_1.contentobject_state_id IN (:dcValue5)
      )
      AND (
        object_state_link_2.contentobject_state_id IN (:dcValue6)
      )
    )
  )
```

#### Checklist:

### DIFF
--- a/eZ/Publish/API/Repository/Tests/Filtering/ContentFilteringTest.php
+++ b/eZ/Publish/API/Repository/Tests/Filtering/ContentFilteringTest.php
@@ -30,11 +30,6 @@ use function sprintf;
  */
 final class ContentFilteringTest extends BaseRepositoryFilteringTestCase
 {
-    protected function setUp(): void
-    {
-        parent::setUp();
-    }
-
     /**
      * Test that special cases of Location Sort Clauses are working correctly.
      *
--- a/eZ/Publish/API/Repository/Tests/Filtering/ContentFilteringTest.php
+++ b/eZ/Publish/API/Repository/Tests/Filtering/ContentFilteringTest.php
@@ -12,11 +12,14 @@ use function array_map;
 use function count;
 use eZ\Publish\API\Repository\Values\Content\Content;
 use eZ\Publish\API\Repository\Values\Content\ContentList;
+use eZ\Publish\API\Repository\Values\Content\Location;
 use eZ\Publish\API\Repository\Values\Content\Query;
 use eZ\Publish\API\Repository\Values\Content\Query\Criterion;
 use eZ\Publish\API\Repository\Values\Content\Query\SortClause;
 use eZ\Publish\API\Repository\Values\Content\Search\SearchHit;
 use eZ\Publish\API\Repository\Values\Filter\Filter;
+use eZ\Publish\API\Repository\Values\ObjectState\ObjectStateCreateStruct;
+use eZ\Publish\API\Repository\Values\ObjectState\ObjectStateGroupCreateStruct;
 use eZ\Publish\Core\FieldType\Keyword;
 use eZ\Publish\SPI\Repository\Values\Filter\FilteringSortClause;
 use IteratorAggregate;
@@ -27,6 +30,13 @@ use function sprintf;
  */
 final class ContentFilteringTest extends BaseRepositoryFilteringTestCase
 {
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->getSetupFactory()->getRepository(true);
+        $this->contentProvider = new TestContentProvider($this->getRepository(true), $this);
+    }
+
     /**
      * Test that special cases of Location Sort Clauses are working correctly.
      *
@@ -330,6 +340,64 @@ final class ContentFilteringTest extends BaseRepositoryFilteringTestCase
         $contentList = $this->find($filter, []);
         self::assertCount(1, $contentList);
         self::assertSame(57, $contentList->getIterator()[0]->id);
+    }
+
+    public function testObjectStateIdCriterionOnMultipleObjectStates(): void
+    {
+        $contentService = $this->getRepository()->getContentService();
+        $contentTypeService = $this->getRepository()->getContentTypeService();
+        $locationService = $this->getRepository()->getLocationService();
+        $objectStateService = $this->getRepository()->getObjectStateService();
+
+        // Create additional Object States
+        $objectStateGroupStruct = new ObjectStateGroupCreateStruct();
+        $objectStateGroupStruct->identifier = 'domain';
+        $objectStateGroupStruct->names = ['eng-GB' => 'Domain'];
+        $objectStateGroupStruct->defaultLanguageCode = 'eng-GB';
+        $objectStateGroup = $objectStateService->createObjectStateGroup($objectStateGroupStruct);
+
+        $objectStateCreateStruct = new ObjectStateCreateStruct();
+        $objectStateCreateStruct->identifier = 'public';
+        $objectStateCreateStruct->names = ['eng-GB' => 'Public'];
+        $objectStateCreateStruct->defaultLanguageCode = 'eng-GB';
+        $objectStatePublic = $objectStateService->createObjectState($objectStateGroup, $objectStateCreateStruct);
+
+        $objectStateCreateStruct->identifier = 'private';
+        $objectStateCreateStruct->names = ['eng-GB' => 'Private'];
+        $objectStatePrivate = $objectStateService->createObjectState($objectStateGroup, $objectStateCreateStruct);
+
+        // Create a new content object and assign object state "Private" to it:
+        $contentCreate = $contentService->newContentCreateStruct(
+            $contentTypeService->loadContentTypeByIdentifier('folder'),
+            'eng-GB'
+        );
+        $contentCreate->setField('name', 'Private Folder');
+        $content = $contentService->createContent(
+            $contentCreate,
+            [$locationService->newLocationCreateStruct(2)]
+        );
+        $contentService->publishVersion(
+            $content->getVersionInfo()
+        );
+        $objectStateService->setContentState($content->contentInfo, $objectStateGroup, $objectStatePrivate);
+
+        $filter = new Filter();
+        $filter
+            ->withCriterion(new Criterion\LogicalAnd([
+                new Criterion\ParentLocationId(2),
+                new Criterion\LogicalAnd([
+                    new Criterion\ObjectStateId(1),
+                    new Criterion\ObjectStateId(4),
+                ]),
+            ]));
+
+        $results = $this->find($filter);
+
+        self::assertEquals(1, $results->getTotalCount(), 'Expected to find only one object which has state "not_locked" and "private"');
+        /** @var \eZ\Publish\API\Repository\Values\Content\Location $result */
+        foreach ($results as $result) {
+            self::assertEquals($result->id, $content->id, 'Expected to find "Private Folder"');
+        }
     }
 
     public function getListOfSupportedSortClauses(): iterable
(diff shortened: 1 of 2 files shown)


eZ/Publish/Core/Persistence/Legacy/Filter/CriterionQueryBuilder/Content/ObjectStateIdQueryBuilder.php
```diff
@@ -29,19 +29,23 @@ public function buildQueryConstraint(
         FilteringQueryBuilder $queryBuilder,
         FilteringCriterion $criterion
     ): ?string {
+        static $counter = 1;
+        $tableAlias = 'object_state_link_' . $counter;
+        ++$counter;
```
alongosz: I'm not a huge fan of static variables. There's no way to know for how long the object will live (maybe it's a batch process?) and if it performs multiple queries then the counter increases even though it doesn't need to.

I'd use something like: 
```suggestion
        $tableAlias = uniqid('osl_');
```
to ensure every alias is unique.

alongosz (commented):
Foremost, please add an **integration** test case showing the issue.

eZ/Publish/Core/Persistence/Legacy/Filter/CriterionQueryBuilder/Content/ObjectStateIdQueryBuilder.php
vidarl: The precision of uniqid() is microseconds, might not be good enough ? : https://www.php.net/manual/en/function.uniqid.php#120123

Precision is increased by using second argument to function ($more_entropy), but that creates IDs in format `12345678.1234567`, and the dot would then need to be removed as it is not accepted in table aliases in sql

Some systems requires the use of `$more_entropy`, not limited to Cygwin....

Still okay with your suggestion, @alongosz  ?

vidarl:
> Foremost, please add an **integration** test case showing the issue.

Honestly, a bit unsure how you want that test. Didn't find a existing test I could reuse by just adding more fixture. So created a new test in 3271b673ddf01a2bd8d2fc06c8f4271d8235485f

eZ/Publish/API/Repository/Tests/Filtering/ContentFilteringTest.php
```diff
@@ -27,6 +30,13 @@
  */
 final class ContentFilteringTest extends BaseRepositoryFilteringTestCase
 {
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->getSetupFactory()->getRepository(true);
+        $this->contentProvider = new TestContentProvider($this->getRepository(true), $this);
```
vidarl: The test runs fine when executing ` time vendor/bin/phpunit -c phpunit-integration-legacy.xml --filter "ContentFilteringTest"`

However, it fails when running all tests ( without `--filter` parameter)..
Trying to reset the repository using the code above have no impact for whatever reason.. Any ideas?

eZ/Publish/Core/Persistence/Legacy/Filter/CriterionQueryBuilder/Content/ObjectStateIdQueryBuilder.php
Steveb-p: `uniqid` increases the internal counter if called multiple times. It ensure no collision, unless between multiple PHP processes.

Steveb-p: Behavior described in the comment in PHP docs does not seem to happen. It might be something that is specific to some environments.

Current `uniqid` implementation ensures that at least a milisecond has passed since generating ID, which makes it almost impossible to hit the same sequence twice.

https://github.com/php/php-src/blob/b1f5ca59d188337e4545f2e8df3d0e6936299fe4/ext/standard/uniqid.c#L55-L62.

This code has 7 years, so you can safely assume it holds true for our supported versions :sweat_smile:

vidarl: Roger. Applied suggestion. Thank you

eZ/Publish/API/Repository/Tests/Filtering/ContentFilteringTest.php
vidarl: I gave up trying to figure out how to reset the database before the test starts, but did instead change the test case a bit so it also work even if the newly created object state doesn't always get  the same id

eZ/Publish/API/Repository/Tests/Filtering/ContentFilteringTest.php
```diff
@@ -27,6 +30,11 @@
  */
 final class ContentFilteringTest extends BaseRepositoryFilteringTestCase
 {
+    protected function setUp(): void
+    {
+        parent::setUp();
+    }
```
alongosz: This is not needed.

eZ/Publish/API/Repository/Tests/Filtering/ContentFilteringTest.php
```diff
@@ -332,6 +340,69 @@ private function performAndAssertSimpleSortClauseQuery(FilteringSortClause $sort
         self::assertSame(57, $contentList->getIterator()[0]->id);
     }
 
+    public function testObjectStateIdCriterionOnMultipleObjectStates(): void
```
alongosz, line 341:
```suggestion
    /**
     * @throws \eZ\Publish\API\Repository\Exceptions\Exception
     */
    public function testObjectStateIdCriterionOnMultipleObjectStates(): void
```

eZ/Publish/API/Repository/Tests/Filtering/ContentFilteringTest.php
```diff
@@ -332,6 +340,69 @@ private function performAndAssertSimpleSortClauseQuery(FilteringSortClause $sort
...
+        );
+        $contentCreate->setField('name', 'Private Folder');
+        $content = $contentService->createContent(
+            $contentCreate,
+            [$locationService->newLocationCreateStruct(2)]
+        );
+        $contentService->publishVersion(
+            $content->getVersionInfo()
+        );
+        $objectStateService->setContentState($content->contentInfo, $objectStateGroup, $objectStatePrivate);
```
alongosz:
```suggestion
        $objectStateService->setContentState(
            $content->getVersionInfo()->getContentInfo(),
            $objectStateGroup,
            $objectStatePrivate
        );
```

eZ/Publish/API/Repository/Tests/Filtering/ContentFilteringTest.php
```diff
@@ -332,6 +340,69 @@ private function performAndAssertSimpleSortClauseQuery(FilteringSortClause $sort
...
+
+        $results = $this->find($filter);
+
+        self::assertEquals(
+            1,
+            $results->getTotalCount(),
+            'Expected to find only one object which has state "not_locked" and "private"'
+        );
+
+        /** @var \eZ\Publish\API\Repository\Values\Content\Location $result */
```
alongosz: It's going to be always Content item, not a Location, because you're filtering on `ContentService`. Moreover, `iterable` typehint as `\eZ\Publish\API\Repository\Values\Content\ContentList` of `find` method is enough, no type hinting should be necessary.
```suggestion
```

alongosz (commented):
Looks good, some minor tweaks are needed:

eZ/Publish/API/Repository/Tests/Filtering/ContentFilteringTest.php
vidarl: Indeed. Left-overs after working with the tests

vidarl:
FYI: Review comments implemented, @Steveb-p , @alongosz

alongosz (approved):

eZ/Publish/Core/Persistence/Legacy/Filter/CriterionQueryBuilder/Content/ObjectStateIdQueryBuilder.php
```diff
@@ -29,19 +29,21 @@ public function buildQueryConstraint(
         FilteringQueryBuilder $queryBuilder,
         FilteringCriterion $criterion
     ): ?string {
+        $tableAlias = uniqid('osl_');
+
         /** @var \eZ\Publish\API\Repository\Values\Content\Query\Criterion\ObjectStateId $criterion */
         $queryBuilder
             ->joinOnce(
```
Steveb-p: Consider it as a non merge blocking comment:

Just a remark - if we will always generate custom table alias for each `ObjectStateId` criteria, then it would make sense to switch into simpler `QueryBuilder::join`, which does not contain additional checks against table uniqueness.

Steveb-p (approved):
Test looks very good. :+1:

konradoboza (approved):

eZ/Publish/Core/Persistence/Legacy/Filter/CriterionQueryBuilder/Content/ObjectStateIdQueryBuilder.php
vidarl: Good point.. Changed `joinOnce()` into a `join()`

tomaszszopinski (approved):
QA approved on IbexaDXP 3.3 experience.

PR #402, opened 2024-02-13
IBX-7579:Richtext: Rows are added to ezurl_object_link on every save

vidarl:
| :ticket: Issue | [IBX-7579](https://issues.ibexa.co/browse/IBX-7579) |
|----------------|-----------|

#### Related PRs: 
- https://github.com/ezsystems/ezplatform-richtext/pull/242

#### Description:

Makes it possible to get data from `ezurl_object_link` table. This is needed by richtext bundle in order
to avoid duplicate links entries to be added to table.

Corresponding richtext PR : https://github.com/ezsystems/ezplatform-richtext/pull/242


eZ/Publish/Core/FieldType/Url/UrlStorage/Gateway/DoctrineStorage.php
```diff
@@ -129,6 +129,48 @@ public function insertUrl($url)
...
+                    ':contentobject_attribute_version'
+                )
+            )
+            ->setParameter(':contentobject_attribute_id', $fieldId, ParameterType::INTEGER)
+            ->setParameter(':contentobject_attribute_version', $versionNo, ParameterType::INTEGER);
+
+        $statement = $selectQuery->execute();
+        $rows = $statement->fetchAllAssociativeIndexed();
+        $result = [];
+        foreach ($rows as $url => $item) {
```
vidarl: It could be an option to return the array as-is, and instead to this foreach in [`vendor/ezsystems/ezplatform-richtext/src/lib/eZ/FieldType/RichText/RichTextStorage/Gateway.php`](https://github.com/ezsystems/ezplatform-richtext/pull/242/files#diff-c54260f9f2ffb301f5dfffd2821f50f16cceca1d7b05fcd2c3fcca31c4e4dc51)

vidarl:
The duplicate entries doesn't cause any harm besides performance and storage.
Should I also create an sql update script to fix the problem ?

vidarl:
@webhdx  : Rebased on top of latest 1.3 branch ( was just 2 commits behind though )

vidarl:
@webhdx : tmp commits removed

eZ/Publish/Core/FieldType/Tests/Url/Gateway/DoctrineStorageTest.php
```diff
@@ -194,6 +194,29 @@ public function testUnlinkUrl()
...
+
+        $gateway->linkUrl($urlIds[0], 10, 1);
+        $gateway->linkUrl($urlIds[1], 10, 1);
+        $gateway->linkUrl($urlIds[1], 12, 2);
+        $gateway->linkUrl($urlIds[2], 14, 1);
+
+        self::assertEquals(['https://ibexa.co/example1' => true, 'https://ibexa.co/example2' => true], $gateway->getUrlsFromUrlLink(10, 1), 'Did not get expected urlS for field 10');
+        self::assertEquals(['https://ibexa.co/example2' => true], $gateway->getUrlsFromUrlLink(12, 2), 'Did not get expected url for field 12');
+        self::assertEquals(['https://ibexa.co/example3' => true], $gateway->getUrlsFromUrlLink(14, 1), 'Did not get expected url for field 14');
+        self::assertEquals([], $gateway->getUrlsFromUrlLink(15, 1), 'Expected no urls for field 15');
```
alongosz: The problem with these tests here is that they're executed on SQLite only, so they don't test anything useful.
You rather should add proper integration test.

At some point (when I have time (tm)) these tests here will be dropped, as they negatively impact performance of unit test suite.

How does the bug manifest itself for the end user / on API level?

vidarl: Right.. So I have changed the test into an integration test.

The bug doesn't manifest itself as a functional defect. It causes duplicated rows in in the ezurl_object_link table. This is wrong and causes extra storage needs and some performance decrease ( not very noticeable in most cases though I )

eZ/Publish/Core/FieldType/Tests/Integration/Url/UrlStorage/UrlStorageGatewayTest.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace eZ\Publish\Core\FieldType\Tests\Integration\Url\UrlStorage;
```
alongosz: Any reason this base class for the test is in the old namespace? The actual test is in the new one, which is correct.
Actually, you can just place all of this in `tests/integration/Core/FieldType/Url/UrlStorage/Gateway/UrlDoctrineStorageGatewayTest.php`. It's separated due to legacy reasons and lack of refactoring. We used to have `DoctrineGateway` and `LegacyGateway` in some places, but the latter one was dropped a long time ago, so base test class is not needed either.

tests/integration/Core/FieldType/Url/UrlStorage/Gateway/UrlDoctrineStorageGatewayTest.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace Ibexa\Tests\integration\Core\FieldType\Url\UrlStorage\Gateway;
```
alongosz: A typo:
```suggestion
namespace Ibexa\Tests\Integration\Core\FieldType\Url\UrlStorage\Gateway;
```

eZ/Publish/Core/FieldType/Tests/Integration/Url/UrlStorage/UrlStorageGatewayTest.php
```diff
@@ -0,0 +1,41 @@
...
+ * Url Field Type external storage gateway tests.
+ */
+abstract class UrlStorageGatewayTest extends BaseCoreFieldTypeIntegrationTest
+{
+    abstract protected function getGateway(): Gateway;
+
+    /**
+     * @covers \eZ\Publish\Core\FieldType\Url\UrlStorage\Gateway\DoctrineStorage::getUrlsFromUrlLink
+     */
+    public function testGetUrlsFromUrlLink()
```
alongosz:
```suggestion
    public function testGetUrlsFromUrlLink(): void
```

eZ/Publish/Core/FieldType/Tests/Integration/Url/UrlStorage/UrlStorageGatewayTest.php
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Url Field Type external storage gateway tests.
+ */
+abstract class UrlStorageGatewayTest extends BaseCoreFieldTypeIntegrationTest
+{
+    abstract protected function getGateway(): Gateway;
+
+    /**
+     * @covers \eZ\Publish\Core\FieldType\Url\UrlStorage\Gateway\DoctrineStorage::getUrlsFromUrlLink
+     */
```
alongosz: When using `@covers` annotation for a specific method, [only that method will be covered, omitting the method it calls](https://docs.phpunit.de/en/9.6/code-coverage-analysis.html#code-coverage-analysis-specifying-covered-parts). Please move this into the actual child class (`./tests/integration/Core/FieldType/Url/UrlStorage/Gateway/UrlDoctrineStorageGatewayTest.php`) PHPDoc comment as 
```php
/**
 * @covers \eZ\Publish\Core\FieldType\Url\UrlStorage\Gateway\DoctrineStorage
 */
final class UrlDoctrineStorageGatewayTest
```

eZ/Publish/Core/FieldType/Url/UrlStorage/Gateway/DoctrineStorage.php
```diff
@@ -129,6 +129,48 @@ public function insertUrl($url)
         );
     }
 
+    /**
+     * Return a list of URLs used by the given field and version.
+     *
+     * @return bool[] An array of URLs, with urls as keys
```
alongosz: This type specification is not correct. ATM you're returning `array<string, boolean>`

eZ/Publish/Core/FieldType/Url/UrlStorage/Gateway/DoctrineStorage.php
```diff
@@ -129,6 +129,48 @@ public function insertUrl($url)
...
+     */
+    public function getUrlsFromUrlLink(int $fieldId, int $versionNo): array
+    {
+        $selectQuery = $this->connection->createQueryBuilder();
+        $selectQuery
+            ->select($this->connection->quoteIdentifier('url.url'))
+            ->from($this->connection->quoteIdentifier(self::URL_TABLE), 'url')
+            ->leftJoin(
```
alongosz: Any reason for using left join instead of [inner] join?

eZ/Publish/Core/FieldType/Url/UrlStorage/Gateway/DoctrineStorage.php
alongosz: > It could be an option to return the array as-is, and instead to this foreach in [`vendor/ezsystems/ezplatform-richtext/src/lib/eZ/FieldType/RichText/RichTextStorage/Gateway.php`](https://github.com/ezsystems/ezplatform-richtext/pull/242/files#diff-c54260f9f2ffb301f5dfffd2821f50f16cceca1d7b05fcd2c3fcca31c4e4dc51)

If it's possible without breaking changes, that's definitely preferred.

ATM to get current result you could do instead
```php
        $rows = $statement->fetchFirstColumn();
        $result = [];
        foreach ($rows as $url) {
            $result[$url] = true;
        }
```
to reduce time & memory complexity (`fetchAllAssociativeIndexed` iterates over a result and then you iterate over it again accumulating more memory and getting 2 * O(k) time instead of O(k)).

But as said, if it's possible to use simple return here and indeed you need this indexed by URL, so you can do `isset` later on, then maybe it's a better option.

---

Another idea: I see the code using this is as follows:
```php
$urlIdMap = $this->gateway->getUrlIdMap( // ...
// ...
$urlLinkSet = $this->gateway->getUrlsFromUrlLink( // ...
foreach ($links as $index => $link) {
// ..

                if (!isset($urlIdMap[$url])) {
                    $urlIdMap[$url] = $this->gateway->insertUrl($url);
                }
                if (!isset($urlLinkSet[$url])) {
                    $this->gateway->linkUrl(
                        $urlIdMap[$url],
                        $field->id,
                        $versionInfo->versionNo
                    );
                    $urlLinkSet[$url] = true;
                }
// ...
```
What if instead of calling `getUrlIdMap` and `getUrlsFromUrlLink` we could call one method which would give us a map of `url => id` and use that there in richtext?

eZ/Publish/Core/FieldType/Tests/Integration/Url/UrlStorage/UrlStorageGatewayTest.php
vidarl: No good reason for using the old namespace, no. Just forgot about it ( again....)... sorry

Moved test as suggested. Only reason why I did it the other way was because I saw that was done with other tests. Thank you for the explanation

All fixed in 26e6145d9ac897fa2181b69fa42345a83129cdfe

eZ/Publish/Core/FieldType/Tests/Integration/Url/UrlStorage/UrlStorageGatewayTest.php
vidarl: Fixed in 26e6145d9ac897fa2181b69fa42345a83129cdfe

eZ/Publish/Core/FieldType/Tests/Integration/Url/UrlStorage/UrlStorageGatewayTest.php
vidarl: Fixed in 26e6145d9ac897fa2181b69fa42345a83129cdfe

eZ/Publish/Core/FieldType/Url/UrlStorage/Gateway/DoctrineStorage.php
vidarl: Fixed in 26e6145d9ac897fa2181b69fa42345a83129cdfe

eZ/Publish/Core/FieldType/Url/UrlStorage/Gateway/DoctrineStorage.php
vidarl: No, inner join is correct. Fixed in 26e6145d9ac897fa2181b69fa42345a83129cdfe

tests/integration/Core/FieldType/Url/UrlStorage/Gateway/UrlDoctrineStorageGatewayTest.php
vidarl: Fixed in 26e6145d9ac897fa2181b69fa42345a83129cdfe

eZ/Publish/Core/FieldType/Url/UrlStorage/Gateway/DoctrineStorage.php
vidarl: > What if instead of calling getUrlIdMap and getUrlsFromUrlLink we could call one method which would give us a map of url => id and use that there in richtext?

I find this tricky, sql wise

Consider the following row set:
```
MariaDB [ezp]> select * from ezurl as url left join ezurl_object_link as link on url.id = link.url_id where url.url in ("https://ibexa.co", "https://ez.no", "https://newurl");
+----+------------+----------+--------------+------------+----------------------------------+------------------+----------------------------+---------------------------------+--------+
| id | created    | is_valid | last_checked | modified   | original_url_md5                 | url              | contentobject_attribute_id | contentobject_attribute_version | url_id |
+----+------------+----------+--------------+------------+----------------------------------+------------------+----------------------------+---------------------------------+--------+
|  5 | 1705676561 |        1 |            0 | 1705676561 | f24389c80f5500a24427031fd45d283d | https://ibexa.co |                     183982 |                               1 |      5 |
|  5 | 1705676561 |        1 |            0 | 1705676561 | f24389c80f5500a24427031fd45d283d | https://ibexa.co |                     183702 |                               2 |      5 |
|  4 | 1705676559 |        1 |            0 | 1705676559 | d74110041197e107722d8821f5f4d89c | https://ez.no    |                     183703 |                               2 |      4 |
|  6 | 1705676559 |        1 |            0 | 1705676559 | aabbvv041197e107722d8821f5f4d89c | /only_in_url     |                       NULL |                            NULL |   NULL |
|  3 | 1705676559 |        1 |            0 | 1705676559 | f24389c80f5500a24427031fd45d283d | https://ibexa.co |                       NULL |                            NULL |   NULL |
+----+------------+----------+--------------+------------+----------------------------------+------------------+----------------------------+---------------------------------+--------+
```

Let's say you want to find the propriate rows for `contentobject_attribute_id=183702`
You'll need to adjust the query so that:
- you get the 2nd row (since contentobject_attribute_id matches 183702)
- you get the 3rd and 4th row (because `https://ez.no` do not have any other rows where contentobject_attribute_id=183702). Maybe possible using a advanced subquery and !EXISTS()?
Maybe you have an idea? if not, I can still persuit this approach if you insist even though I am not so sure it will have much benifit.

(btw, not sure why I have two rows in ezurl table with same url (`url=https://ibexa.co`, id=`3` and id=`5`) but this seems also possible for whatever reason

alongosz: > > What if instead of calling getUrlIdMap and getUrlsFromUrlLink we could call one method which would give us a map of url => id and use that there in richtext?
> 
> I find this tricky, sql wise
> 
> (...)
> 
>   Maybe you have an idea? if not, I can still persuit this approach if you insist even though I am not so sure it will have much benifit.
> 

I don't insist, if it doesn't make sense to you. TBH the issue is hard to follow given I'm not able to respond in time due to too big backlog of things. I trust your judgement :)

How about my first suggestion of `fetchFirstColumn` then?

vidarl: Regarding `fetchFirstColumn`, yes, that is implemented, see 26e6145d9ac897fa2181b69fa42345a83129cdfe

And the foreach loop has been moved from `DoctrineStorage::getUrlsFromUrlLink()` to `EzSystems\EzPlatformRichText\eZ\FieldType\RichText\RichTextStorage\Gateway::getUrlsFromUrlLink()` in c6c0c22428a6948b6134459d98e31de76bc5ee1e and https://github.com/ezsystems/ezplatform-richtext/pull/242/commits/0a3c480163686e3a960ae6337c5807bd963d4a08

PR #409, opened 2024-07-24
IBX-8562: Fixed flooding content attributes table with duplicates

Nattfarinn:
\+ IBX-8492: Fixed replacing only field of Content Type with virtual field

| **JIRA issue**                          | [IBX-8562](https://issues.ibexa.co/browse/IBX-8562), [IBX-8492](https://issues.ibexa.co/browse/IBX-8492)

#### Checklist:

### DIFF
--- a/eZ/Publish/Core/Persistence/Legacy/Content/FieldHandler.php
+++ b/eZ/Publish/Core/Persistence/Legacy/Content/FieldHandler.php
@@ -332,7 +332,7 @@ class FieldHandler
                 if (isset($updateFieldMap[$fieldDefinition->id][$languageCode])) {
                     $field = clone $updateFieldMap[$fieldDefinition->id][$languageCode];
                     $field->versionNo = $content->versionInfo->versionNo;
-                    if (null !== $field->id && array_key_exists($field->languageCode, $existingLanguageCodes)) {
+                    if (isset($field->id) && array_key_exists($field->languageCode, $existingLanguageCodes)) {
                         $this->updateField($field, $content);
                         $updatedFields[$fieldDefinition->id][$languageCode] = $field;
                     } else {
--- a/eZ/Publish/Core/Persistence/Legacy/Content/FieldHandler.php
+++ b/eZ/Publish/Core/Persistence/Legacy/Content/FieldHandler.php
@@ -363,8 +363,7 @@ class FieldHandler
                     $field->versionNo = $content->versionInfo->versionNo;
                     // Persist virtual field
                     if (null === $field->id) {
-                        $this->updateField($field, $content);
-                        $updatedFields[$fieldDefinition->id][$languageCode] = $field;
+                        $this->createNewField($field, $content);
                     }
                 }
 
--- a/eZ/Publish/Core/Persistence/Legacy/Content/FieldHandler.php
+++ b/eZ/Publish/Core/Persistence/Legacy/Content/FieldHandler.php
@@ -332,7 +332,7 @@ class FieldHandler
                 if (isset($updateFieldMap[$fieldDefinition->id][$languageCode])) {
                     $field = clone $updateFieldMap[$fieldDefinition->id][$languageCode];
                     $field->versionNo = $content->versionInfo->versionNo;
-                    if (isset($field->id) && array_key_exists($field->languageCode, $existingLanguageCodes)) {
+                    if (null !== $field->id && array_key_exists($field->languageCode, $existingLanguageCodes)) {
                         $this->updateField($field, $content);
                         $updatedFields[$fieldDefinition->id][$languageCode] = $field;
                     } else {
@@ -358,6 +358,14 @@ class FieldHandler
                     // also update copied field data
                     // Register for processing after all given fields are updated
                     $nonTranslatableCopiesUpdateSet[$fieldDefinition->id][] = $languageCode;
+                } elseif (isset($contentFieldMap[$fieldDefinition->id][$languageCode])) {
+                    $field = clone $contentFieldMap[$fieldDefinition->id][$languageCode];
+                    $field->versionNo = $content->versionInfo->versionNo;
+                    // Persist virtual field
+                    if (null === $field->id) {
+                        $this->updateField($field, $content);
+                        $updatedFields[$fieldDefinition->id][$languageCode] = $field;
+                    }
                 }
 
                 // If no above conditions were met - do nothing
@@ -417,7 +425,7 @@ class FieldHandler
      * @param \eZ\Publish\SPI\Persistence\Content\Field[] $fields
      * @param array $languageCodes
      *
-     * @return \eZ\Publish\SPI\Persistence\Content\Field[][]
+     * @return array<int, array<string, \eZ\Publish\SPI\Persistence\Content\Field>>
      */
     protected function getFieldMap(array $fields, &$languageCodes = null)
     {
--- a/eZ/Publish/Core/Persistence/Legacy/Content/Handler.php
+++ b/eZ/Publish/Core/Persistence/Legacy/Content/Handler.php
@@ -320,7 +320,9 @@ class Handler implements BaseContentHandler
             $this->contentGateway->loadVersionedNameData([[
                 'id' => $id,
                 'version' => $rows[0]['ezcontentobject_version_version'],
-            ]])
+            ]]),
+            'ezcontentobject_',
+            $translations
         );
         $content = $contentObjects[0];
         unset($rows, $contentObjects);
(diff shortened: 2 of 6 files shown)


Steveb-p:
Given how the issue comes up when multiple layers interact with each other, this almost requires tests - either adjustment of already present ones, or new ones.

I'd accept this only with an explanation why this works, ideally with a promise of tests being added at least to 4.6 to show the issue in the new testing framework.

eZ/Publish/Core/Persistence/Legacy/Content/Mapper.php
```diff
@@ -363,7 +368,8 @@ private function loadCachedVersionFieldDefinitionsPerLanguage(
                 continue;
             }
 
-            $languageCodes = $this->extractLanguageCodesFromMask($languageMask, $allLanguages);
+            $allLanguagesCodes = $this->extractLanguageCodesFromMask($languageMask, $allLanguages);
+            $languageCodes = empty($translations) ? $allLanguagesCodes : array_intersect($translations, $allLanguagesCodes);
```
alongosz, line 374: I think that you need here either 
1) in L355 `array $translations = []` instead
or 
2) 
```suggestion
            $languageCodes = null === $translations ? $allLanguagesCodes : array_intersect($translations, $allLanguagesCodes);
```

The latter one allows passing `[]` which would mean skip all language codes. If this doesn't make sense in any case, than maybe we should move for 1) and leave `empty()` check here.

Nattfarinn: I'm afraid we can't.

Ref.: https://github.com/ezsystems/ezplatform-kernel/blob/1.3/eZ/Publish/SPI/Persistence/Content/Handler.php#L63

It may be `string[]` or `null` and both, empty array and null value, are treated the same way (with empty). Ref.: https://github.com/ezsystems/ezplatform-kernel/blob/1.3/eZ/Publish/Core/Persistence/Legacy/Content/Gateway/DoctrineDatabase.php#L808

mnocon:
The JIRA ticket mentions a SQL script to remove the duplicated entries - if we decide to create one please let us (the Knowledge Team) know, we will have to document it in the update doc

micszo (commented):
Current status of retests.

https://issues.ibexa.co/browse/IBX-8562
OK

https://issues.ibexa.co/browse/IBX-8492
Content view works but editing item causes error:
(under verification by @Nattfarinn)

Screenshot 2024-07-26 at 14 26 07

eWert-Online:
Hello 👋 

since the creation of this PR, our clients database grew an additional 5 million rows and 2 GB in disk size. 
Please provide a fix for this issue as fast as possible. 

Thank you!

db

eZ/Publish/Core/Persistence/Legacy/Content/FieldHandler.php
```diff
@@ -332,7 +332,7 @@ public function updateFields(Content $content, UpdateStruct $updateStruct, Type
                 if (isset($updateFieldMap[$fieldDefinition->id][$languageCode])) {
                     $field = clone $updateFieldMap[$fieldDefinition->id][$languageCode];
                     $field->versionNo = $content->versionInfo->versionNo;
-                    if (isset($field->id) && array_key_exists($field->languageCode, $existingLanguageCodes)) {
+                    if (null !== $field->id && array_key_exists($field->languageCode, $existingLanguageCodes)) {
```
Steveb-p: What's the difference? `isset` isn't working with this magic call (I assume)?

eZ/Publish/Core/Persistence/Legacy/Tests/Content/ContentHandlerTest.php
```diff
@@ -1599,6 +1616,18 @@ protected function getContentTypeHandlerMock()
         return $this->contentTypeHandlerMock;
     }
 
+    /**
+     * @return \PHPUnit\Framework\MockObject\MockObject|\eZ\Publish\Core\Persistence\Legacy\Content\Language\Handler
```
Steveb-p:
```suggestion
     * @return \PHPUnit\Framework\MockObject\MockObject&\eZ\Publish\Core\Persistence\Legacy\Content\Language\Handler
```

eZ/Publish/Core/Persistence/Legacy/Tests/Content/ContentHandlerTest.php
```diff
@@ -108,6 +109,11 @@ class ContentHandlerTest extends TestCase
      */
     protected $contentTypeHandlerMock;
 
+    /**
+     * @var \eZ\Publish\Core\Persistence\Legacy\Content\Language\Handler
+     */
+    protected $languageHandlerMock;
```
Steveb-p: Prefer `private`, unless we have a reason to give the extending class access to this service (descending class can declare it's own private field anyway, if needed).
```suggestion
    private $languageHandlerMock;
```

eZ/Publish/Core/Persistence/Legacy/Tests/Content/FieldHandlerTest.php
Steveb-p, line 1: I'm trying to see if we could have a test here that would address the issue that we encountered. Could you take a look and see?

Because I assume that it's the `FieldHandler` that misbehaves in this specific circumstance?

micszo:
Retest status: on v3.3 404 persists when editing item with only field replaced.

eZ/Publish/Core/Persistence/Legacy/Content/FieldHandler.php
Nattfarinn: It is just to satisfy PHPStans complain about "issetability" of `$field->id`. Probably baseline would be sufficient solution.

Steveb-p: I'll need to see the error myself. `isset` should be properly respected in terms of data type change by PHPStan.

alongosz: >>> What's the difference? `isset` isn't working with this magic call (I assume)?

>> It is just to satisfy PHPStans complain about "issetability" of `$field->id`. Probably baseline would be sufficient solution.

> I'll need to see the error myself. `isset` should be properly respected in terms of data type change by PHPStan.

AFAICS, this is a DTO with public properties. What's the exact PHPStan issue @Nattfarinn ?

eZ/Publish/Core/Persistence/Legacy/Content/Handler.php
```diff
@@ -275,6 +283,14 @@ public function createDraftFromVersion($contentId, $srcVersion, $userId, ?string
         // Clone fields from previous version and append them to the new one
         $this->fieldHandler->createExistingFieldsInNewVersion($content);
 
+        // Persist virtual fields
+        $contentType = $this->contentTypeHandler->load($content->versionInfo->contentInfo->contentTypeId);
+        $this->fieldHandler->updateFields($content, new UpdateStruct([
+            'initialLanguageId' => $this->languageHandler->loadByLanguageCode(
+                $content->versionInfo->initialLanguageCode
+            )->id,
+        ]), $contentType);
```
alongosz, line 294: Please update PHPDoc with  `@throws \eZ\Publish\API\Repository\Exceptions\NotFoundException` exception thrown from this new code.

eZ/Publish/Core/Persistence/Legacy/Tests/Content/ContentHandlerTest.php
```diff
@@ -108,6 +109,11 @@ class ContentHandlerTest extends TestCase
      */
     protected $contentTypeHandlerMock;
 
+    /**
+     * @var \eZ\Publish\Core\Persistence\Legacy\Content\Language\Handler
```
alongosz: Please keep the type consistent: intersection of MockObject & Handler

alongosz (commented):
~I'm still missing some coverage on integration level, with database.
You should be able to query database via tests by getting `\Doctrine\DBAL\Connection` service from the test container (it's already exposed).~
// for 4.6

micszo (approved):
OK from my side now.

eWert-Online:
Hi 🙂 

we now have ~10 million duplicates in our table and the database grew around 3GB in disk size.

Since a few days, the production database is throwing error messages (`SQLSTATE[HY000]: General error: 1114 The table 'ezcontentobject_attribute' is full`). 
Some SQL queries fail with `Fatal Error: Allowed memory size exhausted`.

...but sure... take your time discussing what kind of null check you want to do. Nothing to worry about!

fine

adamwojs:
Hi @eWert-Online!

First of all my apologies for the situation. I understand your frustration but I would like to ensure you that we are aware of importance of this patch and we are actively working on providing the solution. 

Unfortunately, Github does not give a full visibility of our intense activities behind the scenes. Currently proposed changes are going through Quality Assurance process. As soon as we will be ready we will do the 3.3.40 and 4.6.10 release.

adamwojs (approved):

konradoboza (approved):

Steveb-p (approved):

alongosz (approved):

ciastektk (approved):

PR #410, opened 2024-08-14
IBX-8562: Command to remove duplicated entries after faulty IBX-5388 fix

Nattfarinn:
# ~Do not merge: TEMP commit inside (ref. https://github.com/ezsystems/ezplatform-kernel/pull/410/commits/46e0e4a8a5233fd77765a6120f94f6ffbee7e4f8 )~ (commit removed)

#### Branch base is set to https://github.com/ezsystems/ezplatform-kernel/commit/98b7b50e60124a3bc7e6d79c123d57a31d0427bd commit to make QA life a bit easier.
 
| **JIRA issue**                          | [IBX-8562](https://issues.ibexa.co/browse/IBX-8562)

Command removes duplicated database entries after faulty IBX-5388 fix (base commit for this branch). 

Usage:
```
  ibexa:content:remove-duplicate-fields [options]
```
Options:
```
  -b, --batch-size[=BATCH-SIZE]          Number of attributes affected per iteration [default: 10000]
  -i, --max-iterations[=MAX-ITERATIONS]  Max iterations count (default or -1: unlimited) [default: -1]
  -s, --sleep[=SLEEP]                    Wait between iterations, in milliseconds [default: 0]
  -f, --force                            Force operation (implies non-interactive mode)
```

### Code is compatible with PHP 7.3+. There should be no need to provide separate command for upper DXP versions (4.6 and above) with code adjusted to newer PHP versions and coding standards.

### DIFF
--- a/src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php
+++ b/src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php
@@ -1,0 +1,258 @@
+<?php
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace Ibexa\Bundle\Core\Command;
+
+use Doctrine\DBAL\Connection;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\Stopwatch\Stopwatch;
+
+final class VirtualFieldDuplicateFixCommand extends Command
+{
+    private const DEFAULT_BATCH_SIZE = 10000;
+
+    private const MAX_ITERATIONS_UNLIMITED = -1;
+
+    private const DEFAULT_SLEEP = 0;
+
+    protected static $defaultName = 'ibexa:content:remove-duplicate-fields';
+
+    protected static $defaultDescription = 'Removes duplicate fields created as a result of faulty IBX-5388 performance fix.';
+
+    /** @var \Doctrine\DBAL\Connection */
+    private $connection;
+
+    public function __construct(
+        Connection $connection
+    ) {
+        parent::__construct();
+
+        $this->connection = $connection;
+    }
+
+    public function configure(): void
+    {
+        $this->addOption(
+            'batch-size',
+            'b',
+            InputOption::VALUE_REQUIRED,
+            'Number of attributes affected per iteration',
+            self::DEFAULT_BATCH_SIZE
+        );
+
+        $this->addOption(
+            'max-iterations',
+            'i',
+            InputOption::VALUE_REQUIRED,
+            'Max iterations count (default or -1: unlimited)',
+            self::MAX_ITERATIONS_UNLIMITED
+        );
+
+        $this->addOption(
+            'sleep',
+            's',
+            InputOption::VALUE_REQUIRED,
+            'Wait between iterations, in milliseconds',
+            self::DEFAULT_SLEEP
+        );
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $style = new SymfonyStyle($input, $output);
+        $stopwatch = new Stopwatch(true);
+        $stopwatch->start('total', 'command');
+
+        $batchSize = (int)$input->getOption('batch-size');
+        if ($batchSize === 0) {
+            $style->warning('Batch size is set to 0. Nothing to do.');
+
+            return Command::INVALID;
+        }
+
+        $maxIterations = (int)$input->getOption('max-iterations');
+        if ($maxIterations === 0) {
+            $style->warning('Max iterations is set to 0. Nothing to do.');
+
+            return Command::INVALID;
+        }
+
+        $sleep = (int)$input->getOption('sleep');
+
+        $totalCount = $this->getDuplicatedAttributeTotalCount($style, $stopwatch);
+
+        if ($totalCount === 0) {
+            $style->success('Database is clean of attribute duplicates. Nothing to do.');
+
+            return Command::SUCCESS;
+        }
+
+        if ($input->isInteractive()) {
+            $confirmation = $this->askForConfirmation($style);
+            if (!$confirmation) {
+                $style->info('Confirmation rejected. Terminating.');
+
+                return Command::FAILURE;
+            }
+        }
+
+        $iteration = 1;
+        $totalDeleted = 0;
+        do {
+            $deleted = 0;
+            $stopwatch->start('iteration', 'sql');
+
+            $attributes = $this->getDuplicatedAttributesBatch($batchSize);
+            foreach ($attributes as $attribute) {
+                $attributeIds = $this->getDuplicatedAttributeIds($attribute);
+
+                if (!empty($attributeIds)) {
+                    $iterationDeleted = $this->deleteAttributes($attributeIds);
+
+                    $deleted += $iterationDeleted;
+                    $totalDeleted += $iterationDeleted;
+                }
+            }
+
+            $style->info(
+                sprintf(
+                    'Iteration %d: Removed %d duplicate database rows (total removed this execution: %d). [Debug %s]',
+                    $iteration,
+                    $deleted,
+                    $totalDeleted,
+                    $stopwatch->stop('iteration')
+                )
+            );
+
+            if ($maxIterations !== self::MAX_ITERATIONS_UNLIMITED && ++$iteration > $maxIterations) {
+                $style->warning('Max iterations count reached. Terminating.');
+
+                return self::SUCCESS;
+            }
+
+            // Wait, if needed, before moving to next iteration
+            usleep($sleep * 1000);
+        } while ($batchSize === count($attributes));
+
+        $style->success(sprintf(
+            'Operation successful. Removed total of %d duplicate database rows. [Debug %s]',
+            $totalDeleted,
+            $stopwatch->stop('total')
+        ));
+
+        return Command::SUCCESS;
+    }
+
+    private function getDuplicatedAttributeTotalCount(
+        SymfonyStyle $style,
+        Stopwatch $stopwatch
+    ): int {
+        $stopwatch->start('total_count', 'sql');
+        $query = $this->connection->createQueryBuilder()
+            ->select('COUNT(a.id) as instances')
+            ->groupBy('version', 'contentclassattribute_id', 'contentobject_id', 'language_id')
+            ->from('ezcontentobject_attribute', 'a')
+            ->having('instances > 1');
+
+        $count = $query->execute()->rowCount();
+
+        if ($count > 0) {
+            $style->warning(
+                sprintf(
+                    'Found %d of affected attributes. [Debug: %s]',
+                    $count,
+                    $stopwatch->stop('total_count')
+                )
+            );
+        }
+
+        return $count;
+    }
+
+    /**
+     * @phpstan-return array<array{
+     *     version: int,
+     *     contentclassattribute_id: int,
+     *     contentobject_id: int,
+     *     language_id: int,
+     * }>
+     */
+    private function getDuplicatedAttributesBatch(int $batchSize): array
+    {
+        $query = $this->connection->createQueryBuilder();
+
+        $query
+            ->select('version', 'contentclassattribute_id', 'contentobject_id', 'language_id')
+            ->groupBy('version', 'contentclassattribute_id', 'contentobject_id', 'language_id')
+            ->from('ezcontentobject_attribute')
+            ->having('COUNT(id) > 1')
+            ->setFirstResult(0)
+            ->setMaxResults($batchSize);
+
+        return $query->execute()->fetchAllAssociative();
+    }
+
+    /**
+     * @phpstan-param array{
+     *     version: int,
+     *     contentclassattribute_id: int,
+     *     contentobject_id: int,
+     *     language_id: int
+     * } $attribute
+     *
+     * @return int[]
+     */
+    private function getDuplicatedAttributeIds(array $attribute): array
+    {
+        $query = $this->connection->createQueryBuilder();
+
+        $query
+            ->select('id')
+            ->from('ezcontentobject_attribute')
+            ->andWhere('version = :version')
+            ->andWhere('contentclassattribute_id = :contentclassattribute_id')
+            ->andWhere('contentobject_id = :contentobject_id')
+            ->andWhere('language_id = :language_id')
+            ->orderBy('id', 'ASC')
+            // Keep the original attribute row, the very first one
+            ->setFirstResult(1);
+
+        $query->setParameters($attribute);
+        $result = $query->execute()->fetchFirstColumn();
+
+        return array_map('intval', $result);
+    }
+
+    private function askForConfirmation(SymfonyStyle $style): bool
+    {
+        $style->warning('Operation is irreversible.');
+
+        return $style->askQuestion(
+            new ConfirmationQuestion(
+                'Proceed with deletion?',
+                false
+            )
+        );
+    }
+
+    private function deleteAttributes($ids): int
+    {
+        $query = $this->connection->createQueryBuilder();
+
+        $query
+            ->delete('ezcontentobject_attribute')
+            ->andWhere($query->expr()->in('id', $ids));
+
+        return (int)$query->execute();
+    }
+}
--- a/eZ/Bundle/EzPublishCoreBundle/Resources/config/commands.yml
+++ b/eZ/Bundle/EzPublishCoreBundle/Resources/config/commands.yml
@@ -62,3 +62,11 @@ services:
             $userHandler: '@ezpublish.spi.persistence.user_handler'
         tags:
             - { name: console.command }
+
+    Ibexa\Bundle\Core\Command\VirtualFieldDuplicateFixCommand:
+        autowire: true
+        autoconfigure: true
+        arguments:
+            $connection: '@ezpublish.persistence.connection'
+        tags:
+            - { name: console.command }
--- a/eZ/Bundle/EzPublishDebugBundle/Twig/DebugTemplate.php
+++ b/eZ/Bundle/EzPublishDebugBundle/Twig/DebugTemplate.php
@@ -76,9 +76,9 @@ class DebugTemplate extends Template
         return new Source('', '');
     }
 
-    protected function doDisplay(array $context, array $blocks = []): string
+    protected function doDisplay(array $context, array $blocks = []): iterable
     {
-        return '';
+        return [];
     }
 
     /**


src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php
```diff
@@ -0,0 +1,272 @@
...
+    private const DEFAULT_SLEEP = 0;
+
+    /** @var \Doctrine\DBAL\Connection */
+    private $connection;
+
+    public function __construct(
+        Connection $connection
+    ) {
+        parent::__construct('ibexa:content:remove-duplicate-fields');
+        $this->setDescription('Removes duplicate fields created as a result of faulty IBX-5388 performance fix.');
```
Steveb-p: use static `$name` and `$description` property instead, as it makes the command lazy (i.e. it does not initialize itself for command lists).

src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php
```diff
@@ -0,0 +1,272 @@
...
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $style = new SymfonyStyle($input, $output);
+        $stopwatch = new Stopwatch(true);
+        $stopwatch->start('total', 'command');
+
+        $force = $input->getOption('force');
+        if ($force) {
+            $input->setInteractive(false);
```
Steveb-p: Interactivity of a command is usually controlled by `-n` or `--no-interaction` options (and sometimes it disables itself based on environmental variables).

`--force` should be used as a "confirmation" that the user understands that he/she performs a potentially destructive operation.

src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php
```diff
@@ -0,0 +1,272 @@
...
+                if (!$confirmation && !$force) {
+                    $style->info('Confirmation rejected. Terminating.');
+
+                    return Command::FAILURE;
+                }
+            } else {
+                $style->success('Database is clean of attribute duplicates. Nothing to do.');
+
+                return Command::SUCCESS;
+            }
```
Steveb-p: Early return instead. Confirmation is not "dependent" on total count of duplicate attributes.

src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php
```diff
@@ -0,0 +1,272 @@
...
+                usleep($sleep * 1000);
+            } while ($batchSize === count($attributes));
+
+            $style->success(sprintf(
+                'Operation successful. Removed total of %d duplicates. [Debug %s]',
+                $totalDeleted,
+                $stopwatch->stop('total')
+            ));
+        } catch (Exception $exception) {
+            $style->error($exception->getMessage());
```
Steveb-p: If you intend to break execution anyway on exception, then allow it to "fall through". You don't need to catch it.

src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php
```diff
@@ -0,0 +1,272 @@
...
+     * @return int[]
+     */
+    private function getDuplicatedAttributeIds(array $attribute): array
+    {
+        $query = $this->connection->createQueryBuilder();
+
+        $query
+            ->select('id')
+            ->from('ezcontentobject_attribute')
+            ->where('version = :version')
```
Steveb-p:
```suggestion
            ->andWhere('version = :version')
```

src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php
```diff
@@ -0,0 +1,272 @@
...
+        );
+    }
+
+    private function deleteAttributes($ids): int
+    {
+        $query = $this->connection->createQueryBuilder();
+
+        $query
+            ->delete('ezcontentobject_attribute')
+            ->where($query->expr()->in('id', $ids));
```
Steveb-p:
```suggestion
            ->andWhere($query->expr()->in('id', $ids));
```

src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php
```diff
@@ -0,0 +1,272 @@
...
+            'b',
+            InputOption::VALUE_OPTIONAL,
+            'Number of attributes affected per iteration',
+            self::DEFAULT_BATCH_SIZE
+        );
+
+        $this->addOption(
+            'max-iterations',
+            'i',
+            InputOption::VALUE_OPTIONAL,
```
Steveb-p: This option, once passed, needs to have value.
```suggestion
            InputOption::VALUE_REQUIRED,
```

src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php
```diff
@@ -0,0 +1,272 @@
...
+
+        $this->connection = $connection;
+    }
+
+    public function configure(): void
+    {
+        $this->addOption(
+            'batch-size',
+            'b',
+            InputOption::VALUE_OPTIONAL,
```
Steveb-p:
```suggestion
            InputOption::VALUE_REQUIRED,
```

src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php
```diff
@@ -0,0 +1,272 @@
...
+            'i',
+            InputOption::VALUE_OPTIONAL,
+            'Max iterations count (default or -1: unlimited)',
+            self::MAX_ITERATIONS_UNLIMITED
+        );
+
+        $this->addOption(
+            'sleep',
+            's',
+            InputOption::VALUE_OPTIONAL,
```
Steveb-p:
```suggestion
            InputOption::VALUE_REQUIRED,
```

src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php
```diff
@@ -0,0 +1,272 @@
...
+                        $deleted,
+                        $totalDeleted,
+                        $stopwatch->stop('iteration')
+                    )
+                );
+
+                if ($maxIterations !== self::MAX_ITERATIONS_UNLIMITED && ++$iteration > $maxIterations) {
+                    $style->warning('Max iterations count reached. Terminating.');
+
+                    return self::FAILURE;
```
Steveb-p: It's not a failure. It's expected execution state.
```suggestion
                    return self::SUCCESS;
```

src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php
```diff
@@ -0,0 +1,272 @@
...
+
+        $query
+            ->select('id')
+            ->from('ezcontentobject_attribute')
+            ->where('version = :version')
+            ->andWhere('contentclassattribute_id = :contentclassattribute_id')
+            ->andWhere('contentobject_id = :contentobject_id')
+            ->andWhere('language_id = :language_id')
+            ->orderBy('id', 'ASC')
+            ->setFirstResult(0);
```
Steveb-p: You're discarding the first row later.

You can skip it then.

```suggestion
            ->orderBy('id', 'ASC')
            ->setFirstResult(1);
```

Also, are you sure that the first row, with the lowest ID, should be kept? Not the highest one?

src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php
```diff
@@ -0,0 +1,272 @@
...
+                foreach ($attributes as $attribute) {
+                    $attributeIds = $this->getDuplicatedAttributeIds($attribute);
+
+                    $deleted += $this->deleteAttributes($attributeIds);
+                    $totalDeleted += $deleted;
+                }
+
+                $style->info(
+                    sprintf(
+                        'Iteration %d: Removed %d duplicates (total removed this execution: %d). [Debug %s]',
```
micszo: This message could be clearer. 🙂 At this moment it's unclear which is duplicated lines and which attributes.

vidarl:
I think it is a bad idea not to merge and release the script in the product.
User would need to download it and place it in `src/Command/` then?

Running this command would be a one-off, so we can make the command hidden : https://symfony.com/doc/current/console/hide_commands.html

That will prevent the command from being displayed by `console list` command and confuse users later.. Command could be removed from 5.0 again IMO

@adamwojs

mnocon:
I didn't know about the `hidden` Command feature - and I like Vidar's suggestion a lot.

It would greatly improve developer experience for people executing the update (no manual steps needed) without polluting the `ibexa` command namespace (which we wanted to avoid by having the Command in the doc)

alongosz:
The problem with that is that it doesn't scale properly. Command might be hidden, true, but if we wanted for every upgrade to have such command, the DI container would get inflated with commands. We already have a lot of them. We need a streamlined solution that is able to execute such updates via e.g. `ibexa:upgrade`. Planned and postponed for ages.

src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php
alongosz: @Steveb-p do we really need it required if there's a default?

mnocon:
> The problem with that is that it doesn't scale properly. Command might be hidden, true, but if we wanted for every upgrade to have such command, the DI container would get inflated with commands. We already have a lot of them. We need a streamlined solution that is able to execute such updates via e.g. `ibexa:upgrade`. Planned and postponed for ages.

You mention the DI container getting inflated with Commands as the main argument against it - but what is the real cost of it? The container taking miliseconds longer to build?

Also we're not discussing every future update here - only this specific case where @Nattfarinn decided that it's better to use a Command than a SQL script.

For me the `hidden command` is a good tradeoff, because we gain a lot in developer experience for the Partners - they don't have to go through all the cumbersome steps (create files manually, add service definitions) to fix an issue we're responsible for.

alongosz:
> You mention the DI container getting inflated with Commands as the main argument against it - but what is the real cost of it? The container taking miliseconds longer to build?
> Also we're not discussing every future update here - only this specific case

I'm talking about the scale. You cannot say that this is only for this case, because it sets precedence. Or rather in this case continues anti-pattern we've used before. We already have a lot of commands that could be wrapped into one streamlined solution.

However, if you really want it in the product, it should follow patterns existing in product:
1. Move all SQL to a Gateway.
2. Add integration tests against that gateway.


This adds significant amount of work to it which could be better spent on a generic upgrade solution. But yes, still much less work to fix this individually than work on a generic one.

For the Official Documentation the solution was fine due to its compactness.

So, to sum up, I'd be fine with this command being hidden and in the product, but it still needs to have layer separation in that case.

src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php
Steveb-p: It is an **option**. Therefore, if you pass `--batch-size`, you expect to provide a value. Therefore, it is required.

```
<command> --batch-size 10
```

Calling this command and using this option without value makes no practical sense, and the value of the option would be simply `true`, which is definitely not what you want.
```
<command> --batch-size
```

Steveb-p:
Theoretical SQL fix for the issue:

PostgreSQL:
```sql
DELETE FROM ezcontentobject_attribute
WHERE ezcontentobject_attribute.id IN (
    SELECT ea_duplicate.id FROM ezcontentobject_attribute AS ea_duplicate
        JOIN ezcontentobject_attribute AS ea ON
            ea.version = ea_duplicate.version
            AND ea.language_id = ea_duplicate.language_id
            AND ea.contentclassattribute_id = ea_duplicate.contentclassattribute_id
            AND ea.contentobject_id = ea_duplicate.contentobject_id
        WHERE ea.id < ea_duplicate.id
)
```

MySQL:
```sql
DELETE ea_duplicate FROM ezcontentobject_attribute AS ea_duplicate
     JOIN ezcontentobject_attribute AS ea ON
        ea.version = ea_duplicate.version
        AND ea.language_id = ea_duplicate.language_id
        AND ea.contentclassattribute_id = ea_duplicate.contentclassattribute_id
        AND ea.contentobject_id = ea_duplicate.contentobject_id
     WHERE ea.id < ea_duplicate.id
```
(Note: for databases it does not matter if condition is part of JOIN or WHERE)

Providing this in case it's valuable, and so it does not get lost in private messages.

Yes, I've tested this manually on databases, and SQL is correct (with their respective language differences). It should give the same result as the above command, **but I haven't checked as I don't currently have an instance of Ibexa DXP with that particular issue**.

src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php
Nattfarinn: I have no strong feeling about this either way; it is matter of PoV. I intended this to be hard limit ("if there are more interations than max, interations exceed expectations", hence failure), not soft limit ("expect to do numer of iterations, without care how many more are there left", hence success). It is matter of intention, and my intention was to make it hard limit. Just how exceeding `max_execution_time` results in an error; I see this as similar case. I would say command should result in success only if there are no more duplicates left.

Steveb-p: Given that this command is never supposed to be called automatically (like in a bash command or something) I can agree to leave it as-is.

However, I disagree that you can assume that having duplicates remaining is a failure condition. The command did exactly what it was asked for.

In a real command we would probably discuss making this an input option, so it can be controlled, but I guess it does not matter in this context.

vidarl:
> You cannot say that this is only for this case, because it sets precedence.

Well, in that case we already have precedence as this is what we have done in the past already. For instance : https://github.com/ibexa/fieldtype-richtext/pull/77

And yes, there we have the gateways, tests etc as you mentioned is needed

src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php
Nattfarinn: Message refers to removed database rows (there was issue in counting elsewhere and made it look like duplicates and total duplicates refers to different things).

alongosz (commented):
@adamwojs we need to make a decision here, either:
1. we decide to go with the in-product hidden command and refactor it to have separate gateway layer at least and integration coverage
2. we go with a regular upgrade script, assuming the SQL provided a few comments above does the job
3. we put this after final review in doc only (there are IMO valid objections in the comments above).

micszo:
ad.2.
A downside of this approach is no support for batches and iterations. Which could be an issue with a significant number of duplicates (in best case, no monitoring of progress).

**Update**: In my tests with 10, 100, 1000 duplicates there is no problematic difference in time between command and direct SQL. With 10k duplicates the command takes seconds and direct SQL several minutes. With 100k duplicates I started the direct SQL in the afternoon and interrupted execution in the morning. With the command 100k is still a matter of seconds.

(In theory this could be related to individual DB config. But then we would have to assume partners would cover tweaking that.)

micszo:
Summing up last meeting. Got following error on a big DB from a live project (14M records):
```
  An exception occurred while executing 'DELETE FROM ezcontentobject_attribute WHERE id IN ()':

  SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ')' at line 1
```

with stack trace:
```
at /Users/michalszoltysek/Downloads/Ibexa-project_latest_repo_22_august_take1/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php:98
 Doctrine\DBAL\Driver\AbstractMySQLDriver->convertException() at /Users/michalszoltysek/Downloads/Ibexa-project_latest_repo_22_august_take1/vendor/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php:182
 Doctrine\DBAL\DBALException::wrapException() at /Users/michalszoltysek/Downloads/Ibexa-project_latest_repo_22_august_take1/vendor/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php:159
 Doctrine\DBAL\DBALException::driverExceptionDuringQuery() at /Users/michalszoltysek/Downloads/Ibexa-project_latest_repo_22_august_take1/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php:2226
 Doctrine\DBAL\Connection->handleExceptionDuringQuery() at /Users/michalszoltysek/Downloads/Ibexa-project_latest_repo_22_august_take1/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php:1538
 Doctrine\DBAL\Connection->executeStatement() at /Users/michalszoltysek/Downloads/Ibexa-project_latest_repo_22_august_take1/vendor/doctrine/dbal/lib/Doctrine/DBAL/Query/QueryBuilder.php:216
 Doctrine\DBAL\Query\QueryBuilder->execute() at /Users/michalszoltysek/Downloads/Ibexa-project_latest_repo_22_august_take1/vendor/ezsystems/ezplatform-kernel/src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php:254
 Ibexa\Bundle\Core\Command\VirtualFieldDuplicateFixCommand->deleteAttributes() at /Users/michalszoltysek/Downloads/Ibexa-project_latest_repo_22_august_take1/vendor/ezsystems/ezplatform-kernel/src/bundle/Core/Command/VirtualFieldDuplicateFixCommand.php:119
 Ibexa\Bundle\Core\Command\VirtualFieldDuplicateFixCommand->execute() at /Users/michalszoltysek/Downloads/Ibexa-project_latest_repo_22_august_take1/vendor/symfony/console/Command/Command.php:298
 Symfony\Component\Console\Command\Command->run() at /Users/michalszoltysek/Downloads/Ibexa-project_latest_repo_22_august_take1/vendor/symfony/console/Application.php:1058
 Symfony\Component\Console\Application->doRunCommand() at /Users/michalszoltysek/Downloads/Ibexa-project_latest_repo_22_august_take1/vendor/symfony/framework-bundle/Console/Application.php:96
 Symfony\Bundle\FrameworkBundle\Console\Application->doRunCommand() at /Users/michalszoltysek/Downloads/Ibexa-project_latest_repo_22_august_take1/vendor/symfony/console/Application.php:301
 Symfony\Component\Console\Application->doRun() at /Users/michalszoltysek/Downloads/Ibexa-project_latest_repo_22_august_take1/vendor/symfony/framework-bundle/Console/Application.php:82
 Symfony\Bundle\FrameworkBundle\Console\Application->doRun() at /Users/michalszoltysek/Downloads/Ibexa-project_latest_repo_22_august_take1/vendor/symfony/console/Application.php:171
 Symfony\Component\Console\Application->run() at /Users/michalszoltysek/Downloads/Ibexa-project_latest_repo_22_august_take1/vendor/symfony/runtime/Runner/Symfony/ConsoleApplicationRunner.php:54
 Symfony\Component\Runtime\Runner\Symfony\ConsoleApplicationRunner->run() at /Users/michalszoltysek/Downloads/Ibexa-project_latest_repo_22_august_take1/vendor/autoload_runtime.php:35
 require_once() at /Users/michalszoltysek/Downloads/Ibexa-project_latest_repo_22_august_take1/bin/console:11
```

vidarl:
We have several customers waiting for this fix. Could we get some progress on it?
@adamwojs, please see andrew's comment https://github.com/ezsystems/ezplatform-kernel/pull/410#pullrequestreview-2266144555 too

@mateuszbieniek

micszo (approved):
The issue found on provided database has been resolved.

I have not found other issues.

tischsoic:
CI green before removing temp commit. ✅

PR #360, opened 2023-01-31
IBX-4477: Added LocationService::getSubtreeSize method

adamwojs:
| Question                                  | Answer
| ---------------------------------------- | ------------------
| **JIRA issue**                          | [IBX-4477](https://issues.ibexa.co/browse/IBX-4477)
| **Required by**                        | ezsystems/ezplatform-admin-ui#2085
| **Type**                                   | bug
| **Target Ibexa version** | `v3.3`
| **BC breaks**                          | no
| **Regression build**         | https://github.com/ibexa/commerce/pull/214

Added dedicated method to retrieve subtree size

```php
interface LocationService {
    # ...

    /**
     * Return the subtree size of a given location.
     *
     * Warning! This method is not permission aware by design.
     */
    public function getSubtreeSize(Location $location): int;

    # ...
}
```

#### Checklist:
- [X] Provided PR description.
- [x] Tested the solution manually.
- [x] Provided automated test coverage.
- [x] Checked that target branch is set correctly (master for features, the oldest supported for bugs).
- [x] Ran PHP CS Fixer for new PHP code (use `$ composer fix-cs`).
- [ ] Asked for a review (ping `@ezsystems/engineering-team`).

PR #412, opened 2025-06-18
IBX-10174 Add limit option to getSubtreeSize

ryanolee:
| **JIRA issue**                          | [IBX-10174](https://issues.ibexa.co/browse/IBX-10174)
| **Target Ibexa version** | `v3/V4`
| **Required by**                      | https://github.com/ezsystems/ezplatform-admin-ui/pull/2124

# What
This PR adds a new "$limit" option to the `LocationService::getSubtreeSize` query in order to reduce the effects of expensive count queries within the CMS. 

In https://github.com/ezsystems/ezplatform-kernel/pull/360 the `LocationService::getSubtreeSize` was introduced to address prior issues with the prior `LocationService::count` being two slow. This PR aims to follow on from that by addressing further issues with `LocationService::getSubtreeSize` resulting in disproportionately slow queries in it's own right.

This has been done by adding a function to limit the overall selected `subTree` size by changing the public API from 
```php
interface LocationService {
    # ...

    public function getSubtreeSize(Location $location): int;

    # ...
}
```
To
```php
interface LocationService {
    # ...

    public function getSubtreeSize(Location $location, ?int $limit  = null): int;

    # ...
}
```

# Why
In some circumstances (Where a content object may have a large number of child items) the impact of a single call to the `getSubtreeSize` function can account for over 30 percent of the total page load time

See the following flame chart for loading the "view" page in the CMS for a content object with a large number of sub-items:
image

With this patch included this becomes
image

The time the function takes to resolve falls from `407MS` -> `1MS` for what is functionally the same query.

# How
This PR introduces a `$limit` parameter that targets the use case presented by `IsWithinCopySubtreeLimit` (The only usecase presented by the current eZ Codebase). The query is aiming to ascertain if there are **more than** a number of given number of sub items in a content object tree. 

The current implementation is prone to over scanning the table due to there being no limit properly set on the table in relation to the primary usecase for the function.

The query it currently generates is:
```sql
SELECT COUNT(node_id) FROM ezcontentobject_tree t WHERE t.path_string LIKE '/1/2/80/136/%'; 
```
Which performs an entire table scan against existing indexes:
```
-> Aggregate: count(t.node_id)  (cost=78629.40 rows=1) (actual time=550.231..550.232 rows=1 loops=1)    -> Filter: (t.path_string like '/1/2/80/136/%')  (cost=58454.40 rows=201750) (actual time=0.028..547.540 rows=85943 loops=1)        -> Covering index range scan on t using ezcontentobject_tree_path over ('/1/2/80/136/                                                                                                                                                                              ' <= path_string <= '/1/2/80/136/????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????')  (cost=58454.40 rows=201750) (actual time=0.024..538.213 rows=85943 loops=1) |
```

In the new approach the  query is terminated early resulting in much faster query times
```sql
 SELECT COUNT(*) FROM (SELECT t.node_id FROM ezcontentobject_tree t WHERE t.path_string LIKE '/1/2/80/136/%' LIMIT 100) AS x;
```
which gives
```
-> Aggregate: count(0)  (cost=58463.76..58463.76 rows=1) (actual time=0.508..0.508 rows=1 loops=1)    -> Table scan on x  (cost=0.04..3.75 rows=100) (actual time=0.002..0.004 rows=100 loops=1)        -> Materialize  (cost=58450.05..58453.76 rows=100) (actual time=0.496..0.502 rows=100 loops=1)            -> Limit: 100 row(s)  (cost=58440.01 rows=100) (actual time=0.045..0.469 rows=100 loops=1)                -> Filter: (t.path_string like '/1/2/80/136/%')  (cost=58440.01 rows=201750) (actual time=0.044..0.463 rows=100 loops=1)                    -> Covering index range scan on t using ezcontentobject_tree_path over ('/1/2/80/136/                                                                                                                                                                  ' <= path_string <= '/1/2/80/136/????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????')  (cost=58440.01 rows=201750) (actual time=0.039..0.446 rows=100 loops=1) |
```

Special care has been taken so that no existing code paths should be effected by this change given the new query is only used _if_ a limitation is supplied (where by default it is not)

#### Checklist:

### DIFF
--- a/eZ/Publish/API/Repository/LocationService.php
+++ b/eZ/Publish/API/Repository/LocationService.php
@@ -133,8 +133,11 @@ interface LocationService
      * Return the subtree size of a given location.
      *
      * Warning! This method is not permission aware by design.
+     * 
+     * @param \eZ\Publish\API\Repository\Values\Content\Location $location
+     * @param int|null $limit Optional limit to the number of locations to count. (Can be used to limit the number of locations counted in large subtrees.)
      */
-    public function getSubtreeSize(Location $location): int;
+    public function getSubtreeSize(Location $location, ?int $limit = null): int;
 
     /**
      * Creates the new $location in the content repository for the given content.
--- a/eZ/Publish/API/Repository/LocationService.php
+++ b/eZ/Publish/API/Repository/LocationService.php
@@ -133,7 +133,7 @@ interface LocationService
      * Return the subtree size of a given location.
      *
      * Warning! This method is not permission aware by design.
-     * 
+     *
      * @param \eZ\Publish\API\Repository\Values\Content\Location $location
      * @param int|null $limit Optional limit to the number of locations to count. (Can be used to limit the number of locations counted in large subtrees.)
      */
--- a/eZ/Publish/API/Repository/Tests/LocationServiceTest.php
+++ b/eZ/Publish/API/Repository/Tests/LocationServiceTest.php
@@ -3552,6 +3552,43 @@ class LocationServiceTest extends BaseTest
         return $location;
     }
 
+    public function testGetSubtreeSizeWithLimit(): Location
+    {
+        $repository = $this->getRepository();
+        $locationService = $repository->getLocationService();
+
+        $folder = $this->createFolder(['eng-GB' => 'Parent Folder'], 2);
+        $location = $folder->getVersionInfo()->getContentInfo()->getMainLocation();
+        self::assertSame(1, $locationService->getSubtreeSize($location));
+
+        for ($i = 1; $i <= 10; $i++) {
+            $this->createFolder(['eng-GB' => 'Child ' . $i], $location->id);
+        }
+
+        self::assertSame(3, $locationService->getSubtreeSize($location, 3));
+
+        return $location;
+    }
+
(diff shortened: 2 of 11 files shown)


ryanolee:
Superseded by https://github.com/ezsystems/ezplatform-kernel/pull/413

